Issue #7, opened 2018-03-14
Gas cost optimization - storage pointer computation for mappings


loredanacirstea:
Not urgent now, but should be considered after we are completely settled on the contracts ABIs.
Any other gas cost optimization suggestions or issues should be gathered here for now.

- https://github.com/raiden-network/raiden-contracts/commit/921c006a8fac5e556e238808667b48016b4c2bab#r28074132
> How clever is the solidity compiler for repeated expressions like channels[last_channel_index]? Will it notice that this is the same storage location and compute the pointer only once?

loredanacirstea:
tl;dr @hackaugusto , seems it computes the pointer every time :|. good observation.  

```js
   function use(uint256 index) public {
        channels[index].settle_timeout = 5;
        channels[index].settle_timeout = 5;
    }
    
    // saves 109 gas
    function use2(uint256 index) public {
        Channel storage channel = channels[index];
        channel.settle_timeout = 5;
        channel.settle_timeout = 5;
    }
```

`use2` - 31756
```
358 SSTORE
359 POP
360 PUSH1 05
362 DUP2
363 PUSH1 00
365 ADD
366 DUP2
367 SWAP1
368 SSTORE
369 POP
```

`use` - 31865
```
567 SSTORE
568 POP
569 PUSH1 05
571 PUSH1 00
573 DUP1
574 DUP4
575 DUP2
576 MSTORE
577 PUSH1 20
579 ADD
580 SWAP1
581 DUP2
582 MSTORE
583 PUSH1 20
585 ADD
586 PUSH1 00
588 SHA3
589 PUSH1 00
591 ADD
592 DUP2
593 SWAP1
594 SSTORE
595 POP
```

hackaugusto:
Humm, interesting, thanks for investigating it :)

LefterisJP:
@loredanacirstea @hackaugusto  In defense of solidity, there are optimizations. You just need to activate them ;)

Example is with `solc v0.4.20`

Example contract:

```solidity
contract foo {

    struct Channel {
        uint256 settle_timeout;
    }

    mapping(uint256 => Channel) channels;
    function use(uint256 index) public {
        channels[index].settle_timeout = 5;
        channels[index].settle_timeout = 5;
    }

    function use2(uint256 index) public {
        Channel storage channel = channels[index];
        channel.settle_timeout = 5;
        channel.settle_timeout = 5;
    }
}
```

> solc --gas TEST.sol

```
[...]

PR #8, opened 2018-03-15
Compute mapping pointer once

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/7

Initial gas cost:
```
TokenNetwork.openChannel 110848
TokenNetwork.setDeposit 68243
TokenNetwork.closeChannel 106743
TokenNetwork.updateTransfer 67055
TokenNetwork.settleChannel 63017
```

Gas cost after fix:
```
TokenNetwork.openChannel 109253
TokenNetwork.setDeposit 67872
TokenNetwork.closeChannel 106133
TokenNetwork.updateTransfer 66588
TokenNetwork.settleChannel 62907
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+# Config file for automatic testing at travis-ci.org
+
+language: python
+python:
+  - 3.6
+
+# Command to install dependencies, e.g. pip install -r requirements.txt --use-mirrors
+install: pip install -r requirements.txt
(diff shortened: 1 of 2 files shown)


hackaugusto:
I like this change, not because of the optimization, but because to me the code is clearer. Otherwise I would say it's not a good change, since that is the compiler's job.

Issue #10, opened 2018-03-15
Make contracts installable as pip package


pcppcp:
To integrate this repo with other projects, we'll need a working `setup.py` that allows us to install the contracts as a pip package.

### Installable `setup.py`
working with `pip install -e .` and `py setup.py build` targets

### utility library
`setup.py` should also have a contract compilation target that'll build the ABI & bytecode. These should be available with a helper library to other projects. The idea is to get rid of `solc` dependency for all projects except this one.

Quick & dirty example
```
import raiden_contracts

registry_abi = raiden_contracts.get_abi('RegistryContract')
event_abi = raiden_contracts.get_event_abi('RegistryContract', ''SomethingHappenedEvent)
```

palango:
Quick note as I stumbled into this as well: The bytecode of the `TokenNetworkRegistry` with all dependencies required 6283184 gas some days ago. This was to much to deploy it in a single block and we need to link it to the requirements during deployment. I haven't found a nice solution to that yet, but we should keep in mind that just the json created by `populus compile` (which includes all dependencies in the bytecode) is not enough.

pcppcp:
ok then, I will also include a contract manager able to compile contracts using libraries address using `solc`. `solc` won't be a hard requirement, the library simply won't provide such functionality if compiler is not installed. Getting ABI from pre-compiled `contracts.json` will still be available.

pcppcp:
Foxed by #159

PR #12, opened 2018-03-15
add coincurve to requirements

czepluch:
Discovered that `coincurve` was missing when I tried to run the tests.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ eth-keyfile==0.4.1
 eth-keys==0.1.0-beta.4
 web3==4.0.0b4
 click
+coincurve


hackaugusto:
@loredanacirstea , this repo will also include the python utilities? I thought the goal of this repo was to be a language agnostic way to distribute the smart contracts

loredanacirstea:
@hackaugusto 
You are right, so we should have a `requirements-dev.txt` for testing & building. Otherwise, just provide the contracts source code + a variable that exports the contracts.json compiled data. Right?

pcppcp:
From what I understand this repo is intended to be used for development of the smart contracts. This obviously requires a set of tools & testsuite and Python is a obvious choice here. I don't see any point in having a separate repository just for solidity files.

Since several projects need to use smart contracts ABI without duplicating the code, the repository should also provide a simple python library that'll provide an easy way of getting data required to instantiate `Contract` classes from python web3/ethereum libs.
The library must provide
1) a dumb interface for getting ABI/event ABI (it'll just read the compiled `contracts.json` and return whatever it finds there)
2) an optional contract manager capable of using `solc` to compile bytecode, with possibility to use contract libraries. This is required for deploying contracts on tester chains in individual projects (See Paul's comment at #11 ). If `solc` is missing, this functionality won't be available.

loredanacirstea:
This discussion is off-topic here. It should be continued in https://github.com/raiden-network/raiden-contracts/issues/10 and the issue should be close after we settle on what the package should export.

PR #13, opened 2018-03-19
Contract manager

pcppcp:


palango:
I think @loredanacirstea asked this before. Should we just include the compiled contracts by default, so unless we change the contracts we won't need solidity?

loredanacirstea:
I think including the compiled contracts by default would be a good idea. The package should provide everything you need to use the contracts off-the-shelf.

pcppcp:
The problem is that with precompiled contracts you still need `solc` to link two of them together. Compiled bytecode contains strings that are references to other libraries and only when linking these get replaced by an address.

The other option is to link everything as populus does, but then you have this issue with unpractical gas cost for the deployment.

palango:
For running the tests `solc` would be required, but for just running Raiden or the services the ABI should be enough.

raiden_contracts/contract_manager.py
```diff
@@ -0,0 +1,139 @@
...
+            if os.path.basename(x).split('.', 1)[0] == contract_name
+        ]
+
+    def get_mappings(self):
+        """Return dict of mappings to use as solc argument."""
+        return ['%s=%s' % (k, v) for k, v in self.contracts_source_dirs.items()]
+
+    @staticmethod
+    @assert_has_solidity
+    def precompile_contracts(contracts_dir: str, map_dirs: list()) -> dict:
```
hackaugusto: this should not be `list()`

Issue #4, opened 2018-03-13
Add `chain_id` in the balance proof signature


loredanacirstea:
Following the agreed upon specs (https://github.com/raiden-network/spec/blob/master/smart_contracts.rst#balance-proof), `chain_id` must be added to the balance proof.

As mentioned also in https://github.com/raiden-network/raiden/pull/1286#discussion_r173968199

Changes needed to:
- `closeChannel`
- `updateTransfer` & related functions
- `recoverAddressFromSignature`

loredanacirstea:
If we want to use `chain_id`, we need to store it in the contract, because it cannot be accessed from the EVM.

Therefore, the constructor will be:
```
function TokenNetwork(address _token_address, address _secret_registry, uint256 chain_identifier) public
```

This means that you have to be careful when deploying the contract on the different chains.

LefterisJP:
Seems correct to me. We would automate the deployment and the chain id would be filled automatically so I don't think there is any danger there.

loredanacirstea:
Actually registering an independent `TokenNetwork` type contract is a bad idea. It could contain malicious code and we should not add it to the registry unless it was deployed in a controlled manner through `TokenNetworksRegistry.createERC20TokenNetwork`.

LefterisJP:
yeah that is a good point. So let's not provide any means of registering an externally deployed `TokenNetwork` contract. Can you edit your response and strike it through with an explanation?

Issue #2, opened 2018-03-13
See if uint64 for block numbers should be changed to uint256


loredanacirstea:
https://github.com/raiden-network/raiden/pull/1286#discussion_r171927735
> Maybe we should upcast the block number? My idea is to match the block.number type.

Needs to be done:
- see if there are any storage gas cost optimizations from `uint64`. This is the main reason to downcast.

LefterisJP:
I just noticed this issue.

So we actually **should** upcast to `uint256` to save gas -- not the other way around. When solidity works with anything other than `uint256` it has more shifts/other operations in order to make it work with the native uint256 that the EVM works with.

Only if it is part of a struct do we save some gas. Now for the mapping I am not sure -- we would have to check.

loredanacirstea:
Yes, no reason to downcast for mappings.

1)
```
mapping (uint256 => uint256) public channels;
    
    function create(uint256 index) public {
        channels[index] = block.number;
    }
```
`create` - 41634 gas

```
173 SHA3  # from computing the mapping key
174 DUP2
175 SWAP1
176 SSTORE
177 POP
```

2)
```
mapping (uint256 => uint64) public channels;
    
    function create(uint256 index) public {
        channels[index] = uint64(block.number);
    }
```
`create` - 41887 gas

```
201 SHA3  # from computing the mapping key
202 PUSH1 00
204 PUSH2 0100
207 EXP
208 DUP2
209 SLOAD
210 DUP2
211 PUSH8 ffffffffffffffff
220 MUL
221 NOT
222 AND
223 SWAP1
224 DUP4
225 PUSH8 ffffffffffffffff
234 AND
235 MUL
236 OR
237 SWAP1
238 SSTORE
239 POP
```

PR #15, opened 2018-03-21
Add SC Tests/Fixtures and some SC changes

pcppcp:
fixes https://github.com/raiden-network/raiden-contracts/issues/2
fixes https://github.com/raiden-network/raiden-contracts/issues/4

(LC:)
- add more tests and fixtures
- deploy script uses web3 directly; compiled contracts data must be provided (done with `populus compile`

Smart contract changes:
- `TokenNetworkRegistry` -> `TokenNetworksRegistry`
- block numbers are `uint256` instead of `uint64`
- add `chain_id` to `TokenNetwork`, `TokenNetworksRegistry` constructors 
- `event ChannelClosed(uint256 channel_identifier, address closing_address);` (replaced old argument `closing_participant`, because we provide the address of the caller - can be a 3rd party)
- support for closing & updating the channel without any existent off-chain transfers ('nonce' issue: https://github.com/raiden-network/raiden-contracts/pull/15/files#diff-1bdf2ff343918b98864fc26df16cc1ebR276)
- add channel data getters until we are able to return structs in Solidity: 
```
function getChannelInfo(
        uint256 channel_identifier)
        view
        external
        returns (uint256, address, uint256)
    {
        ....
        return (
            channel.settle_timeout,
            closing_request.closing_participant,
            closing_request.settle_block_number
        );
    }

function getChannelParticipantInfo(
        uint256 channel_identifier,
        address participant)
        view
        external
        returns (bool, uint256, uint256, uint64, bytes32)
    {
       ....
        return (
            participant_state.initialized,
            participant_state.deposit,
            participant_state.transferred_amount,
            participant_state.nonce,
            participant_state.locksroot
        );
    }

```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,36 @@
-# Config file for automatic testing at travis-ci.org
+dist: trusty
+sudo: true
+language: generic
 
-language: python
-python:
-  - 3.6
 
-# Command to install dependencies, e.g. pip install -r requirements.txt --use-mirrors
-install: pip install -r requirements.txt
+jobs:
+  include:
+    - language: python
+      if: type = pull_request
+      python: '3.5'
+      env:
+        - TEST_TYPE=core-contracts
+        - SOLC_URL='https://github.com/ethereum/solidity/releases/download/v0.4.19/solc-static-linux'
+        - SOLC_VERSION='v0.4.19'
 
-# Command to run tests, e.g. python setup.py test
-script: true
+      cache:
+        pip: true
+        directories:
+          - $HOME/.cache/pip
+
+      before_install:
+            - mkdir -p $HOME/.bin
+            - export PATH=$PATH:$HOME/.bin
+            - ./.travis/download_solc.sh
+
+      install:
+        - pip install -U pip wheel coveralls "coverage<4.4"
+        - pip install pytest-travis-fold
+        - pip install flake8
+        - pip install -r requirements.txt
+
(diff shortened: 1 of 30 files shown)


hackaugusto:
> event ChannelClosed(uint256 channel_identifier, address closing_address); (replaced old argument closing_participant, because we provide the address of the caller - can be a 3rd party)

Why was this changed? Why is a third party allowed to call close? This should not be the case. Additionally, how can a node tell which direction of the channel was updated first without the closing_participant?

> support for closing & updating the channel without any existent off-chain transfers ('nonce' issue: https://github.com/raiden-network/raiden-contracts/pull/15/files#diff-1bdf2ff343918b98864fc26df16cc1ebR276)

why should **update** allow for a call if there is no data to be provided? what is the end result of this? (close must allow it)

LefterisJP:
I understood it as delegate closing but let's see what @loredanacirstea says.

But if it's 3rd party (i.e. a monitor) closing then that's definitely not something to be allowed.

loredanacirstea:
Naming for event argument reverted https://github.com/raiden-network/raiden-contracts/pull/19. There was no good reason to change this. Logic stayed the same. There is no 3rd party close.

loredanacirstea:
> why should update allow for a call if there is no data to be provided? what is the end result of this? 
fixed: https://github.com/raiden-network/raiden-contracts/pull/20

PR #21, opened 2018-03-22
Update python to 3.6 - travis

loredanacirstea:
Reason: 
- stay in line with Raiden Network project
- support new string formatting such as `f"Event '{event_name}' not found.”`


loredanacirstea:
This was failing, it did not fix travis

LefterisJP:
Arghhh .... so more errors? Well then .. more PRs! -.-

PR #24, opened 2018-03-22
Make populus optional; setup.py adjusts

pcppcp:
changes:
- Make populus optional in the build/installation step
- Move contracts to raiden_contracts (to be able to include sol sources
  in the package). A symlink contracts -> raiden_contracts/contracts
  is used to keep the old directory structure.
- setup.py adjustments to build the package properly


setup.py
```diff
@@ -36,6 +37,7 @@ def run(self):
         if os.path.isfile(COMPILED_CONTRACTS) is False:
             try:
                 self.run_command('build')
+                pass
```
LefterisJP: what is this doing here @pcppcp? Did you mean to write `return`? It is not the same thing.

setup.py
```diff
@@ -36,6 +37,7 @@ def run(self):
...
             except SystemExit:
                 pass
```
LefterisJP: Same here. `pass` and `return` do not have the same effect.

setup.py
pcppcp: `pass` in try block shouldn't be there of course. 
The reason for catching the `SystemExit` is that upon completion, `run_command(build)` internally calls `sys.exit` and the whole thing stops.

PR #25, opened 2018-03-27
Update type annotations

palango:
Some small style and annotation fixes for the contract manager.

- Update and add some type annotations
- Remove a comparison with `is True`
- Add some test cases for non-existant events and contracts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,11 @@ __pycache__
 build/docs/
 .cache
 .pytest_cache
+venv/
 
 # ignore private keys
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -7,6 +7,10 @@ __pycache__
 build/docs/
 .cache
 .pytest_cache
+venv/
 
 # ignore private keys
 *.pem
+
+# editors
+.vscode/
```
pcppcp: add newline

PR #29, opened 2018-03-29
Extract address verification in a separate reusable contract

loredanacirstea:
We can reuse the `ECVerify` library for all projects. It is better to have it separately.
Small refactoring for the address recover code.
Added tests.

### DIFF
--- a/raiden_contracts/contracts/test/SignatureVerifyTest.sol
+++ b/raiden_contracts/contracts/test/SignatureVerifyTest.sol
@@ -1,0 +1,32 @@
+pragma solidity ^0.4.17;
+
+/*
+This is a contract used for testing the ECVerify library and ecrecover behaviour.
+.*/
+
+import "raiden/lib/ECVerify.sol";
+
+contract SignatureVerifyTest {
+    function verify(
+        bytes32 _message_hash,
+        bytes _signed_message)
+        pure
+        public
+        returns (address signer)
+    {
+        // Derive address from signature
+        signer = ECVerify.ecverify(_message_hash, _signed_message);
+    }
+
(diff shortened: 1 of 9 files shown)


raiden_contracts/utils/sign_utils.py
```diff
@@ -3,6 +3,8 @@
 
 
 def sign(privkey: str, msg: bytes, v=0) -> bytes:
+    if isinstance(privkey, bytes):
```
LefterisJP: hmm shouldn't that be an assertion? The type hint expects a `str`. So it's a programming error to provide bytes for the `privkey`

raiden_contracts/tests/test_recover_from_signature.py
```diff
@@ -0,0 +1,126 @@
...
+        balance_proof_hash,
+        r,
+        s,
+        int.from_bytes(v, byteorder='big')
+    )
+    assert address == A
+
+
+def test_ecrecover_output_zero(signature_test_contract, get_accounts, get_private_key):
+    A = get_accounts(1)[0]
```
LefterisJP: Add a reminder here that the zero check you are looking for is due to error (caused by the v parameter)

raiden_contracts/contracts/test/SignatureVerifyTest.sol
```diff
@@ -0,0 +1,32 @@
+pragma solidity ^0.4.17;
+
+/*
+This is a contract used for testing the ECVerify library and ecrecover behaviour.
```
LefterisJP: Make multiline comments with stars like this:

```c
/**
 * Bla bla 
*/
```

PR #31, opened 2018-03-31
last_channel_index is wrong

nkbai:


LefterisJP:
Hello @nkbai . You are indeed correct that the index should be the `channel_identifier` there. Thank you for taking a look at this contract. But please note that the new contracts (the ones in this repository) and especially the unlock mechanism is a complete work in progress and things will change as we develop it.

Regardless thanks for looking and for your fix!

PR #37, opened 2018-04-09
[WIP] MS smart contract

czepluch:
Initial implementation of the Monitoring Service smart contracts in accordance to https://github.com/raiden-network/spec/issues/46.

Still WIP

### DIFF
--- a/raiden_contracts/contracts/MonitoringService.sol
+++ b/raiden_contracts/contracts/MonitoringService.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.17;
+pragma solidity ^0.4.21;
 
 import "./Token.sol";
 import "./Utils.sol";
--- a/raiden_contracts/contracts/MonitoringService.sol
+++ b/raiden_contracts/contracts/MonitoringService.sol
@@ -3,7 +3,6 @@ pragma solidity ^0.4.21;
 import "./Token.sol";
 import "./Utils.sol";
 import "./lib/ECVerify.sol";
-import "./SecretRegistry.sol";
 import "./TokenNetwork.sol";
 
 contract MonitoringService is Utils {
@@ -81,7 +80,8 @@ contract MonitoringService is Utils {
     /// @param _minimum_deposit The minimum value that a MS must deposit to be registered
     function MonitoringService(
         address _token_address,
-        uint256 _minimum_deposit)
+        uint256 _minimum_deposit
+    )
         public
     {
         require(_token_address != 0x0);
@@ -135,14 +135,19 @@ contract MonitoringService is Utils {
     /// being monitored exists.
     function monitor(
         uint256 channel_identifier,
+<<<<<<< HEAD
 		bytes32 balance_hash,
+=======
+        bytes32 balance_hash,
+>>>>>>> update to comply with changes in style guide
         uint64 nonce,
         bytes32 additional_hash,
         bytes closing_signature,
         bytes non_closing_signature,
         address reward_sender_address,
         bytes reward_proof_signature,
-        address token_network_address)
+        address token_network_address
+    )
         isMonitor(msg.sender)
         public
     {
@@ -165,7 +170,7 @@ contract MonitoringService is Utils {
         // Call updateTransfer in the corresponding TokenNetwork
         token_network.updateNonClosingBalanceProof(
             channel_identifier,
-			balance_hash,
+            balance_hash,
             nonce,
             additional_hash,
             closing_signature,
@@ -190,7 +195,8 @@ contract MonitoringService is Utils {
         uint256 channel_identifier,
         address token_network_address,
         uint192 reward_amount,
-        address monitor_address)
+        address monitor_address
+    )
         public
         returns (bool)
     {
--- a/raiden_contracts/contracts/MonitoringService.sol
+++ b/raiden_contracts/contracts/MonitoringService.sol
@@ -135,11 +135,7 @@ contract MonitoringService is Utils {
     /// being monitored exists.
     function monitor(
         uint256 channel_identifier,
-<<<<<<< HEAD
-		bytes32 balance_hash,
-=======
         bytes32 balance_hash,
->>>>>>> update to comply with changes in style guide
         uint64 nonce,
         bytes32 additional_hash,
         bytes closing_signature,
--- a/raiden_contracts/contracts/MonitoringService.sol
+++ b/raiden_contracts/contracts/MonitoringService.sol
@@ -128,7 +128,7 @@ contract MonitoringService is Utils {
     /// is newer than the current newest registered BP.
     /// @param channel_identifier Unique identifier for the channel being monitored in a
     /// specific TokenNetwork.
-    /// @param nonce Strictly monotonic value used to order PBs
+    /// @param nonce Strictly monotonic value used to order BPs
     /// omitting PB specific params, since these will not be provided in the future
     /// @param reward_proof_signature Signature of the Raiden Node signing the reward
     /// @param token_network_address Address of the Token Network in which the channel
--- a/raiden_contracts/contracts/MonitoringService.sol
+++ b/raiden_contracts/contracts/MonitoringService.sol
@@ -9,29 +9,49 @@ import "./TokenNetwork.sol";
 contract MonitoringService is Utils {
 
     Token token; // Only allow RDN?
+
+    // The minimum allowed deposit for a MS
     uint256 minimum_deposit;
-    uint256 minimum_fee;
 
     // keep track of registered Monitoring Services
     // A Monitoring Services must deposit to register
     mapping(address => bool) registered_monitoring_services;
 
     // channel_identifier => Struct
+    // Keep track of the rewards per channel
     mapping(uint256 => Reward) rewards;
 
+    // Deposits made by the MSs
     mapping(address => uint) ms_deposits;
+    // Deposits made by the raiden nodes
     mapping(address => uint) raiden_node_deposits;
 
-    /* Structs */
+    /*
+     *  Structs
+     */
     struct Reward{
+        // The amount of the reward to be paid out to the MS
+        // that provided latest BP
         uint192 reward_amount;
+
+        // Nonce of the most recently provided BP
         uint64 nonce;
+
+        // Address of the monitoring service that provided the most recent BP
         address ms_address;
+
+        // Address of the Raiden Node that was monitored
+        // This is also the address that has the reward deducted from its deposit
         address raiden_node_address;
+
+        // Unique ID to prevent replay attacks on other chains
         uint256 chain_id;
     }
 
-    /* Events */
+    /*
+     *  Events
+     */
+
     event MonitoringServiceNewDeposit(address sender, address receiver, uint amount);
     event RaidenNodeNewDeposit(address sender, address receiver, uint amount);
     event RegisteredMonitoringService(address monitoring_service);
@@ -41,17 +61,27 @@ contract MonitoringService is Utils {
         address ms_address,
         address raiden_node_address
     );
-    event RewardClaimed(address ms_address, uint amount);
+    event RewardClaimed(address ms_address, uint amount, uint256 channel_id);
+
+    /*
+     *  Modifiers 
+     */
 
     modifier isMonitor(address _ms_address) {
         require(registered_monitoring_services[_ms_address]);
         _;
     }
 
+    /*
+     *  Constructor
+     */
+
+    /// @notice Set the default values for the smart contract
+    /// @param _token_address The address of the token to use for rewards
+    /// @param _minimum_deposit The minimum value that a MS must deposit to be registered
     function MonitoringService(
         address _token_address,
-        uint256 _minimum_deposit,
-        uint256 _minimum_fee)
+        uint256 _minimum_deposit)
         public
     {
         require(_token_address != 0x0);
@@ -59,26 +89,32 @@ contract MonitoringService is Utils {
         require(contractExists(_token_address));
 
         minimum_deposit = _minimum_deposit;
-        minimum_fee = _minimum_fee;
 
         token = Token(_token_address);
         // Check if the contract is indeed a token contract
         require(token.totalSupply() > 0);
     }
 
-    function monitorServiceDeposit(address beneficiary, uint amount) returns (bool) {
+    /// @notice Sets the deposit of a Monitoring Service and registers the MS
+    /// Can be called by anyone
+    /// @param beneficiary Address of the MS to be registered 
+    /// (deposit doesn't have to be paid by the MS address)
+    /// @param amount The amount to deposit. Must be higher or equal to minimum_deposit
+    function monitorServiceDeposit(address beneficiary, uint amount) {
         require(amount >= minimum_deposit);
-        // Do we allow several topping up for the MS?
+        // Do we allow topping up for the MS?
         require(ms_deposits[beneficiary] == 0);
 
+        // Add the deposit for the MS_address
         ms_deposits[beneficiary] =+ amount;
 
+        // Transfer the deposit to the smart contract
         require(token.transferFrom(msg.sender, address(this), amount));
 
+        // Register MS
         registered_monitoring_services[beneficiary] = true;
 
         emit MonitoringServiceNewDeposit(msg.sender, beneficiary, amount);
-        //emit RegisteredMonitoringService(beneficiary);
     }
 
     function raidenNodeDeposit(address beneficiary, uint amount) returns (bool) {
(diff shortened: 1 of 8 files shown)


raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+     */
+
+    //event MonitoringServiceNewDeposit(address sender, address receiver, uint amount);
+    event NewDeposit(address sender, address receiver, uint amount);
+    event RegisteredMonitoringService(address monitoring_service);
+    event NewBalanceProofReceived(
+        bytes reward_proof_signature,
+        uint256 nonce,
+        address ms_address,
+        address raiden_node_address
```
loredanacirstea: This is called in `monitor`. I am not entirely sure that this is needed. Where do you use this currently?
If you really need it (e.g. restoring MS state), you can only emit `balance_proof_id` (`keccak256(channel_identifier, token_network_address)`), `ms_address`, because this should be the only thing needed to see what rewards you need to claim.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+    // The minimum allowed deposit for a MS
+    uint256 public minimum_deposit;
+
+    // keep track of registered Monitoring Services
+    // A Monitoring Services must deposit to register
+    mapping(address => bool) public registered_monitoring_services;
+
+    // channel_identifier => Struct
+    // Keep track of the rewards per channel
+    mapping(bytes32 => Reward) rewards;
```
loredanacirstea: `bytes32` should be `keccak256(channel_identifier, token_network_address)`. This contract will be used for all token networks and `channel_identifier` is not enough for uniqueness.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+    //event MonitoringServiceNewDeposit(address sender, address receiver, uint amount);
+    event NewDeposit(address sender, address receiver, uint amount);
+    event RegisteredMonitoringService(address monitoring_service);
+    event NewBalanceProofReceived(
+        bytes reward_proof_signature,
+        uint256 nonce,
+        address ms_address,
+        address raiden_node_address
+    );
+    event RewardClaimed(address ms_address, uint amount, bytes32 channel_id);
```
loredanacirstea: Replace `channel_identifier` with something like `balance_proof_id` - `keccak256(channel_identifier, token_network_address)`

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+    event MonitoringServiceRegistered(address ms_address);
```
loredanacirstea: No need for it.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+    /*
+     *  Constructor
+     */
+
+    /// @notice Set the default values for the smart contract
+    /// @param _token_address The address of the token to use for rewards
+    /// @param _minimum_deposit The minimum value that a MS must deposit to be registered
+    constructor(
+        address _token_address,
+        uint256 _minimum_deposit
```
loredanacirstea: This should be modifiable. Token value will fluctuate.
Drawbacks - adds an `owner` role with some responsibilities, so it must be discussed.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+        token = Token(_token_address);
+        // Check if the contract is indeed a token contract
+        require(token.totalSupply() > 0);
+    }
+
+    /// @notice Deposit tokens used to either pay MSs or to regsiter as MS
+    /// Can be called by anyone several times and on behalf of other accounts
+    /// @param beneficiary The account benefitting from the deposit
+    /// @param amount The amount of tokens to be depositted
+    function deposit(address beneficiary, uint amount) public {
```
loredanacirstea: As mentioned, we should have 2 functions for MS deposits and user deposits.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+        require(amount > 0);
+
+        // This also allows for registered MSs to deposit and use other MSs
+        balances[beneficiary] += amount;
+
+        // Transfer the deposit to the smart contract
+        require(token.transferFrom(msg.sender, address(this), amount));
+
+        emit NewDeposit(msg.sender, beneficiary, amount);
```
loredanacirstea: We don't need the `msg.sender` here.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    /// @notice Deposit and register as Monitoring Service in one function call
+    /// Can be called once and is only callable by addresses not already registered as MS
+    function depositAndRegisterMonitoringService() isNotMonitor(msg.sender) public {
```
loredanacirstea: We should not need this. Depositing as an MS should check the minimum deposit and having a balance in the contract should be enough to be considered registered.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+        require(registerMonitoringService());
+    }
+
+    function updateReward(
+        address token_network_address,
+        address closing_participant,
+        address non_closing_participant,
+        bytes reward_proof_signature,
+        uint256 nonce,
+        address reward_sender_address
```
loredanacirstea: We don't need `reward_sender_address`. This is always the `non_closing_participant`.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+        address reward_sender_address,
+        bytes reward_proof_signature,
+        address token_network_address,
+        address closing_participant,
+        address non_closing_participant
+    )
+        isMonitor(msg.sender)
+        public
+    {
+        updateReward(
```
loredanacirstea: Reward proof checks should be done here + checks that the signature belongs to the `non_closing_participant`.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+    /// @param reward_proof_signature Signature of the Raiden Node signing the reward
+    /// @param token_network_address Address of the Token Network in which the channel
+    /// being monitored exists.
+    function monitor(
+        bytes32 balance_hash,
+        uint256 nonce,
+        bytes32 additional_hash,
+        bytes closing_signature,
+        bytes non_closing_signature,
+        address reward_sender_address,
```
loredanacirstea: Not needed.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+    /// Can be called once per settled channel by everyone on behalf of MS
+    /// @param token_network_address Address of the Token Network in which the channel
+    /// @param reward_amount Amount to be paid as reward to the MS
+    /// @param monitor_address Address of the MS
+    function claimReward(
+        address token_network_address,
+        uint192 reward_amount,
+        address monitor_address,
+        address closing_participant,
+        address non_closing_participant
```
loredanacirstea: `claimReward` should only need the `balance_proof_id`. The contract will know who to send the tokens to.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+        returns (bool)
+    {
+        TokenNetwork token_network = TokenNetwork(token_network_address);
+        bytes32 channel_identifier = token_network.getChannelIdentifier(closing_participant, non_closing_participant);
+
+        // Only allowed to claim, if channel is settled
+        // Channel is settled if it's data has been deleted
+        uint256 channel_settle_block;
+        (, channel_settle_block, ) = token_network.getChannelInfo(closing_participant, non_closing_participant);
+        require(channel_settle_block == 0);
```
loredanacirstea: Now we use `channel.state`. This should be `0`. But yes, `settle_block` can also be used.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+            monitor_address,
+            reward.reward_proof_signature
+        );
+
+        // Deduct reward from raiden_node deposit
+        balances[raiden_node_address] -= reward_amount;
+        // Add reward to the monitoring services' balance.
+        // This minimizes the amount of gas cost
+        // Only use token.transfer in the withdraw function
+        balances[monitor_address] += reward_amount;
```
loredanacirstea: Very good idea.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+
+        // delete storage
+        delete rewards[channel_identifier];
+    }
+
+    /// @notice Withdraw depositted tokens. If msg.sender is MS don't allow to
+    /// withdraw below minimum_deposit.
+    /// Can be called by addresses with a deposit as long as they have a positive balance
+    /// @param amount Amount of tokens to be withdrawn
+    function withdraw(uint amount) public {
```
loredanacirstea: For clarity, I would separate this into 2 functions for MS and for user (Raiden node).

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+            emit MonitoringServiceRegistered(msg.sender);
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    /// @notice Let Monitoring Services deregister as monitoring services.
+    /// Only callable by registered Monitoring Services
+    function deregisterMonitoringService() isMonitor(msg.sender) public {
```
loredanacirstea: I would either make this function also transfer all the tokens to the MS, or include this logic into the `withdraw` for MSs (if MS wants to withdraw all tokens).

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+        internal
+        returns (address signature_address)
+    {
+        bytes32 message_hash = keccak256(
+            channel_identifier,
+            reward_amount,
+            token_network_address,
+            chain_id,
+            nonce,
+            monitor_address
```
loredanacirstea: Should not be included.

raiden_contracts/contracts/MonitoringService.sol
czepluch: We actually had this in one of the first versions, but changed it to one function, but I am fine with either.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,321 @@
...
+
+    // Token to be used for paying the rewards
+    Token public token; // Only allow RDN?
+
+    // The minimum allowed deposit for a MS
+    uint256 public minimum_deposit;
+
+    // keep track of registered Monitoring Services
+    // A Monitoring Services must deposit to register
+    mapping(address => bool) public registered_monitoring_services;
```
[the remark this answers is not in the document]
czepluch: The current design should allow for MS addresses to also be an MS user. This is allowed as long as they have more tokens deposited than the `minimum_deposit`. But the point about separating the two kinds of deposits is valid if we want to use the deposit for the entire Service Bundle.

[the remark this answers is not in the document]
loredanacirstea: Ah, you are right about the current design.

> But the point about separating the two kinds of deposits is valid if we want to use the deposit for the entire Service Bundle.

I am assuming this because we do not have specs for any other contract at this point and the Raiden Bundle needs to deposit anyway + the MS needs to deposit for it to be considered registered.
This being said, I'm also pinging @LefterisJP here, because it seems to be an issue that's not clearly specced out.

raiden_contracts/contracts/MonitoringService.sol
czepluch: It would really be nice to not have an owner, but it's a good point that I didn't really consider. I guess an ideal solution would include an oracle that could update the value according to the price every day, but that would also make a mess, if an MS then needs to deposit more tokens to be re-registered. We need to think about this indeed.

pcppcp: my idea was to have an auction contract that would reset the registry each N blocks and the MSs would be re-elected depending on deposit they've put in.

raiden_contracts/contracts/MonitoringService.sol
pcppcp: this will be probably used when multiple MSs can update the channel - if the event is emited, it no longer makes sense to call `monitor`. emitting just the hash is ok 
(but I am not sure about costs - isn't `keccak256` more expensive than plain `emit`?)

loredanacirstea: logs are `375` + `8` for each argument if not indexed. `keccak256` is 30 + `6` for each argument.

raiden_contracts/contracts/MonitoringService.sol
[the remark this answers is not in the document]
LefterisJP: Yes this is something we need to discuss and spec further. The RSB idea requires a single deposit per operator of a Bundle. So there should be a different contract holding the deposits. How would that looks remains TBD. Made this issue to track the discussion: https://github.com/raiden-network/spec/issues/79

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,298 @@
...
+
+    // Raiden Service Bundle contract to use for checking if MS has deposits
+    RaidenServiceBundle public rsb;
+
+    // keccak256(channel_identifier, token_network_address) => Struct
+    // Keep track of the rewards per channel
+    mapping(bytes32 => Reward) rewards;
+
+    // Keep track of balances
+    mapping(address => uint) public balances;
```
loredanacirstea: `uint` -> `uint256` for consistency

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,298 @@
...
+
+        token = Token(_token_address);
+        rsb = RaidenServiceBundle(_rsb_address);
+        // Check if the contract is indeed a token contract
+        require(token.totalSupply() > 0);
+        // Check if the contract is indeed an rsb contract
+        // TODO: Check that some function exists in the contract
+    }
+
+    /// @notice Deposit tokens used to either pay MSs or to regsiter as MS
```
loredanacirstea: comment fix: `or to regsiter as MS` - done in the RSB contract now

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,298 @@
...
+    /// Can be called by anyone several times and on behalf of other accounts
+    /// @param beneficiary The account benefitting from the deposit
+    /// @param amount The amount of tokens to be depositted
+    function deposit(address beneficiary, uint amount) public {
+        require(amount > 0);
+
+        // This also allows for MSs to deposit and use other MSs
+        balances[beneficiary] += amount;
```
loredanacirstea: In the TokenNetwork contract we have decided to use an idempotent `setDeposit` function, which receives the total amount of tokens that should end up as deposited in the contract, instead of the amount that is added when calling the function. The main reason for this was to ensure that you don't `deposit` tokens by mistake, due to an additional API call.
I think we should be consistent and also use this approach here.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,298 @@
...
+        require(token.transferFrom(msg.sender, address(this), amount));
+
+        emit NewDeposit(beneficiary, amount);
+    }
+
+    /// @notice Internal function that updates the Reward struct if a newer balance proof
+    /// is provided in the monitor() function
+    /// @param token_network_address Address of the TokenNetwork being monitored
+    /// @param closing_participant The address of the closing participant
+    /// @param non_closing_participant Address of the other channel participant
```
loredanacirstea: You could add that this is the participant on whom's behalf the MS acts / who signed the reward proof.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,298 @@
...
+        // Transfer the deposit to the smart contract
+        require(token.transferFrom(msg.sender, address(this), amount));
+
+        emit NewDeposit(beneficiary, amount);
+    }
+
+    /// @notice Internal function that updates the Reward struct if a newer balance proof
+    /// is provided in the monitor() function
+    /// @param token_network_address Address of the TokenNetwork being monitored
+    /// @param closing_participant The address of the closing participant
```
loredanacirstea: Something like `Address of the participant who closed the channel` may be clearer.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,298 @@
...
+        // Make sure that the reward proof is signed by the non_closing_participant
+        address raiden_node_address = recoverAddressFromRewardProof(
+            channel_identifier,
+            reward_amount,
+            token_network_address,
+            token_network.chain_id(),
+            nonce,
+            reward_proof_signature
+        );
+        require(raiden_node_address == non_closing_participant);
```
loredanacirstea: Doesn't need changes in this PR, but keep in mind this https://github.com/raiden-network/raiden-contracts/issues/47 - code style decision of keeping the checks grouped. If directly related to the input, they should be first. Otherwise, they can be added last. This requires attention to not change a variable between the operation that you want to check and the actual check - so there are some drawbacks...

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,298 @@
...
+
+        // delete storage
+        delete rewards[reward_identifier];
+    }
+
+    /// @notice Withdraw depositted tokens.
+    /// Can be called by addresses with a deposit as long as they have a positive balance
+    /// @param amount Amount of tokens to be withdrawn
+    function withdraw(uint amount) public {
+        require(balances[msg.sender] - amount >= 0);
```
loredanacirstea: This does not protect from underflows, we need a `require(balances[msg.sender] >= amount);`

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -0,0 +1,298 @@
...
+        );
+
+        // delete storage
+        delete rewards[reward_identifier];
+    }
+
+    /// @notice Withdraw depositted tokens.
+    /// Can be called by addresses with a deposit as long as they have a positive balance
+    /// @param amount Amount of tokens to be withdrawn
+    function withdraw(uint amount) public {
```
loredanacirstea: `uint` -> `uint256`

PR #38, opened 2018-04-09
Bump solc version

pcppcp:


czepluch:
@pcppcp can you either close this or update to newer version?

loredanacirstea:
Just be aware that we are not upgrading yet anyway. Version 0.4.22 is not stable.

czepluch:
0.4.23 isn't stable either?

LefterisJP:
We can get rid of a lot of warning by upgrading but 0.4.22 had a lot of problems so perhaps let's wait a few days before going to 0.4.23

palango:
This was fixed by #57 as well.

Issue #49, opened 2018-04-26
Implement cooperativeSettle as per specifications


loredanacirstea:
https://github.com/raiden-network/spec/blob/master/smart_contracts.rst

```
function cooperativeSettle(
    uint channel_identifier,
    address participant1,
    address participant2,
    uint256 balance1,
    uint256 balance2,
    bytes signature1,
    bytes signature2)
    public
```

PR #50, opened 2018-04-27
Cooperative settle

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/49

Adds support for an instant channel close & settle if both participants exchange signatures on balances calculated off-chain.

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -576,10 +576,13 @@ contract TokenNetwork is Utils {
 
     function cooperativeSettle(
         uint256 channel_identifier,
+        address participant1_address,
         uint256 participant1_balance,
+        address participant2_address,
         uint256 participant2_balance,
         bytes participant1_signature,
-        bytes participant2_signature)
+        bytes participant2_signature
+    )
         public
     {
         address participant1;
@@ -589,14 +592,18 @@ contract TokenNetwork is Utils {
 
         participant1 = recoverAddressFromCooperativeSettleSignature(
             channel_identifier,
+            participant1_address,
             participant1_balance,
+            participant2_address,
             participant2_balance,
             participant1_signature
         );
 
         participant2 = recoverAddressFromCooperativeSettleSignature(
             channel_identifier,
+            participant1_address,
             participant1_balance,
+            participant2_address,
             participant2_balance,
             participant2_signature
         );
@@ -605,6 +612,10 @@ contract TokenNetwork is Utils {
         Participant storage participant2_state = channel.participants[participant2];
 
         total_deposit = participant1_state.deposit + participant2_state.deposit;
+
+        // The provided addresses must be the same as the recovered ones
+        require(participant1 == participant1_address);
+        require(participant2 == participant2_address);
 
         // The channel must be open
         require(channel.state == 1);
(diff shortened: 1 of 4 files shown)


LefterisJP:
Starting review of this.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -766,6 +827,31 @@ contract TokenNetwork is Utils {
         signature_address = ECVerify.ecverify(message_hash, non_closing_signature);
     }
 
+    function recoverAddressFromCooperativeSettleSignature(
```
LefterisJP: Spec question: Shouldn't the format of this signature be in the spec?

raiden_contracts/tests/test_channel_cooperative_settle.py
```diff
@@ -0,0 +1,409 @@
...
+
+    balance_A_fail1 = 20
+    balance_B_fail1 = 11
+    balance_A_fail2 = 6
+    balance_B_fail2 = 8
+
+    # Create channel and deposit
+    channel_identifier = create_channel_and_deposit(A, B, deposit_A, deposit_B)
+
+    # Create cooperative settle signatures
```
LefterisJP: pedantic: No reason for the comment as the function is very aptly named :)

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -766,6 +827,31 @@ contract TokenNetwork is Utils {
...
+        uint256 channel_identifier,
+        address participant1,
+        uint256 participant1_balance,
+        address participant2,
```
LefterisJP: From chat: 

> some context about me adding the addresses in the signature (not only the balance values) is - only having the balance values there in an unenforced order, can make it confusing to determine/check whose is the balance.

Why would it make it confusing? First balance would be participant1's balance and second balance participant2 balance.

raiden_contracts/contracts/TokenNetwork.sol
loredanacirstea: Yes. I will make a spec PR after this is merged. This was something that we did not discuss before.

raiden_contracts/tests/test_channel_cooperative_settle.py
loredanacirstea: Yes, removing them. The comments were there first. Then came the rest.

raiden_contracts/contracts/TokenNetwork.sol
loredanacirstea: Yes, but in the Raiden client side: - you need to choose an order first and then send it to the other participant to sign. This is were I see issues - you have to pay extra care that the node receiving this message to sign knows how to interpret the balances correctly. 
I think this format brings more clarity.

LefterisJP: > This is were I see issues - you have to pay extra care that the node receiving this message to sign knows how to interpret the balances correctly.

Each node will know its balance and the balance of its counterparty so I still don't see the scenario where having the addresses in the signature is really required.

Would like to hear @hackaugusto's opinion on this too.

> I think this format brings more clarity.

True, but this signature is intended to be made and consumed by a raiden client, so human readable clarity is not a goal here.

LefterisJP: I agree. Changes to the signature if needed can happen in another PR. Whenever @hackaugusto becomes available we can see what his opinion is.

hackaugusto: I'm fine with either. having the participant addresses is a bit more explicit wich is nice

PR #57, opened 2018-05-03
Remove pyethereum dependency

palango:
This should be merged after raiden-network/raiden-libs#64 to get clean dependencies.


raiden_contracts/contract_manager.py
```diff
@@ -54,14 +35,24 @@ def assert_has_solidity(func):
     """decorator - Raise CompileError if HAS_SOLIDITY is set to False"""
     @wraps(func)
     def func_wrap(*args, **kwargs):
-        if _solidity is None:
-            raise _solidity.CompileError(
-                'solc not found in $PATH. Check your path or set $SOLC_BINARY.'
-            )
+        # if _solidity is None:
```
palango: This needs to be replaced.

raiden_contracts/contract_manager.py
```diff
@@ -84,12 +75,10 @@ def __init__(self, path: Union[str, List[str]]) -> None:
     @assert_has_solidity
     def compile_contract(self, contract_name: str, libs=None, *args):
```
palango: Are these functions used anywhere?

pcppcp:
solc update, great! I'll take a look at the CM/compilation test tomorrow.

pcppcp:
I think having a separate venv for `ContractManager` tests is the only option now. Theoretically it should work with current `requirements.txt` file.

palango:
Yeah, next problem is the `compile_contract` function. I'm not able to get unlinked bytecode, so testing won't work.

raiden_contracts/contract_manager.py
```diff
@@ -81,16 +60,19 @@ def __init__(self, path: Union[str, List[str]]) -> None:
...
-            self.get_contract_path(contract_name)[0],
-            contract_name,
-            combined='abi,bin',
-            libraries=libs,
-            extra_args=self.get_mappings()
+        contract_json = compile_files(
+            [self.get_contract_path(contract_name)[0]],
+            output_values=('abi', 'bin'),
+            import_remappings=self.get_mappings(),
+            optimize=True
```
loredanacirstea: Please set to false. We have decided to not use the optimizer for the moment.

raiden_contracts/contract_manager.py
```diff
@@ -125,18 +106,22 @@ def precompile_contracts(contracts_dir: str, map_dirs: List) -> Dict:
...
+        try:
+            res = compile_files(
+                files,
+                output_values=('abi', 'bin'),
+                import_remappings=map_dirs
             )
-        return ret
+            return fix_contract_key_names(res)
+        except FileNotFoundError:
+            raise Exception('Could not compile the contract. Check that solc is avialable.')
```
loredanacirstea: `avialable` -> `available`

Issue #51, opened 2018-04-27
Review token contracts used in tests


loredanacirstea:
These contracts were mostly ported from the old raiden contracts, they are only used in testing.

However, there are some weird implementations in there. E.g.:

https://github.com/raiden-network/raiden-contracts/blob/81698bb48c94d0a1be556d8b062d316f1e0fc083/raiden_contracts/contracts/test/StandardToken.sol#L19

https://github.com/raiden-network/raiden-contracts/blob/81698bb48c94d0a1be556d8b062d316f1e0fc083/raiden_contracts/contracts/test/StandardToken.sol#L31

PR #59, opened 2018-05-07
Remove populus

pcppcp:
- this makes tests indenpendent on populus library


requirements.txt
```diff
@@ -1,4 +1,5 @@
-ethereum
 click
-coincurve
-web3>=4.0.0-beta.4
+rlp<1.0.0
+coincurve>=7.1.0,<8.0
+ethereum>=2.3.1,<3.0
```
palango: is this still used in the tests? or just necessary for the stuff that is fixed in #57?

raiden_contracts/contracts/test/StandardToken.sol
```diff
@@ -11,7 +11,7 @@ import "raiden/Token.sol";
 
 contract StandardToken is Token {
     uint256 internal _total_supply;
-    mapping (address => uint256) balances;
+    mapping (address => uint256) public balances;
```
loredanacirstea: This is actually not needed, because we have `balanceOf` and this is the correct interface to use. Adding another public interface is confusing.
Anyway, the token contracts are messed up, so you can leave it like this - opened https://github.com/raiden-network/raiden-contracts/issues/51 for them.

requirements.txt
pcppcp: I've used versions from `raiden-libs` and I am happy the tests pass finally :) 

We can probably rebase #57 on top of this PR when it gets to master and update the versions accordingly.

raiden_contracts/contracts/test/StandardToken.sol
pcppcp: ah, you are right, this should not be here. I was just playing with the contract source and forgot to revert to the original version.

requirements.txt
palango: Hehe, I get that. Updating it in #57 is fine for me.

raiden_contracts/tests/fixtures/secret_registry.py
```diff
@@ -1,16 +1,6 @@
...
-        contract = create_contract(SecretRegistry, arguments, transaction)
-        return contract
-    return get
-
-
-@pytest.fixture()
-def secret_registry(get_secret_registry):
-    return get_secret_registry([])
+def secret_registry(secret_registry_contract):
+    return secret_registry_contract
```
loredanacirstea: Where is `secret_registry_contract` defined?

requirements.txt
```diff
@@ -1,4 +1,5 @@
...
+web3>=4.0.0b1
```
loredanacirstea: You can use `web3>=4.0.0`.
Tests fail for `web3 < 4.0.0b13` (e.g. `4.0.0b11`) .

raiden_contracts/tests/fixtures/secret_registry.py
pcppcp: it's part of fixtures defined in `raiden-libs` and used for testing in PFS & MS.

https://github.com/raiden-network/raiden-libs/blob/master/raiden_libs/test/fixtures/contracts.py#L93

(well it'd probably make more sense to remove this redefinition and use original fixture instead)

requirements.txt
pcppcp: okay I'll create an issue to update the version in libs and other projects as well

loredanacirstea:
Just to mention - I am still reviewing this PR.

loredanacirstea:
Aside from the `token_network` fixture, things seems good. There is still the rebase to be done at this point.
The only thing I have some issues with is having fixtures that are in another repo. This can potentially lead to issues if the fixtures are changed in some subtle way that change the tests logic.
But as long as we keep in sync about `raiden-libs` fixture changes, it can work.

raiden_contracts/tests/fixtures/secret_registry.py
loredanacirstea: Can also be done in another PR. I don't have anything against renaming it to `secret_registry_contract` though.

palango:
It would make sense to move the fixtures into the contracts repo now that they are used here as well.

loredanacirstea:
Yes, it makes more sense to have them in the contracts repo, agreed.

raiden_contracts/tests/test_print_gas.py
```diff
@@ -8,46 +8,39 @@
...
-        owner,
-        token_network_registry,
-        custom_token,
-        secret_registry,
-        print_gas
+    web3,
+    deploy_tester_contract_txhash,
+    secret_registry_contract,
+    custom_token,
+    print_gas
```
loredanacirstea: Indentation from style guide ..

raiden_contracts/tests/test_token_network_registry.py
```diff
@@ -55,63 +56,52 @@ def test_constructor_call_state(chain, get_token_network_registry, secret_regist
...
 
     token_network_registry.transact().createERC20TokenNetwork(custom_token.address)
 
 
-def test_create_erc20_token_network(chain, web3, token_network_registry, custom_token):
+def test_create_erc20_token_network(
+    register_token_network,
+    token_network_registry,
+    custom_token,
+    get_accounts
```
loredanacirstea: Indentation from style guide.

pcppcp:
I'll move the fixtures from libs to contracts, but let's do that in another PR, because it'll break practically everything. :)

Issue #53, opened 2018-04-30
unlocked_amount may be 0 too.


nkbai:
```
function unlock(
        uint256 channel_identifier,
        address participant,
        address partner,
        bytes merkle_tree_leaves)
        public

        // Transfer the unlocked tokens to the participant
        require(token.transfer(participant, unlocked_amount));

        // Transfer the rest of the tokens back to the partner
        if (returned_tokens > 0) {
            require(token.transfer(partner, returned_tokens));
        }
```
this code assumes that `returned_tokens` maybe 0,and `unlocked_amount` must  greater than 0。
for example ,the caller (msg.sender) maybe `partner`(argument), and all locks are locked.
in this condition, `returned_tokens` equals `nonce_or_locked_amount`。

and some ERC20 token doesn't allow transfer 0 token.

loredanacirstea:
@nkbai , thank you for reporting! 
This will be fixed in https://github.com/raiden-network/raiden-contracts/pull/64

Issue #47, opened 2018-04-25
Assess keeping invariants grouped


loredanacirstea:
(must be updated with a proper description)

As mentioned in this thread: https://github.com/raiden-network/raiden-contracts/pull/39#discussion_r183729507

> keeping the invariants grouped? I think it is easier to verify them if we don't need to jump around in the contract code.
> IMO this should not take priority over clarity if the input is invalid, only in cases where the input is valid but a race happened (e.g. two monitoring services called update at the same time, only one of the transactions will work)

LefterisJP:
Just my 2 cents. I also would really like to have all the require invariants somewhere. This would make each function so much easier to read and reason about.

loredanacirstea:
Yes, I thought about this and I think I went too far with making sure the gas cost is minimal.
So, I would stick to using `require` strictly for input values (as the official Solidity docs recommend anyway) and move all other invariants in `assert`s at the bottom of the function.
Does this sound good?

LefterisJP:
Grouping the invariants is fine but you should only use asserts for internal errors and things that should never ever happen. I am commenting on #64 about it. 

Docs which help us understand the different between `require()` and `assert(): http://solidity.readthedocs.io/en/v0.4.23/control-structures.html?highlight=assert#error-handling-assert-require-revert-and-exceptions

hackaugusto:
Just to clarify my point, code should be written to be readable first. I don't have a hard rule for how to format code for readability, specially because of solidity's constraints on the number of variables in a stack, which require intermediary functions to be introduced, which may hide validation inside these functions.

But I would strive for a contract-based approach, where the beginning of a function starts with the requirements, first validating/requiring the value inputs, then the code is followed by the computation, which is followed by the validation of the computation and external side-effects, and the code finishes with the asserts/invariants that must be true after the execution of the code.

loredanacirstea:
> But I would strive for a contract-based approach, where the beginning of a function starts with the requirements, first validating/requiring the value inputs, then the code is followed by the computation, which is followed by the validation of the computation and external side-effects, and the code finishes with the asserts/invariants that must be true after the execution of the code.

I fully agree with this. 
In my opinion this is how you get more secure code & more readable.
Especially more secure code. When you group all the post-state-change (computation) checks to the end of the function it is way harder to make sure that that variables did not change their state in between the computation and the check. Sometimes you need to change them and you end up needing to store the initial state in another local variable (very ugly).

[...]

Issue #45, opened 2018-04-25
Add overflow and underflow checks for all operations


loredanacirstea:
Make sure we have `assert`s for all smart contracts operations (e.g. `+`, `-`, `*`)

E.g. for `settleChannel` , `getSettleTransferAmounts`

```
        total_deposit_available = (
            participant1_deposit +
            participant2_deposit -
            participant1_locked_amount -
            participant2_locked_amount
        );

        // Overflow and underflow checks
        require(total_deposit_available <= (participant1_deposit + participant2_deposit));
        require(total_deposit_available > 0);

        participant1_amount = (
            participant1_deposit +
            participant2_transferred_amount -
            participant1_transferred_amount
        );

        // Overflow and underflow checks
        require(participant1_amount <= participant1_deposit + participant2_transferred_amount);
        require(participant1_amount > ?);
```

Issue #32, opened 2018-03-31
Fix nonce check in closeChannel


loredanacirstea:
As mentioned in https://github.com/raiden-network/raiden-contracts/commit/0634fa0c4d4dc41092716f1d1f1bc27c0b58b75c#r28362284,  `nonce` check must be done before trying to recover `partner_address`. Otherwise, the transaction will fail.

https://github.com/raiden-network/raiden-contracts/blob/3bcb0e40e5f044e45ce8b2969b39b6f2ef5ae2dd/raiden_contracts/contracts/TokenNetwork.sol#L276

PR #64, opened 2018-05-09
Core contracts - add & group invariant checks

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/45 (overflow and underflow checks)
fixes https://github.com/raiden-network/raiden-contracts/issues/32 (fix nonce check in closeChannel)
fixes https://github.com/raiden-network/raiden-contracts/issues/47 (group invariants)
fixes https://github.com/raiden-network/raiden-contracts/issues/53 (`unlock_amount` can be `0`)

Mostly groups and adds invariant checks + some tests to make sure they work as assumed.

### DIFF
--- a/raiden_contracts/contracts/SecretRegistry.sol
+++ b/raiden_contracts/contracts/SecretRegistry.sol
@@ -23,7 +23,7 @@ contract SecretRegistry {
     /// @return true if secret was registered, false if the secret was already registered.
     function registerSecret(bytes32 secret) public returns (bool) {
         bytes32 secrethash = keccak256(secret);
-        if (secrethash_to_block[secrethash] > 0) {
+        if (secret == 0x0 || secrethash_to_block[secrethash] > 0) {
             return false;
         }
         secrethash_to_block[secrethash] = block.number;
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -190,18 +190,18 @@ contract TokenNetwork is Utils {
         Channel storage channel = channels[channel_identifier];
         Participant storage participant_state = channel.participants[participant];
 
-        require(participant_state.deposit < total_deposit);
-
         // Calculate the actual amount of tokens that will be transferred
         added_deposit = total_deposit - participant_state.deposit;
 
         // Update the participant's channel deposit
         participant_state.deposit += added_deposit;
 
-        // Do the transfer
-        require(token.transferFrom(msg.sender, address(this), added_deposit));
-
         emit ChannelNewDeposit(channel_identifier, participant, participant_state.deposit);
+
+        // Do the transfer
+        assert(token.transferFrom(msg.sender, address(this), added_deposit));
+
+        assert(participant_state.deposit >= added_deposit);
     }
 
     /// @notice Close the channel defined by the two participant addresses. Only a participant
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -137,7 +137,7 @@ contract TokenNetwork is Utils {
         chain_id = _chain_id;
 
         // Make sure the contract is indeed a token contract
-        assert(token.totalSupply() > 0);
+        require(token.totalSupply() > 0);
     }
 
     /*
@@ -472,8 +472,8 @@ contract TokenNetwork is Utils {
         participant1_amount = max(participant1_amount - participant1_locked_amount, 0);
         participant2_amount = max(participant2_amount - participant2_locked_amount, 0);
 
-        assert(participant1_amount <= total_deposit);
-        assert(participant2_amount <= total_deposit);
+        require(participant1_amount <= total_deposit);
+        require(participant2_amount <= total_deposit);
         assert(total_deposit == (
             participant1_amount +
             participant2_amount +
@@ -544,8 +544,8 @@ contract TokenNetwork is Utils {
 
         require(computed_locksroot != 0);
         require(locked_amount > 0);
+        require(locked_amount >= returned_tokens);
         assert(locked_amount >= unlocked_amount);
-        assert(locked_amount >= returned_tokens);
     }
 
     function cooperativeSettle(
@@ -615,7 +615,7 @@ contract TokenNetwork is Utils {
         require(participant2 == participant2_address);
 
         // The sum of the provided balances must be equal to the total deposit
-        assert(total_deposit == (participant1_balance + participant2_balance));
+        require(total_deposit == (participant1_balance + participant2_balance));
     }
 
     /// @dev Returns the unique identifier for the channel
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -326,11 +326,6 @@ contract TokenNetwork is Utils {
         assert(non_closing_participant == recovered_non_closing_participant);
     }
 
-    /// @notice Registers the lock secret in the SecretRegistry contract.
-    function registerSecret(bytes32 secret) public {
-        require(secret_registry.registerSecret(secret));
-    }
-
     /// @notice Settles the balance between the two parties.
     /// @param participant1 Channel participant.
     /// @param participant1_transferred_amount The latest known amount of tokens transferred
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -133,11 +133,11 @@ contract TokenNetwork is Utils {
 
         token = Token(_token_address);
 
-        // Check if the contract is indeed a token contract
-        require(token.totalSupply() > 0);
-
         secret_registry = SecretRegistry(_secret_registry);
         chain_id = _chain_id;
+
+        // Make sure the contract is indeed a token contract
+        assert(token.totalSupply() > 0);
     }
 
     /*
@@ -477,6 +477,16 @@ contract TokenNetwork is Utils {
         participant1_amount = max(participant1_amount - participant1_locked_amount, 0);
         participant2_amount = max(participant2_amount - participant2_locked_amount, 0);
 
+        assert(total_deposit >= participant1_deposit);
+        assert(participant1_amount <= total_deposit);
+        assert(participant2_amount <= total_deposit);
+        assert(total_deposit == (
+            participant1_amount +
+            participant2_amount +
+            participant1_locked_amount +
+            participant2_locked_amount
+        ));
+
         return (participant1_amount, participant2_amount);
     }
 
@@ -502,10 +512,6 @@ contract TokenNetwork is Utils {
 
         channel_identifier = getChannelIdentifier(participant, partner);
 
-        // Channel must be settled and channel data deleted
-        require(channels[channel_identifier].state == 0);
-
-
         // Calculate the locksroot for the pending transfers and the amount of tokens
         // corresponding to the locked transfers with secrets revealed on chain.
         (computed_locksroot, unlocked_amount) = getMerkleRootAndUnlockedAmount(merkle_tree_leaves);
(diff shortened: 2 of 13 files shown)


LefterisJP:
Review this PR.

raiden_contracts/contracts/SecretRegistry.sol
```diff
@@ -23,7 +23,7 @@ contract SecretRegistry {
     /// @return true if secret was registered, false if the secret was already registered.
     function registerSecret(bytes32 secret) public returns (bool) {
         bytes32 secrethash = keccak256(secret);
-        if (secrethash_to_block[secrethash] > 0) {
+        if (secret == 0x0 || secrethash_to_block[secrethash] > 0) {
```
LefterisJP: even though this is not part of this PR, just wanna mention we should make this fail with a revert as per [this](https://github.com/raiden-network/spec/issues/76#issuecomment-387373440) issue.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -190,18 +190,18 @@ contract TokenNetwork is Utils {
...
         added_deposit = total_deposit - participant_state.deposit;
 
         // Update the participant's channel deposit
         participant_state.deposit += added_deposit;
 
+        emit ChannelNewDeposit(channel_identifier, participant, participant_state.deposit);
+
         // Do the transfer
-        require(token.transferFrom(msg.sender, address(this), added_deposit));
+        assert(token.transferFrom(msg.sender, address(this), added_deposit));
```
LefterisJP: This should be a require(), not an assert. As per solidity's [docs](http://solidity.readthedocs.io/en/v0.4.23/control-structures.html?highlight=assert#error-handling-assert-require-revert-and-exceptions):

```
The require function should be used to ensure valid conditions, such as inputs, or contract state variables are met, or to validate return values from calls to external contracts.
```

Specifically note the `validate return values from calls to external contracts`

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -190,18 +190,18 @@ contract TokenNetwork is Utils {
...
 
-        emit ChannelNewDeposit(channel_identifier, participant, participant_state.deposit);
+        assert(participant_state.deposit >= added_deposit);
```
LefterisJP: also a require() here, not an assert. In general assert should be for something that should never ever happen under any circumstances, i.e.: internal errors.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -235,24 +235,23 @@ contract TokenNetwork is Utils {
...
+                signature
+            );
+
             updateBalanceProofData(channel, recovered_partner_address, nonce, balance_hash);
-        }
 
-        // Signature must be from the channel partner
-        assert(partner == recovered_partner_address);
+            // Signature must be from the channel partner
+            assert(partner == recovered_partner_address);
```
LefterisJP: this should be a `require()`, it depends on external input.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -315,23 +308,24 @@ contract TokenNetwork is Utils {
...
-        // Make sure the first signature is from the closing participant
-        require(closing_participant_state.is_the_closer);
-
         // Update the balance proof data for the closing_participant
         updateBalanceProofData(channel, closing_participant, nonce, balance_hash);
 
         emit NonClosingBalanceProofUpdated(channel_identifier, closing_participant);
 
+        // Channel must be closed
+        assert(channel.state == 2);
```
LefterisJP: should be a require(), depends on external input (time of call of the function)

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -315,23 +308,24 @@ contract TokenNetwork is Utils {
...
+
+        // Channel must be in the settlement window
+        assert(channel.settle_block_number >= block.number);
```
LefterisJP: same as above

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -315,23 +308,24 @@ contract TokenNetwork is Utils {
...
+
+        // Make sure the first signature is from the closing participant
+        assert(closing_participant_state.is_the_closer);
```
LefterisJP: should be require()

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -531,19 +527,25 @@ contract TokenNetwork is Utils {
         // Remove partner's unlock data
         delete unlock_data.locksroot_to_locked_amount[computed_locksroot];
 
-        // Transfer the unlocked tokens to the participant
-        require(token.transfer(participant, unlocked_amount));
+        // Transfer the unlocked tokens to the participant. unlocked_amount can be 0
+        if (unlocked_amount > 0) {
+            assert(token.transfer(participant, unlocked_amount));
```
LefterisJP: should be a require()

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -531,19 +527,25 @@ contract TokenNetwork is Utils {
...
+        }
 
         // Transfer the rest of the tokens back to the partner
         if (returned_tokens > 0) {
-            require(token.transfer(partner, returned_tokens));
+            assert(token.transfer(partner, returned_tokens));
```
LefterisJP: should be a require()

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -587,31 +591,31 @@ contract TokenNetwork is Utils {
...
         delete channel.participants[participant1];
         delete channel.participants[participant2];
         delete channels[channel_identifier];
 
+        emit ChannelSettled(channel_identifier);
+
         // Do the token transfers
         if (participant1_balance > 0) {
-            require(token.transfer(participant1, participant1_balance));
+            assert(token.transfer(participant1, participant1_balance));
```
LefterisJP: should be a require()

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -587,31 +591,31 @@ contract TokenNetwork is Utils {
...
         }
 
         if (participant2_balance > 0) {
-            require(token.transfer(participant2, participant2_balance));
+            assert(token.transfer(participant2, participant2_balance));
```
LefterisJP: should be a require()

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -587,31 +591,31 @@ contract TokenNetwork is Utils {
...
         }
 
-        emit ChannelSettled(channel_identifier);
+        // The channel must be open
+        assert(initial_state == 1);
```
LefterisJP: should be a require() as it depends on external input (time of call of the function)

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -587,31 +591,31 @@ contract TokenNetwork is Utils {
...
+
+        // The provided addresses must be the same as the recovered ones
+        assert(participant1 == participant1_address);
```
LefterisJP: should be a require() as it depends on external input (provided address)

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -587,31 +591,31 @@ contract TokenNetwork is Utils {
...
+        assert(participant2 == participant2_address);
```
LefterisJP: should be a require() as it depends on external input (provided address)

raiden_contracts/tests/fixtures/channel.py
```diff
@@ -61,13 +69,95 @@ def get(participant, deposit, partner):
...
+
+@pytest.fixture()
+def create_settled_channel(
+        web3,
+        token_network,
+        create_channel_and_deposit,
+        create_balance_proof,
+        create_balance_proof_update_signature
+):
+    def get(participant1, locked_amount1, locksroot1, participant2, locked_amount2, locksroot2, settle_timeout=SETTLE_TIMEOUT_MIN):
```
LefterisJP: line too long, should split it

raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -64,6 +66,141 @@ def test_merkle_root(
...
+        pending_transfers_tree_A.merkle_root,
+        B,
+        0,
+        fake_bytes(32),
+        settle_timeout
+    )
+
+    # This should pass, even though the locked amount in storage is bigger.
+    # But those locked tokens will remain locked in the contract
+    token_network.transact().unlock(
```
LefterisJP: perhaps here (and in the next test) also add a check for the token balance of the contract to check they are still locked there?

LefterisJP:
Also added a tiny commit fixing a style thing and wording of a comment (make sure to pull)

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -478,6 +472,15 @@ contract TokenNetwork is Utils {
         participant1_amount = max(participant1_amount - participant1_locked_amount, 0);
         participant2_amount = max(participant2_amount - participant2_locked_amount, 0);
 
+        assert(participant1_amount <= total_deposit);
```
LefterisJP: Until this morning I would have left those as asserts as it "should not happen". But I had a talk with Christian today and showed him these lines. 

He said these should still be `require()`. Overflow here can happen if the appropriate input is given to this function. The only way this should be an assert is if we actually somehow in the code path that leads here have already checked that no overflow can happen. Iff we already did then this should be an assert here. If not a require.

The thinking is that there will be automatic tools that could be testing the program for correctness, feeding it with random input. Or trying to verify correctness through formal proofing tools. Asserts should never ever be reached under any input or conditions.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -541,9 +539,13 @@ contract TokenNetwork is Utils {
...
 
-        assert(computed_locksroot != 0);
-        assert(unlocked_amount > 0);
-        assert(locked_amount > 0);
+        // Channel must be settled and channel data deleted
+        require(channels[channel_identifier].state == 0);
+
+        require(computed_locksroot != 0);
+        require(locked_amount > 0);
+        assert(locked_amount >= unlocked_amount);
```
LefterisJP: Just let me see if I understand correctly. This here can be an `assert` since at [this](https://github.com/raiden-network/raiden-contracts/pull/64/files#diff-2f6fb78ac7aaf29aaec1ed8ab59468f8R522) line we already make sure that `unlocked_amount` is the min of `unlocked_amount` and `locked_amount` so this should always hold true.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -541,9 +539,13 @@ contract TokenNetwork is Utils {
...
+        assert(locked_amount >= returned_tokens);
```
LefterisJP: If this is an overflow check for [this](https://github.com/raiden-network/raiden-contracts/pull/64/files#diff-2f6fb78ac7aaf29aaec1ed8ab59468f8R522) line: `returned_tokens = locked_amount - unlocked_amount;`, then with the logic explained above it should still be a `require()`

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -611,7 +607,15 @@ contract TokenNetwork is Utils {
...
-        emit ChannelSettled(channel_identifier);
+        // The channel must be open
+        require(initial_state == 1);
+
+        // The provided addresses must be the same as the recovered ones
+        require(participant1 == participant1_address);
+        require(participant2 == participant2_address);
+
+        // The sum of the provided balances must be equal to the total deposit
+        assert(total_deposit == (participant1_balance + participant2_balance));
```
LefterisJP: With the same logic as above, this overflow check should be a `require()`

Issue #65, opened 2018-05-09
Implement withdraw functionality for TokenNetwork


loredanacirstea:
Implement `withdraw` functionality. Described in https://github.com/raiden-network/spec/issues/26

PR #67, opened 2018-05-11
TokenNetwork - add support for withdrawing tokens

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/65

Add support for withdrawing tokens without closing the channel.
Also discussed here: https://github.com/raiden-network/spec/issues/26


raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -204,6 +210,76 @@ contract TokenNetwork is Utils {
...
+
+        // Channel must be open
+        require(channel.state == 1);
+
+        // Check recovered addresses from signatures
+        require(participant == recovered_participant_address);
+        require(partner == recovered_partner_address);
+
+        // Underflow check
+        require(participant_state.withdrawn_amount >= amount_to_withdraw);
```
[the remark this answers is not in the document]
loredanacirstea: `participant_state.withdrawn_amount == amount_to_withdraw` when you first withdraw.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -204,6 +210,76 @@ contract TokenNetwork is Utils {
...
+
+        bytes32 channel_identifier;
+        address recovered_participant_address;
+        address recovered_partner_address;
+        uint256 total_deposit;
+
+        channel_identifier = getChannelIdentifier(participant, partner);
+        Channel storage channel = channels[channel_identifier];
+
+        recovered_participant_address = recoverAddressFromWithdrawMessage(
```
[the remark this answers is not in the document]
loredanacirstea: Yes, I will use the same approach as in `setDeposit` - monotonically increasing `amount_to_withdraw` in the signed data.

raiden_contracts/tests/fixtures/channel_test_values.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+ChannelValues = namedtuple('ChannelValues', [
+    'deposit',
+    'withdrawn',
+    'transferred',
+    'locked'
+])
+
+channel_test_vals = [
```
LefterisJP: In the codebase we try not to use abbrevitions, especially if they don't save much space. Can you change `vals` to `values`?

raiden_contracts/tests/test_channel_settle.py
```diff
@@ -46,107 +47,62 @@ def test_settle_no_bp_success(
...
-        channel_deposit,
-        get_accounts,
-        get_block,
-        create_balance_proof,
-        create_balance_proof_update_signature
+        create_channel_and_deposit,
+        withdraw_channel,
+        close_and_update_channel,
+        settle_state_tests,
+        channel_test_values
```
LefterisJP: So from what I understand and by reading [this](https://github.com/raiden-network/raiden-contracts/pull/67/files#diff-155b1944c9c54cfb5ecab176aefe584dR11) line I see this is a way to parametrize the test with different values.

It's not really readable, and forces me to go back and read the fixture's definition and then to follow the code upwards to see what's happening. I suggest not having it in a separate fixture that you use only once, but instead to use `pytest.mark.parametrize()` like we do in Raiden. For example [here](https://github.com/raiden-network/raiden/blob/master/raiden/tests/unit/test_binary_encoding.py#L80)

raiden_contracts/tests/test_channel_withdraw.py
```diff
@@ -0,0 +1,436 @@
...
+from raiden_contracts.utils.events import check_withdraw
+from .fixtures.config import empty_address, fake_bytes
+from raiden_contracts.utils.config import (
+    E_CHANNEL_WITHDRAW,
+    CHANNEL_STATE_NONEXISTENT_OR_SETTLED,
+    CHANNEL_STATE_OPEN,
+    CHANNEL_STATE_CLOSED,
+    SETTLE_TIMEOUT_MIN
+)
+from .utils import MAX_UINT256
```
LefterisJP: Maybe not for this PR, but we tend to not use relative imports at all. You use them quite a bit. Can you at some point change them to full path imports? e.g. for here: `from raiden_contracts.test.utils import MAX_UINT256` ?

raiden_contracts/tests/test_channel_withdraw.py
```diff
@@ -0,0 +1,436 @@
...
+
+    (signature_A_for_A, signature_B_for_A) = create_withdraw_signatures(
+        [A, B],
+        channel_identifier,
+        A, withdraw_A
+    )
+
+    with pytest.raises(ValidationError):
+        token_network.transact({'from': A}).withdraw(
+            0x0, withdraw_A,
```
LefterisJP: in this and the rest of the `transact()` why are the first 2 arguments grouped together? Style guide is to keep them all in their own line. Is there a semantic reason you want them together?

raiden_contracts/tests/test_channel_withdraw.py
```diff
@@ -0,0 +1,436 @@
...
+
+    balance_A = custom_token.call().balanceOf(A)
+    balance_B = custom_token.call().balanceOf(B)
+    balance_C = custom_token.call().balanceOf(C)
+    balance_contract = custom_token.call().balanceOf(token_network.address)
+
+    withdraw_channel(B, withdraw_B, A)
+
+    withdraw_state_tests(
+        B, deposit_B, withdraw_B, 0, balance_B,
```
LefterisJP: Same thing with the arguments grouping here.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -204,6 +210,70 @@ contract TokenNetwork is Utils {
...
+        );
+
+        // Underflow check
+        require(participant_state.withdrawn_amount >= current_withdraw);
+        require(participant_state.withdrawn_amount == total_withdraw);
+
+        // Entire withdrawn amount must not be bigger than the entire channel deposit
+        require(participant_state.withdrawn_amount <= (total_deposit - partner_state.withdrawn_amount));
+
+        assert(total_deposit >= participant_state.deposit);
```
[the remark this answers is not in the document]
loredanacirstea: Yes, but two overflow checks that should never fail. It's not related to user input either. I think it should be an `assert`

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -204,6 +210,70 @@ contract TokenNetwork is Utils {
...
+        assert(total_deposit >= partner_state.deposit);
+        assert(participant_state.nonce == 0);
```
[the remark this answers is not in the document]
loredanacirstea: `nonce > 0` happens only if the channel participant has a balance proof stored. This happens when closing the channel. 
I know we already have a `state` check. This is an additional check that should never fail. May be redundant, but as a general rule, we should not allow token withdrawal if we already have balance proofs stored - in case we change the protocol somehow or there's a bug.

raiden_contracts/tests/fixtures/channel_test_values.py
loredanacirstea: Used that name because I called the fixture `channel_test_values`, but as you suggested in another comment, I'll use `pytest.mark.parametrize()` and also change the name.

raiden_contracts/tests/test_channel_withdraw.py
loredanacirstea: In case of files from the same directory it seems a bit too much to restrict to full paths. Why is this?

LefterisJP: It's I would say 50% a style choice and 50% being future-proof for if you move the file itself out of the directory the imports would still work.

raiden_contracts/tests/test_channel_withdraw.py
loredanacirstea: I used this way of writing only when I had a combination of short symbols / numbers (that did not took a lot of space; e.g. did not like to use an entire line for an `A`) + wanted to separate values for `A` and `B`. But I can go with the code style if needed.

raiden_contracts/tests/test_channel_withdraw.py
loredanacirstea: Same for me - we changed folder structure a couple of times and relative paths seemed better because of this.
But I didn't know that full paths were the chosen style for Raiden, so I'll change.

LefterisJP: Thank you. Does not need to be in this PR, it's already quite bit. Just remember it for later.

raiden_contracts/tests/test_channel_withdraw.py
LefterisJP: Style guide asks for the 2nd for consistency. But in general as you said we should avoid having so many arguments in functions.

I can see why you would like to group them like in the first example though.

raiden_contracts/contracts/TokenNetwork.sol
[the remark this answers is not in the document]
LefterisJP: I see. So yes from looking at the code again this can only happen if `updateBalanceProof()` is called, which is done only during `closeChannel()`. There is a already a check for the channel state up there, requiring it to be open.

So this can never happen. So can be an assert. Funny how asserts can only be redundant checks....

raiden_contracts/contracts/TokenNetwork.sol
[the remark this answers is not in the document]
loredanacirstea: I view it here as something that should never happen at this level. But, I get your point and I am also not entirely convinced of either, so I'll change.

PR #69, opened 2018-05-13
Use 'bin' as the source of a contracts bytecode

rakanalh:
Hello,

While trying to run contract deployment i came across the fact that `contracts.json` will contain `bin` instead of `bytecode`. Which matches the requested `output_values` here: https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/contract_manager.py#L67

My solc version is as follows:
```
solc, the solidity compiler commandline interface
Version: 0.4.23+commit.124ca40d.Linux.g++
```


loredanacirstea:
Thank you! After https://github.com/raiden-network/raiden-contracts/pull/59 we probably have some outdated scripts.

PR #74, opened 2018-05-15
Remove `contracts` symlink

pcppcp:
- it was only required for populus build; now it can go


LefterisJP:
LGTM, let's wait till tests pass.

PR #78, opened 2018-05-18
Freeze solc to 2.1.0

pcppcp:
- temporary fix until we make 3.0.0 work

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ click
 rlp>=1.0.0
 coincurve>=7.1.0,<8.0
 web3>=4.0.0
-py-solc
+py-solc==2.1.0


requirements.txt
```diff
@@ -2,4 +2,4 @@ click
 rlp>=1.0.0
 coincurve>=7.1.0,<8.0
 web3>=4.0.0
-py-solc
\ No newline at end of file
+py-solc==2.1.0
```
czepluch: Title says 2.2.0, but change is 2.1.0. Which one is the intended correct version?

pcppcp: `2.1.0` actually, `2.2.0` also doesn't work. I'll rename the PR.

PR #79, opened 2018-05-18
Fix contract compilation with `py-solc` >= 2.2.0

pcppcp:
- adds `ast` output to compile options to make `py-solc` happy

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ click
 rlp>=1.0.0
 coincurve>=7.1.0,<8.0
 web3>=4.0.0
-py-solc==2.1.0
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -2,4 +2,4 @@ click
 rlp>=1.0.0
 coincurve>=7.1.0,<8.0
 web3>=4.0.0
-py-solc==2.1.0
+py-solc
```
LefterisJP: I would still suggest to have the version specified here.

pcppcp: sounds good, I'll make `3.0.0` a minimum then

PR #80, opened 2018-05-18
Idempotency of MS deposits - closes #76

### LINKED ISSUE #76: Use idempotent deposit function in MS contract?
Comment from Loredana:

> In the TokenNetwork contract we have decided to use an idempotent setDeposit function, which receives the total amount of tokens that should end up as deposited in the contract, instead of the amount that is added when calling the function. The main reason for this was to ensure that you don't deposit tokens by mistake, due to an additional API call.
I think we should be consistent and also use this approach here.

See how it's done in the TokenNetwork contract and implement similar functionality in the MS contract.

---

edit by @pirapira: In MSC and UDC, currently, `deposit(totalDeposit)` can cause more-than-expected deposits.  This `totalDeposit` pattern works in `TokenNetwork` because it has only-ever-increasing `totalDeposit`.  In MSC and UDC, `balance[user]` might decrease as well.

err508:


raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -90,18 +90,25 @@ contract MonitoringService is Utils {
...
+    /// @param total_deposit Idempotent function that sets the total_deposit of
+    //// tokens of the beneficiary
+    function deposit(address beneficiary, uint256 total_deposit) public
+    {
+        require(total_deposit > 0);
+
+        uint256 added_deposit;
+
+        // Calculate the actual amount of tokens that will be transferred
+        added_deposit = total_deposit - balances[beneficiary];
```
czepluch: This is not good. uint256 can only be positive integers, and if the new `total_deposit` is lower than the current balance you have a problem. You must add a check that doesn't allow for this to happen. Ideally you also add a small test to show that it works correctly.

err508: Changed.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -90,18 +90,27 @@ contract MonitoringService is Utils {
 
     /// @notice Deposit tokens used to reward MSs
     /// Can be called by anyone several times and on behalf of other accounts
-    /// @param beneficiary The account benefitting from the deposit
-    /// @param amount The amount of tokens to be depositted
-    function deposit(address beneficiary, uint amount) public {
-        require(amount > 0);
+    /// @param beneficiary The account benefiting from the deposit
+    /// @param total_deposit Idempotent function that sets the total_deposit of
```
czepluch: This should explain what the parameter is, not what the function does. The `@notice` is meant for explaining what the function does.

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -90,18 +90,27 @@ contract MonitoringService is Utils {
...
+        balances[beneficiary] += added_deposit;
+
+        emit NewDeposit(beneficiary, added_deposit);
 
         // Transfer the deposit to the smart contract
-        require(token.transferFrom(msg.sender, address(this), amount));
+        require(token.transferFrom(msg.sender, address(this), added_deposit));
 
-        emit NewDeposit(beneficiary, amount);
+        require(balances[beneficiary] >= added_deposit);
```
czepluch: Couldn't we cover both `require(total_deposit != 0);` and `require(balances[beneficiary] >= added_deposit);` by doing `total_deposit > balances[beneficary]` instead?

raiden_contracts/contracts/MonitoringService.sol
```diff
@@ -90,18 +90,27 @@ contract MonitoringService is Utils {
...
-    /// @param beneficiary The account benefitting from the deposit
-    /// @param amount The amount of tokens to be depositted
-    function deposit(address beneficiary, uint amount) public {
-        require(amount > 0);
+    /// @param beneficiary The account benefiting from the deposit
+    /// @param total_deposit Idempotent function that sets the total_deposit of
+    //// tokens of the beneficiary
+    function deposit(address beneficiary, uint256 total_deposit) public
+    {
+        require(total_deposit != 0);
```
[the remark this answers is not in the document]
loredanacirstea: Ah, ok. I understand what you mean. But I see it as: the scope is to make sure  `total_deposit  > 0`. Because Solidity is still a young language that can still behave in unwanted ways, I would not completely trust it especially if it doesn't really cost us much. Plus, this `> 0` pattern was agreed upon in a previous smart contracts review session for uRaiden. I would rather keep it like this for now.

raiden_contracts/contracts/MonitoringService.sol
loredanacirstea: I used the pattern that was also used here in TokenNetwork.setDeposit because I was also trying to keep the checks grouped, as agreed. In this case, `require(total_deposit > balances[beneficary])` can work.

raiden_contracts/contracts/MonitoringService.sol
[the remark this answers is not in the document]
err508: Changed it back to the way it was. I was referring to UINT(-1) above, but doesn't make a difference I think.

raiden_contracts/contracts/MonitoringService.sol
err508: I think `require(total_deposit > balances[beneficiary])` solves under/overflow in a short way,  not sure though which conventions Loredana refers too.

loredanacirstea: @err508 https://github.com/raiden-network/raiden-contracts/issues/47. In `TokenNetwork.setDeposit` it wasn't so straight forward to get the participant deposit and do the check at the beginning of the function, so I opted for a check at the end. As said, for this case here, `require(total_deposit > balances[beneficiary])` does the job.

czepluch: As Loredana says, `require(total_deposit > balances[beneficiary])` does the trick and we get the same checks guarantees from one line rather than two. So please use that instead.

Issue #86, opened 2018-05-25
Fix `keccak256` and fallback function breaking changes for solc 0.5.0


loredanacirstea:
The following 0.5.0 breaking changes currently affect the contracts:

- fallback functions must be `external` (https://github.com/ethereum/solidity/blob/develop/Changelog.md#0417-2017-09-21)
- `keccak256` can only be used with a single `bytes` argument; `abi.encodePacked` can be used to tightly pack arguments (https://github.com/ethereum/solidity/blob/develop/Changelog.md#0424-2018-05-16)

LefterisJP:
Can you provide a link to all the features of solidity 0.5.0 that you want to account for? With 0.5.0 not yet released we can't really consider this issue complete.

loredanacirstea:
@LefterisJP , copied the PR description here.

PR #87, opened 2018-05-25
Fix upcoming solc 0.5.0 breaking changes

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/86

- fallback functions must be `external` (https://github.com/ethereum/solidity/blob/develop/Changelog.md#0417-2017-09-21)
- `keccak256` can only be used with a single `bytes` argument; `abi.encodePacked` can be used to tightly pack arguments (https://github.com/ethereum/solidity/blob/develop/Changelog.md#0424-2018-05-16)

### DIFF
--- a/raiden_contracts/contracts/EndpointRegistry.sol
+++ b/raiden_contracts/contracts/EndpointRegistry.sol
@@ -71,7 +71,7 @@ contract EndpointRegistry{
 
     function equals(string a, string b) internal pure returns (bool result)
     {
-        if (keccak256(a) == keccak256(b)) {
+        if (keccak256(abi.encodePacked(a)) == keccak256(abi.encodePacked(b))) {
             return true;
         }
 
--- a/raiden_contracts/contracts/SecretRegistry.sol
+++ b/raiden_contracts/contracts/SecretRegistry.sol
@@ -22,7 +22,7 @@ contract SecretRegistry {
     /// @param secret The secret used to lock the hash time lock.
     /// @return true if secret was registered, false if the secret was already registered.
     function registerSecret(bytes32 secret) public returns (bool) {
-        bytes32 secrethash = keccak256(secret);
+        bytes32 secrethash = keccak256(abi.encodePacked(secret));
         if (secret == 0x0 || secrethash_to_block[secrethash] > 0) {
             return false;
         }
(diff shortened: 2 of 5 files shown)


LefterisJP:
What about:

> Enforce view / constant with error as experimental 0.5.0 feature. 

and

> Disallow conversions between bytesX and uintY of different size


All taken from [here](https://github.com/ethereum/solidity/blob/develop/Changelog.md#050-unreleased).

Not saying you need to do something in this PR, just wanna know if you have them in mind/checked for them.

raiden_contracts/contracts/EndpointRegistry.sol
```diff
@@ -71,7 +71,7 @@ contract EndpointRegistry{
 
     function equals(string a, string b) internal pure returns (bool result)
     {
-        if (keccak256(a) == keccak256(b)) {
+        if (keccak256(abi.encodePacked(a)) == keccak256(abi.encodePacked(b))) {
```
LefterisJP: Did this also throw a warning? It's not multiple arguments, but just one argument. Is it because it's `string` and not `bytes`?

raiden_contracts/contracts/SecretRegistry.sol
```diff
@@ -22,7 +22,7 @@ contract SecretRegistry {
     /// @param secret The secret used to lock the hash time lock.
     /// @return true if secret was registered, false if the secret was already registered.
     function registerSecret(bytes32 secret) public returns (bool) {
-        bytes32 secrethash = keccak256(secret);
+        bytes32 secrethash = keccak256(abi.encodePacked(secret));
```
LefterisJP: Hmm here also it's 1 single argument and it's also bytes

loredanacirstea:
@LefterisJP , there were no other compiler errors aside from the fixed ones.

> Enforce view / constant with error as experimental 0.5.0 feature.

We already use `view` and `pure` correctly.

> Disallow conversions between bytesX and uintY of different size

We don't have this.

raiden_contracts/contracts/SecretRegistry.sol
loredanacirstea: secret is `bytes32`. It expects `bytes`. So yes, compiler gave an error here.

PR #101, opened 2018-05-31
Raiden integration: update ChannelManager

pcppcp:
- fixes exception that is being raised when CM is initialized with a  single directory as its constructor argument
- better filtering of files that will be compiled
- removes duplicate token fixture

### DIFF
--- a/raiden_contracts/tests/fixtures/token.py
+++ b/raiden_contracts/tests/fixtures/token.py
@@ -39,6 +39,6 @@ def deploy_token_contract(deploy_tester_contract):
 
 
 @pytest.fixture
-def standard_token_contract(deploy_token_contract):
-    """Deployed HumanStandardToken contract"""
-    return deploy_token_contract(1000000, 10, 'TT', 'TTK')
+def standard_token_contract(custom_token):
+    """Deployed CustomToken contract"""
+    return custom_token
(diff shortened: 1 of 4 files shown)


raiden_contracts/tests/fixtures/token.py
```diff
@@ -39,6 +39,6 @@ def f(initial_amount: int, decimals: int, token_name: str, token_symbol: str):
 
 
 @pytest.fixture
-def standard_token_contract(deploy_token_contract):
+def standard_token_contract(custom_token):
```
loredanacirstea: Is this actually needed then? We already have `custom_token` and `deploy_token_contract` (for `HumanStandardToken`). This is now just a duplicate for the `custom_token` with a misleading description.

pcppcp: MS and Pathfinder still use this fixture, we can remove it later once their code is updated.

raiden_contracts/tests/fixtures/token.py
```diff
@@ -39,6 +39,6 @@ def f(initial_amount: int, decimals: int, token_name: str, token_symbol: str):
...
     """Deployed HumanStandardToken contract"""
```
loredanacirstea: Then maybe fix the description to say `"""Deployed CustomToken contract"""`, so it's not confusing.

PR #104, opened 2018-06-04
Adds basic unit test for endpoint registry SC. Closes #84

err508:

Issue #89, opened 2018-05-27
Add more test values for the channel settle token amounts.


loredanacirstea:
Add more cases to `channel_test_values.py` - testing channel settle token amounts.

LefterisJP:
Perhaps this is where you could use something like property testing to test ranges of values?

loredanacirstea:
Added values manually for known edge cases in
https://github.com/raiden-network/raiden-contracts/commit/8decf565ff1098662c94b148cdb1b041aba6ee19#diff-155b1944c9c54cfb5ecab176aefe584d
https://github.com/raiden-network/raiden-contracts/commit/bf56d07c9207f7330e6c169d491cf560647e36df#diff-155b1944c9c54cfb5ecab176aefe584d

This can be closed. We can open other issues for  adding more settlement calculations tests.

PR #105, opened 2018-06-04
Adds a fixture to generate random test values for channel settle. 

err508:


loredanacirstea:
Please add a proper title for what the PR is and a short description. Also, please put the issue link in the description, because it's a pain to search for the issue instead of just clicking a link.

loredanacirstea:
Also, branch the PR from `master`, to not include already reviewed code in the diff. This branch has nothing to share with the branch from https://github.com/raiden-network/raiden-contracts/pull/104

err508:
Related to Issue #89

raiden_contracts/tests/fixtures/channel_test_values_random.py
```diff
@@ -0,0 +1,45 @@
...
+                )
+                for withdrawn, transferred, deposit, locked in
+                [sorted(randint(0, maxsize) for _ in range(4)) for _ in range(n)]
+            ]),
+            False
+        )
+    ]
+
+
+channel_settle_test_values_random = channel_settle_test_values_random(100)
```
loredanacirstea: Where is this used? Is there a test using this?

raiden_contracts/tests/fixtures/channel_test_values_random.py
```diff
@@ -0,0 +1,45 @@
...
+        (
+            tuple([
+                ChannelValues(
+                    deposit=deposit,
+                    withdrawn=withdrawn,
+                    transferred=transferred,
+                    locked=locked
+                )
+                for withdrawn, transferred, deposit, locked in
+                [sorted(randint(0, maxsize) for _ in range(4)) for _ in range(n)]
```
loredanacirstea: As mentioned, it is hard to understand what you are trying to achieve here. Either use readable code or make sure to comment. The two cases look similar and you have to pay attention to see that the `withdrawn, transferred, deposit, locked` values are actually reordered (though I don't really understand why).

`[sorted(randint(0, maxsize) for _ in range(4)) for _ in range(n)]` seems like you are choosing sets of 4 sorted random values. In the `True` case, you will always have `deposit <= withdrawn <= transferred <= locked`, which is not really useful and not correct. Am I wrong?

PR #111, opened 2018-06-06
Fixes getSettleTransferAmounts when total_deposit = 0.

### LINKED ISSUE #106: Make sure channel can be settled and storage deleted
There are cases in which `settleChannel` fails. E.g. see https://github.com/raiden-network/raiden-contracts/blob/58e56d642edbbecdc2177b3477178d9f766c2c21/raiden_contracts/tests/test_channel_settle.py#L55

```
channel_settle_test_values = [
    (
        ChannelValues(deposit=40, withdrawn=10, transferred=26, locked=6),
        ChannelValues(deposit=35, withdrawn=5, transferred=24, locked=4),
        True  # settleChannel transaction should pass
    ),
    (
        ChannelValues(deposit=10, withdrawn=10, transferred=26, locked=6),
        ChannelValues(deposit=5, withdrawn=5, transferred=24, locked=4),
        False  # settleChannel transaction should fail
    )
]
```

`settleChannel` fails if the balance proofs where somehow incorrect and all the tokens have already been withdrawn - second case above. 

TODO:
- document what cases trigger `settleChannel` to fail
- decide what to do in those cases - I would propose to not let the transaction fail, but make sure tokens are transferred correctly and just remove the channel data if there are no more tokens to be transferred. If the channel cannot be settled, then it cannot be reopened and used at a later time, because it remains in `state == 2` (closed)

err508:
Fixes #106

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -552,22 +552,25 @@ contract TokenNetwork is Utils {
 
         // At this point `participant1_amount` is between [0, total_available_deposit],
         // so this is safe.
-        participant2_amount = total_available_deposit - participant1_amount;
 
+        participant2_amount = total_available_deposit - participant1_amount;
         // Handle old balance proofs with a high locked_amount
         participant1_amount = max(participant1_amount - participant1_locked_amount, 0);
         participant2_amount = max(participant2_amount - participant2_locked_amount, 0);
 
-        require(participant1_amount <= total_available_deposit);
-        require(participant2_amount <= total_available_deposit);
-        assert(total_available_deposit == (
-            participant1_amount +
-            participant2_amount +
-            participant1_locked_amount +
-            participant2_locked_amount
-        ));
-
-        return (participant1_amount, participant2_amount);
+        if (participant1_amount <= total_available_deposit &&
+            participant2_amount <= total_available_deposit
+            ) {
+            assert(total_available_deposit == (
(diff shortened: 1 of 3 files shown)


raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -552,22 +552,24 @@ contract TokenNetwork is Utils {
...
-        assert(total_available_deposit == (
-            participant1_amount +
-            participant2_amount +
-            participant1_locked_amount +
-            participant2_locked_amount
-        ));
-
-        return (participant1_amount, participant2_amount);
+        if (participant1_amount <= total_available_deposit &&
+            participant1_amount <= total_available_deposit
```
czepluch: your mean `participant2_amount`, right?

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -552,22 +552,24 @@ contract TokenNetwork is Utils {
...
+            ) {
+            assert (total_available_deposit ==
+                participant1_amount +
+                participant2_amount +
+                participant1_locked_amount +
+                participant2_locked_amount);
+                return (participant1_amount, participant2_amount);
```
czepluch: One too many indents here.

raiden_contracts/contracts/TokenNetwork.sol
err508: yes, thx for the catch.

PR #112, opened 2018-06-07
Fix settlement calculations

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/106

We must first calculate the amount corresponding to the participant who did NOT close the channel.
Channel settlement was refactored and the above requirement was (very) wrongly discarded. 
There have been previous discussions on the matter, the most important conclusion being https://github.com/raiden-network/raiden/issues/278#issuecomment-265177676

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -91,12 +91,18 @@ contract TokenNetwork is Utils {
         uint256 settle_timeout
     );
 
-    event ChannelNewDeposit(bytes32 indexed channel_identifier, address indexed participant, uint256
-    total_deposit);
+    event ChannelNewDeposit(
+        bytes32 indexed channel_identifier,
+        address indexed participant,
+        uint256 total_deposit
+    );
 
     // withdrawn_amount is the total amount withdrawn by the participant from the channel
-    event ChannelWithdraw(bytes32 indexed channel_identifier, address indexed participant, uint256
-    total_withdraw);
+    event ChannelWithdraw(
+        bytes32 indexed channel_identifier,
+        address indexed participant,
+        uint256 total_withdraw
+    );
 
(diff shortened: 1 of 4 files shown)


raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -531,43 +547,64 @@ contract TokenNetwork is Utils {
...
+        }
+        else {
+            non_closing_locked_amount = non_closing_amount;
+            non_closing_amount = 0;
+        }
 
-        require(participant1_amount <= total_available_deposit);
-        require(participant2_amount <= total_available_deposit);
+        require(closing_amount <= total_available_deposit);
+        require(non_closing_amount <= total_available_deposit);
```
loredanacirstea: I would make this an `assert` , because it should never happen and it's not testing something from user input. It is testing `non_closing_amount = min(non_closing_amount, total_available_deposit);`

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -531,43 +547,64 @@ contract TokenNetwork is Utils {
...
+            non_closing_amount = non_closing_amount - non_closing_locked_amount;
+        }
+        else {
+            non_closing_locked_amount = non_closing_amount;
+            non_closing_amount = 0;
+        }
 
-        require(participant1_amount <= total_available_deposit);
-        require(participant2_amount <= total_available_deposit);
+        require(closing_amount <= total_available_deposit);
```
loredanacirstea: I would make this an `assert` also , because it should never happen and it's not testing something from user input. It is testing 
```
non_closing_amount = min(non_closing_amount, total_available_deposit);
closing_amount = total_available_deposit - non_closing_amount;
```

loredanacirstea:
Closing this in favor of https://github.com/raiden-network/raiden-contracts/pull/114.
Order of calculation (closer, non-closer) does not matter. The problem is now explained in the smart contract comments from the above PR.

PR #116, opened 2018-06-08
Settle overflows.

err508:
overflow attack vector (test_settlement_overflow) still applies on #114


hackaugusto:
Did we add these tests?

PR #120, opened 2018-06-08
Use defined contract and event names in tests

### LINKED ISSUE #115: Use constants.py contract & event names in the tests

palango:
Fixes #115


palango:
Fixed all other usages of the contract or event names. @loredanacirstea

Issue #117, opened 2018-06-08
overflow/underflow mitigation


err508:
When reviewing `TokenNetwork.sol`, some overflow/underflow issues arose. In this issue we should discuss how to adapt and fix, w.r.t to typing and testing. Potential mitigations paths are:

i) changing from UINT to INT, as suggested by @hackaugusto 
This would allow to do something like `require(x>=0)`, helping to mitigate the underflow vectors.

ii) Use SafeMaths `add` and `sub`:
https://github.com/OpenZeppelin/openzeppelin-solidity/blob/6e66ba321e545c7840f51ba978a0dd9aaad3ad99/contracts/math/SafeMath.sol#L27-L31

ii)Decrease contract internal `type_size` to `UINT128` to do the computations in. Add a modulo logic in `INT256` to give the  contract the space to represent/deal with overflows/underflows. This would also decrease gas cost for storage but must be evaluated and formalized first. 

iv) Add a corresponding testing environment that mimics the contract behaviour, to check that local and contract-internal computation is identical.

err508:
#114 #116 # [raiden/#1529](https://github.com/raiden-network/raiden/issues/1529)

hackaugusto:
> i) changing from UINT to INT, as suggested by @hackaugusto

This is not necessary, we are using uint256 because that is what is used by the ERC20 token standard, converting back and forth from these two types would be probably a bad idea.

loredanacirstea:
> i) changing from UINT to INT

Same - opens a whole other set of issues - compatibility with token standards, conversion issues between `uint` and `int`, max `int` value lower than max `uint` value etc.
http://solidity.readthedocs.io/en/v0.4.24/types.html#conversions-between-elementary-types

> ii) Use SafeMath

SafeMath would only force the transaction to fail. We don't want that in `settleChannel`, for example - this is where we currently have issues. In other functions, we can have proper checks in place. I'm not against using SafeMath for all other functions except `settleChannel`, but last time this was discussed for uRaiden, we settled on not using additional libraries. We can reopen this discussion.

> ii)Decrease contract internal type_size to UINT128 to do the computations in. Add a modulo logic in INT256

- decrease size for what variables? It would help to have a proposed implementation, to see exactly what you have in mind.

LefterisJP:
Why not add the checks ourselves where needed like we did in uRaiden and in the token sale contract where under/overflow checks are needed?

err508:
[...]

PR #123, opened 2018-06-11
Fix settlement overflow on transferred amount

loredanacirstea:
This should fix the overflow issue for the old `deposit + transferred_amount` approach, demonstrated in https://github.com/raiden-network/raiden-contracts/pull/116/files#diff-cf6314cb143087ba6788de239e94298aR7 and mentioned in the thread started here: https://github.com/raiden-network/raiden-contracts/issues/117#issuecomment-395758666 

TODO:
(update: the following is not necessary - detailed explanation in https://github.com/raiden-network/raiden-contracts/pull/123#pullrequestreview-128277844)
- if possible, create test that would trigger an underflow when subtracting the withdrawn amount, to test that we handle it correctly; think some more if this approach is fair.
```
        // Subtract already withdrawn amount
        (participant1_amount, ) = failsafe_subtract(
            participant1_amount,
            participant1_state.withdrawn_amount
        );
```

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -550,7 +550,7 @@ contract TokenNetwork is Utils {
         private
         returns (uint256, uint256, uint256, uint256)
     {
-        // Cases require attention:
+        // Cases that require attention:
         // case1. If participant1 does NOT provide a balance proof or provides an old balance proof.
         // participant2_transferred_amount can be [0, real_participant2_transferred_amount)
         // We need to punish participant1.
@@ -564,7 +564,7 @@ contract TokenNetwork is Utils {
         // we have the following check in settleChannel:
         // require(participant2_transferred_amount >= participant1_transferred_amount)
 
-        uint256 participant1_netted_transferred_amount;
+        uint256 participant1_net_transferred_amount;
         uint256 participant1_amount;
         uint256 participant2_amount;
         uint256 total_available_deposit;
@@ -579,18 +579,18 @@ contract TokenNetwork is Utils {
             participant2_state
         );
 
-        // Calculate the netted transferred amounts
+        // Calculate the net transferred amounts
         // We are sure this does not underflow
-        participant1_netted_transferred_amount = (
+        participant1_net_transferred_amount = (
             participant2_transferred_amount -
             participant1_transferred_amount
         );
 
         // case2. participant1_transferred_amount can be [0, real_participant1_transferred_amount)
-        // participant1_netted_transferred_amount is bigger than expected and can cause an overflow.
+        // participant1_net_transferred_amount is bigger than expected and can cause an overflow.
         // We just take the biggest amount that we can in this case.
         participant1_amount = failsafe_addition(
-            participant1_netted_transferred_amount,
+            participant1_net_transferred_amount,
             participant1_state.deposit
         );
 
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -22,6 +22,8 @@ contract TokenNetwork is Utils {
     // Chain ID as specified by EIP155 used in balance proof signatures to avoid replay attacks
     uint256 public chain_id;
 
+    uint256 constant public MAX_SAFE_UINT256 = 115792089237316195423570985008687907853269984665640564039457584007913129639935;
+
     // channel_identifier => Channel, where the channel identifier is the keccak256 of the
     // addresses of the two participants
     mapping (bytes32 => Channel) public channels;
@@ -454,6 +456,8 @@ contract TokenNetwork is Utils {
     )
         public
     {
+        // Convention required in getSettleTransferAmounts
+        require(participant2_transferred_amount >= participant1_transferred_amount);
         bytes32 channel_identifier;
 
         channel_identifier = getChannelIdentifier(participant1, participant2);
@@ -544,6 +548,20 @@ contract TokenNetwork is Utils {
         private
         returns (uint256, uint256, uint256, uint256)
     {
+        // Cases require attention:
+        // case1. If participant1 does NOT provide a balance proof or provides an old balance proof.
+        // participant2_transferred_amount can be [0, real_participant2_transferred_amount)
+        // We need to punish participant1.
+        // case2. If participant2 does NOT provide a balance proof or provides an old balance proof.
+        // participant1_transferred_amount can be [0, real_participant1_transferred_amount)
+        // We need to punish participant2.
+        // case3. If neither participants provide a balance proof, we just subtract their
+        // withdrawn amounts from their deposits.
+
+        // case1 is solved by making sure we only have to handle case2 or case3:
+        require(participant2_transferred_amount >= participant1_transferred_amount);
+
+        uint256 participant1_netted_transferred_amount;
         uint256 participant1_amount;
         uint256 participant2_amount;
         uint256 total_available_deposit;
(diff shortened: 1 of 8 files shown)


raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -558,58 +576,44 @@ contract TokenNetwork is Utils {
...
-        ) {
-            participant1_amount = 0;
-        }
+        // case2. participant1_transferred_amount can be [0, real_participant1_transferred_amount)
+        // participant1_netted_transferred_amount is bigger than expected and can cause an overflow.
+        // We just take the biggest amount that we can in this case.
+        participant1_amount = failsafe_addition(
+            participant1_netted_transferred_amount,
+            participant1_state.deposit
+        );
```
loredanacirstea: We can potentially try to re-add the overflowed amount again after subtracting the withdrawn amount. Not entirely sure it is needed. Opinions?

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -544,6 +548,20 @@ contract TokenNetwork is Utils {
...
+        // participant2_transferred_amount can be [0, real_participant2_transferred_amount)
+        // We need to punish participant1.
+        // case2. If participant2 does NOT provide a balance proof or provides an old balance proof.
+        // participant1_transferred_amount can be [0, real_participant1_transferred_amount)
+        // We need to punish participant2.
+        // case3. If neither participants provide a balance proof, we just subtract their
+        // withdrawn amounts from their deposits.
+
+        // case1 is solved by making sure we only have to handle case2 or case3:
+        require(participant2_transferred_amount >= participant1_transferred_amount);
```
err508: This is already done in line 460

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -454,6 +456,8 @@ contract TokenNetwork is Utils {
     )
         public
     {
+        // Convention required in getSettleTransferAmounts
+        require(participant2_transferred_amount >= participant1_transferred_amount);
```
err508: I think you could interchange `participant1` and `participant2` here appropriately if needed, so we have the assumption, but transactions do not fail.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -1160,4 +1164,34 @@ contract TokenNetwork is Utils {
...
+    /// @param a Addend
+    /// @param b Addend
+    /// @return Maximum between the result of the addition or the maximum uint256 value.
+    function failsafe_addition(uint256 a, uint256 b)
+        pure
+        internal
+        returns (uint256)
+    {
+        uint256 sum = a + b;
+        if (sum >= a && sum >= b) {
```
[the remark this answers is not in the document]
loredanacirstea: Yes, you are right.

raiden_contracts/contracts/TokenNetwork.sol
loredanacirstea: I really want this check to stay with the `getSettleTransferAmounts` function, to protect from people changing the code without knowing the importance of it.

I would rather remove the one from 460. 
But I also wanted:
- the main `settleChannel` to be aware that the order is important (I can just add a comment in the docstring to solve this one)
- fail faster and return more gas to the sender. (not really an issue)

So, I am fine with adding the info the docstring + removing line 460.

err508: You could add a comment in order _to protect from people changing the code without knowing the importance of it_ and get the other two for free.

raiden_contracts/contracts/TokenNetwork.sol
loredanacirstea: If we have valid balance proofs, `participant1_netted_transferred_amount + participant1_state.deposit` should never overflow, because we already ensure that `participant1_state.deposit + participant2_state.deposit` does not overflow.

If we do not have valid balance proofs, then it is fair to cap  `participant1_netted_transferred_amount + participant1_state.deposit` at `MAX_UINT256`.

So, I think the current approach is ok. We don't need to do what I proposed in my comment above.

err508: I think `participant1_state.deposit + participant2_state.deposit < MAX_UINT256 -> participant1_netted_transferred_amount + participant1_state.deposit < MAX_UINT256` does not hold, or do we have a require somewhere that I missed? 
I would, however, also do not add logic to deal with the case we're it might, as it is imo pretty unlikely to occur in a non-adverse scenario.

loredanacirstea: > does not hold

why? 
I was talking about the case where balance proofs are computed valid. Therefore, if a participant needs to receive a **positive** netted amount, this amount can only come from the partner's `deposit` => `participant1_netted_transferred_amount` <= `participant2_state.deposit`

err508: participant1_netted_transferred_amount = 2**255 <= 2**255 = participant2_state.deposit would be a valid variable assignment for a balance proof, that would overflow, no?

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -22,6 +22,8 @@ contract TokenNetwork is Utils {
     // Chain ID as specified by EIP155 used in balance proof signatures to avoid replay attacks
     uint256 public chain_id;
 
+    uint256 constant public MAX_SAFE_UINT256 = 115792089237316195423570985008687907853269984665640564039457584007913129639935;
```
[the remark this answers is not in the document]
loredanacirstea: I was reluctant to use `2 ** 256 - 1`, even though it works correctly now, because `2 ** 256` is not a valid value for `uint256`. 
I was reluctant to use `0 - 1` in case anything changes in the EVM regarding how underflows are handled.

I thought that having the actual number there is on the safer side. Maybe an overkill to think about future EVM changes, but the current implementation should be fine.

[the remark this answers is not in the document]
hackaugusto: I asked because `115792089237316195423570985008687907853269984665640564039457584007913129639935` is a number that I don't know by heart, so I had to run `2 ** 256 - 1` to check the value was correct.

but I can see your point, either way is fine to me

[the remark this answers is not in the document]
loredanacirstea: I also added a test comparing this number with `2 ** 256 - 1` (python) and with `0 - 1` (Solidity)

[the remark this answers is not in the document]
LefterisJP: > I was reluctant to use 0 - 1 in case anything changes in the EVM regarding how underflows are handled.

Stuff like that won't change anytime soon. It would break way too many production systems.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -544,6 +550,21 @@ contract TokenNetwork is Utils {
         private
         returns (uint256, uint256, uint256, uint256)
     {
+        // Cases require attention:
```
LefterisJP: Cases **that** require

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -544,6 +550,21 @@ contract TokenNetwork is Utils {
...
+        // participant1_transferred_amount can be [0, real_participant1_transferred_amount)
+        // We need to punish participant2.
+        // case3. If neither participants provide a balance proof, we just subtract their
+        // withdrawn amounts from their deposits.
+
+        // case1 is solved by making sure we only have to handle case2 or case3:
+        // we have the following check in settleChannel:
+        // require(participant2_transferred_amount >= participant1_transferred_amount)
+
+        uint256 participant1_netted_transferred_amount;
```
LefterisJP: netted -> net   both here and in all below occurences

raiden_contracts/contracts/TokenNetwork.sol
LefterisJP: > So, I think the current approach is ok. We don't need to do what I proposed in my comment above.

Yes it is definitely fair to cap participant1 at UINT256_safe_max at the case where participant2 has cheated.

As a sidenote, the comments here are really nice and explain the edge cases nicely but we should also document those in the spec eventually

raiden_contracts/tests/fixtures/channel.py
```diff
@@ -213,16 +222,11 @@ def get(
...
-            participant1,
-            transferred_amount1,
-            locked_amount1,
-            locksroot1,
-            participant2,
-            transferred_amount2,
-            locked_amount2,
-            locksroot2
-        ).transact({'from': participant1})
+        participant1_values = ChannelBalanceHashData(transferred=transferred_amount1, locked=locked_amount1, locksroot=locksroot1)
```
LefterisJP: I think those lines are too big, no?

raiden_contracts/tests/test_monitoring_service.py
```diff
@@ -107,14 +107,14 @@ def test_msc_happy_path(
...
-        0,                   # participant1_locked_amount
-        b'\x00' * 32,        # participant1_locksroot
         B,                   # participant2
         10,                  # participant2_transferred_amount
         0,                   # participant2_locked_amount
         b'\x00' * 32,        # participant2_locksroot
+        A,                   # participant1
+        20,                  # participant1_transferred_amount
+        0,                   # participant1_locked_amount
+        b'\x00' * 32,        # participant1_locksroot
```
LefterisJP: Just use `EMPTY_MERKLE_ROOT` from utils/merkle.py for this and the above. Optionally move it into a more central utils location.

raiden_contracts/tests/test_print_gas.py
```diff
@@ -129,8 +129,8 @@ def test_channel_cycle(
 
     web3.testing.mine(settle_timeout)
     txn_hash = token_network.functions.settleChannel(
+        B, 5, locked_amount2, locksroot2,
```
LefterisJP: style: each arg in its own line

loredanacirstea:
Opened https://github.com/raiden-network/raiden-contracts/issues/127 for the locked amount attack (old valid balance proof, high locked amount)

Issue #122, opened 2018-06-11
Add overflow check in setTotalDeposit for deposit1 + deposit2


loredanacirstea:
When calling `setTotalDeposit`, make sure the two participant's deposits don't overflow when added.

TODO:
- add `require`
- add test

PR #124, opened 2018-06-12
Add channel deposit overflow check & test

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/122

When calling `setTotalDeposit`, make sure the two participant's deposits don't overflow when added. This is very important for the settlement calculations.


loredanacirstea:
Fixed the tests. Thank you.

PR #128, opened 2018-06-13
Adding get_onchain_settlement_amounts function for testing

err508:
The function can be used to compare expected settlement
amounts and those that a realized onchain mod 2**256.

### DIFF
--- a/raiden_contracts/tests/fixtures/channel_test_values.py
+++ b/raiden_contracts/tests/fixtures/channel_test_values.py
@@ -24,13 +24,13 @@ class ChannelValues():
 channel_settle_test_values = [
     # both balance proofs provided are valid
     (
-        ChannelValues(deposit=40, withdrawn=10, transferred=20030, locked=6),
-        ChannelValues(deposit=35, withdrawn=5, transferred=20020, locked=4)
+        ChannelValues(deposit=35, withdrawn=5, transferred=20020, locked=4),
+        ChannelValues(deposit=40, withdrawn=10, transferred=20030, locked=6)
     ),
-    # participant1 does not provide a balance proof, locked amount ok
+    # participant2 does not provide a balance proof, locked amount ok
     (
-        ChannelValues(deposit=40, withdrawn=10, transferred=20030, locked=0),
-        ChannelValues(deposit=35, withdrawn=5, transferred=0, locked=0)
+        ChannelValues(deposit=35, withdrawn=5, transferred=0, locked=0),
+        ChannelValues(deposit=40, withdrawn=10, transferred=20030, locked=0)
     ),
     # participant2 does not provide a balance proof + bigger locked amount
     (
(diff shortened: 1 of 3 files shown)


raiden_contracts/tests/utils.py
```diff
@@ -156,6 +156,87 @@ def get_settlement_amounts(
...
+):
+    """ Settlement algorithm
+
+    Calculates the token amounts to be transferred to the channel participants when
+    a channel is settled.
+
+    !!! Don't change this unless you really know what you are doing.
+    """
+
+    def failsafe_add(a, b):
```
loredanacirstea: Would be nice to have these helper functions outside the `get_onchain_settlement_amounts` function, so we can use them to test the internal `failsafe_add` function in the smart contract separately.

raiden_contracts/tests/utils.py
```diff
@@ -156,6 +156,87 @@ def get_settlement_amounts(
         participant2_locked=participant2_locked,
     )
 
+def get_onchain_settlement_amounts(
```
loredanacirstea: Please use this in `settle_state_tests`, so we can compare both `get_onchain_settlement_amounts` and `get_settlement_amounts` and make sure there is no difference / no failing tests.
https://github.com/raiden-network/raiden-contracts/blob/40197cc59c61065c8d33eb2cefa35123e27b91fe/raiden_contracts/tests/fixtures/channel.py#L354

raiden_contracts/tests/fixtures/channel_test_values.py
```diff
@@ -24,13 +24,13 @@ def __repr__(self):
 channel_settle_test_values = [
     # both balance proofs provided are valid
     (
-        ChannelValues(deposit=40, withdrawn=10, transferred=20030, locked=6),
-        ChannelValues(deposit=35, withdrawn=5, transferred=20020, locked=4)
+        ChannelValues(deposit=35, withdrawn=5, transferred=20020, locked=4),
+        ChannelValues(deposit=40, withdrawn=10, transferred=20030, locked=6)
```
loredanacirstea: Why did you change the order?

err508: We require `transferred_amount_2 > transferred_amount_1` in `settleChannel`, that was the easiest way I could think off.

loredanacirstea: That's my point. It does not change anything. We call settle with the correct order here: https://github.com/raiden-network/raiden-contracts/blob/40197cc59c61065c8d33eb2cefa35123e27b91fe/raiden_contracts/tests/fixtures/channel.py#L586
so we don't have to think about ordering the test values ourselves. 
We even test that an incorrect ordering makes the `settleChannel` transaction to fail.

err508: Ah, I had to change them to make the tests work before rebasing. We also have to get rid of this one, if we keep it this way:
https://github.com/err508/raiden-contracts/blob/ca24b8e2509dd40f16192083461d77636627e880/raiden_contracts/tests/utils.py#L189-L201

I wrote it that way, as I wanted to leave the testing interface explicit, i.e. that every contract requirement must be thought off in order to have a valid tuple. If the order gets switched, it might have some unintuitive results.

loredanacirstea: > Ah, I had to change them to make the tests work before rebasing.

ok, that's fine.

```
def get_onchain_settlement_amounts(
        participant1,
        participant2
):
    assert(participant2.transferred >= participant1.transferred)
```
This is fine. Ok, I understand now. I forgot that `get_settlement_amounts` does not need this check.

err508: I think we should either get rid of this: 
https://github.com/raiden-network/raiden-contracts/blob/40197cc59c61065c8d33eb2cefa35123e27b91fe/raiden_contracts/tests/fixtures/channel.py#L586-L592

or this: 

https://github.com/raiden-network/raiden-contracts/pull/128/files#diff-a8f4ba6a11d8d896b2d05b76786b76d7R201

Otherwise we have working transactions with failing tests for no apparent reason.

PR #131, opened 2018-06-16
Fix how locked amounts are computed at settlement for old balance proof with high locked amount

### LINKED ISSUE #127: Old balance proof with high locked amount attack
https://github.com/raiden-network/raiden-contracts/pull/123#pullrequestreview-128356501
As @hackaugusto observed, there is an edge case where providing an old balance proof with a high locked amount attack enables cheating.

Example:

Last valid balance proofs:
```
A_deposit = 5
A_transferred = 3
A_locked = 0

B_deposit = 5
B_transferred = 8
B_locked = 0

withdrawn = 0 for both A, B
```

In this case, the settlement amounts received by the nodes are:
```
A_amount = 10
B_amount = 0
```

If B sends an old (valid) balance proof of A, with a higher locked amount:
```
A_deposit = 5
A_transferred = 0
A_locked = 3
```
the settlement amounts received by the nodes are:
```
A_amount = 7 -> contract locks 3
B_amount = 0 + 3 locked from A
```

This happens because we subtract the locked amounts after limiting the final balances to the maximum available deposit: https://github.com/raiden-network/raiden-contracts/blob/40197cc59c61065c8d33eb2cefa35123e27b91fe/raiden_contracts/contracts/TokenNetwork.sol#L604-L619

loredanacirstea:
Fix how locked amounts are computed at settlement

Fixes #127

```
P1 = participant1
D1 = deposit P1
W1 = withdrawn P1
T1 = transferred P1 -> P2
L1 = pending transfers P1 -> P2
A1 = amount to be received by P1 at settlement 

P2 = participant2
D2 = deposit P2
W2 = withdrawn P2
T2 = transferred P2 -> P1
L2 = pending transfers P2 -> P1
A2 = amount to be received by P2 at settlement

DC = D1 + D2 - W1 - W2 = current channel deposit
```

To simplify calculations, we artificially require the `settleChannel` arguments to be ordered:
```
T2 + L2 >= T1 + L1
```
https://github.com/raiden-network/raiden-contracts/blob/1046322d015a4ec06f6ad0b9961db194399f3563/raiden_contracts/contracts/TokenNetwork.sol#L882-L884

This comes from the way we compute the maximum receivable amount for `P1`
```
A1max = (T2 + L2) - (T1 + L1) + D1 - W1
```
https://github.com/raiden-network/raiden-contracts/blob/1046322d015a4ec06f6ad0b9961db194399f3563/raiden_contracts/contracts/TokenNetwork.sol#L889-L906

Note that we are including the locked amounts. 
These can be: 
- pending transfers that have not been finalized off-chain; we don't know at this point if the `secret`s have been registered or not, do we don't know if the locked amount will go entirely to the intended participant
- old pending transfers that have actually finalized on-chain and included in a newer balance proof; this is the case where cheating can be done - the participant can make sure to also register the `secret`s on-chain and use that later to try to withdraw more tokens - this attack is fixed by this PR, because:
```
A1max = min(A1max, DC)
A2max = DC - A1max
```
and `A1max` also contains `L2`. `A2max` contains `L1`.
https://github.com/raiden-network/raiden-contracts/blob/1046322d015a4ec06f6ad0b9961db194399f3563/raiden_contracts/contracts/TokenNetwork.sol#L599

We then subtract the locked amounts and bound them:
```
L2 = min(L2, A1max)
L1 = min(L1, A2max)
A1 = A1max - L2
A2 = A2max - L1
```
https://github.com/raiden-network/raiden-contracts/blob/1046322d015a4ec06f6ad0b9961db194399f3563/raiden_contracts/contracts/TokenNetwork.sol#L613-L623

And make sure we store the new locked amount in the contract and not the initial one.

**Tests**

Added tests that reverse balance proof values for `T1` and `L1` / `T2` and `L1` , and also provide random values for the locked amounts e.g. `L1 = randint(L1, L1 + T1)`, to simulate old balance proofs. All these cases are equivalent and should settle following this constraint:

```
A1 + L2 = A2 + L1
```

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -82,6 +82,13 @@ contract TokenNetwork is Utils {
         mapping(address => Participant) participants;
     }
 
+    struct SettlementData {
+        uint256 deposit;
+        uint256 withdrawn;
+        uint256 transferred;
+        uint256 locked;
+    }
+
     /*
      *  Events
      */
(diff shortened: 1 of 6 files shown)


loredanacirstea:
There are possibly other issues with how we compute the locked amounts. Specifically, I have not tested this model to see how it behaves for high/low values for the "unlockable" amount in a `locked_amount` vs. the "not-unlockable" amount. I opened https://github.com/raiden-network/raiden-contracts/issues/132 for this issue.

The current PR resumes to fix cases similar to https://github.com/raiden-network/raiden-contracts/issues/127 and does that by making the assumption that the entire `locked_amount1` might go to `participant2`. This assumption might not be enough for a fair calculation, but it improves the current settlement calculations anyway and adds more tests.

PR #133, opened 2018-06-18
Use remappings with import

rakanalh:
Since ContractManager already construct path remappings when invoking `solc`, this PR changes smart contracts to have them use `raiden/...` import path instead of `./`

Resolves https://github.com/raiden-network/raiden/issues/162


rakanalh:
I checked, it does break the raiden builds. Will investigate tomorrow.

rakanalh:
Tested on Raiden here:
https://travis-ci.org/rakanalh/raiden/builds/395022782
Build passed, so no breakage was introduced by this change.

Issue #92, opened 2018-05-28
Go over raiden smart contracts tests and make sure we cover each unit test case


loredanacirstea:
https://github.com/raiden-network/raiden/tree/5d60a3b064ed9eb8d61f69e25c1b185af4c09beb/raiden/tests/smart_contracts and https://github.com/raiden-network/raiden/tree/991275c9e8e9e4a4ccfdccb3a66a9e20950dd9b5/raiden/tests/property will be removed from the `raiden` repository, because it only contains smart contracts unit tests. These test cases should be covered in this repo.

loredanacirstea:
This can be closed. It was split in multiple issues, which are now done.

PR #134, opened 2018-06-18
unit tests for TokenNetwork internals

### LINKED ISSUE #129: unit testing of contract auxiliary functions
Rebuild the [auxiliary](https://github.com/raiden-network/raiden/blob/991275c9e8e9e4a4ccfdccb3a66a9e20950dd9b5/raiden/tests/smart_contracts/netting_channel/test_auxiliary_netting_channel.py) function testing that was removed with [#1437](https://github.com/raiden-network/raiden/pull/1437)

err508:
Contribution to #92, fixes #129. The unit tests from /raiden have so far been moved to raiden-contracts according to this mapping: 

```
**raiden/test_auxiliary_netting_channel -> raiden-contracts/test_unit_internals**

test_min_uses_usigned -> test_unit_internals.test_min_uses_usigned
test_max_uses_unsigned -> test_unit_internals.test_min_uses_usigned
test_min -> test_unit_internals.test_min
test_max -> test_unit_internals.test_max

test_recoverAddressFromSignature -> test_unit_internals.test_recoverAddressFromBalanceProof,
                                 -> test_recoverAddressFromBalanceProofUpdateMessage, 
                                 -> test_recoverAddressFromCooperativeSettleSignature,                     
                                 -> test_recoverAddressFromWithdrawMessage

 test_merkle_proof_one_lock -> computeMerkleRoot does not exist anymore, removed
 test_merkle_proof -->                       --- " ---
 test_merkle_proof_missing_byte ->           --- " ---
 test_signature_split ->                     --- " ---

```

### DIFF
--- a/raiden_contracts/tests/fixtures/channel.py
+++ b/raiden_contracts/tests/fixtures/channel.py
@@ -477,8 +477,10 @@ def create_balance_proof(token_network, get_private_key):
             locksroot=None,
             additional_hash=None,
             v=27,
-            signer=None
+            signer=None,
+            other_token_network=None
     ):
+        _token_network = other_token_network or token_network
         private_key = get_private_key(signer or participant)
         locksroot = locksroot or b'\x00' * 32
         additional_hash = additional_hash or b'\x00' * 32
@@ -487,8 +489,8 @@ def create_balance_proof(token_network, get_private_key):
 
         signature = sign_balance_proof(
             private_key,
-            token_network.address,
-            int(token_network.functions.chain_id().call()),
+            _token_network.address,
+            int(_token_network.functions.chain_id().call()),
             channel_identifier,
             balance_hash,
             nonce,
@@ -513,14 +515,16 @@ def create_balance_proof_update_signature(token_network, get_private_key):
             nonce,
             additional_hash,
             closing_signature,
-            v=27
+            v=27,
+            other_token_network=None
     ):
+        _token_network = other_token_network or token_network
         private_key = get_private_key(participant)
 
         non_closing_signature = sign_balance_proof_update_message(
             private_key,
-            token_network.address,
-            int(token_network.functions.chain_id().call()),
+            _token_network.address,
+            int(_token_network.functions.chain_id().call()),
             channel_identifier,
             balance_hash,
             nonce,
@@ -541,15 +545,17 @@ def create_cooperative_settle_signatures(token_network, get_private_key):
             participant1_balance,
             participant2_address,
             participant2_balance,
-            v=27
+            v=27,
+            other_token_network=None
     ):
+        _token_network = other_token_network or token_network
         signatures = []
         for participant in participants_to_sign:
             private_key = get_private_key(participant)
             signature = sign_cooperative_settle_message(
                 private_key,
-                token_network.address,
-                int(token_network.functions.chain_id().call()),
+                _token_network.address,
+                int(_token_network.functions.chain_id().call()),
                 channel_identifier,
                 participant1_address,
                 participant1_balance,
--- a/raiden_contracts/tests/test_unit_internals.py
+++ b/raiden_contracts/tests/test_unit_internals.py
@@ -1,0 +1,338 @@
+# -*- coding: utf-8 -*-
+
+from itertools import chain, product
+
+import pytest
+from eth_tester.constants import UINT256_MIN, UINT256_MAX
+from eth_tester.exceptions import TransactionFailed
+
+from raiden_contracts.tests.fixtures import fake_bytes
+from web3.exceptions import ValidationError
+
+from raiden_contracts.tests.fixtures.config import fake_hex
+
+
+def test_min_uses_usigned(token_network_test):
+    """ Min cannot be called with negative values. """
+    INVALID_VALUES = [-UINT256_MAX, -1]
+    VALID_VALUES = [UINT256_MIN, UINT256_MAX, UINT256_MAX]
+
+    all_invalid = chain(
+        product(VALID_VALUES, INVALID_VALUES),
+        product(INVALID_VALUES, VALID_VALUES),
+    )
+
+    for a, b in all_invalid:
+        with pytest.raises(ValidationError):
+            token_network_test.functions.minPublic(a, b).call()
+
+
+def test_max_uses_usigned(token_network_test):
+
+    INVALID_VALUES = [-UINT256_MAX, -1]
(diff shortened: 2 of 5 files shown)


raiden_contracts/tests/test_unit_internals.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def test_max(token_network_test):
+
+    VALUES = [UINT256_MIN, 1, UINT256_MAX, UINT256_MAX]
+    for a, b in product(VALUES, VALUES):
+        assert token_network_test.functions.maxPublic(a, b).call() == max(a, b)
+
+
+def test_recoverAddressFromWithdrawMessage(
```
loredanacirstea: I would propose to keep signature verification in the same place. Either add tests for this in a separate file for each type of signature, or add them to the specific test file - e.g. I already have a test for `recoverAddressFromWithdrawMessage` in https://github.com/raiden-network/raiden-contracts/blob/5468fcdc249644838d21ef1e90849b6aa1273b20/raiden_contracts/tests/test_channel_withdraw.py#L17

We don't need to keep the original `raiden` order if we cannot easily track what additional tests need to be added.

raiden_contracts/tests/test_unit_internals.py
```diff
@@ -0,0 +1,246 @@
...
+            signature_A
+        ).call()
+    with pytest.raises(TransactionFailed):
+        token_network_test.functions.verifyWithdrawSignaturesPublic(
+            channel_identifier,
+            A,
+            B,
+            3,
+            signature_A,
+            fake_bytes(64)
```
loredanacirstea: Consider adding a constant for `FAKE_SIGNATURE`.

raiden_contracts/tests/test_unit_internals.py
```diff
@@ -0,0 +1,246 @@
...
+        token_network_test.functions.verifyWithdrawSignaturesPublic(
+            channel_identifier,
+            A,
+            B,
+            3,
+            fake_bytes(64),
+            signature_B
+        ).call()
+
+    recovered_address_A = token_network_test.functions.recoverAddressFromWithdrawMessagePublic(
```
loredanacirstea: This should be a separate test. You are testing another function, also partially tested in https://github.com/raiden-network/raiden-contracts/blob/5468fcdc249644838d21ef1e90849b6aa1273b20/raiden_contracts/tests/test_channel_withdraw.py#L17

raiden_contracts/tests/test_unit_internals.py
```diff
@@ -0,0 +1,246 @@
...
+
+    with pytest.raises(TransactionFailed):
+        token_network_test.functions.recoverAddressFromWithdrawMessagePublic(
+            channel_identifier,
+            A,
+            withdraw_A,
+            fake_hex(64)
+        ).call()
+
+    assert recovered_address_A != \
```
loredanacirstea: Code style: instead of `\` , please use an additional variable to store the contract call result.

raiden_contracts/tests/test_unit_internals.py
```diff
@@ -0,0 +1,246 @@
...
+            A,
+            withdraw_A,
+            signature_B
+        ).call()
+
+
+def test_recoverAddressFromBalanceProof(
+        token_network_test,
+        create_balance_proof_internal,
+        create_balance_proof_update_signature_internal,
```
loredanacirstea: Why use `create_balance_proof_internal` and `create_balance_proof_update_signature_internal`  and not the already existent `create_balance_proof`, `create_balance_proof_update_signature` ?

raiden_contracts/tests/fixtures/channel.py
```diff
@@ -532,6 +531,72 @@ def get(
     return get
 
 
+@pytest.fixture()
+def create_balance_proof_internal(token_network_test, get_private_key):
```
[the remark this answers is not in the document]
err508: The fixture has to use `token_network_test` s.t. the signatures are generated for `token_network_test .address`. I ended up with duplication to achieve this, please suggest are more elegant solution to make the fixture inheritance more flexible, if you are aware of one.

raiden_contracts/tests/test_unit_internals.py
err508: The order/grouping of the tests is currently changing. My current approach is to have tests in the `test_unit_internals` for every function in `token_network` in the same order as they appear there. Imo if ordered in this way, it will come out most clearly that every function is tested. This would imply refactored/duplicated testing of some tests that are already in place. Let me know your stance pls.

raiden_contracts/tests/test_unit_internals.py
err508: Done.

raiden_contracts/tests/test_unit_internals.py
err508: Agreed.

raiden_contracts/tests/fixtures/channel.py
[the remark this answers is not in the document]
loredanacirstea: You can add an additional argument `other_token_network=None` to the `create_*` fixture + 
`token_network = other_token_network or token_network` in the fixture.

raiden_contracts/tests/test_unit_internals.py
```diff
@@ -0,0 +1,246 @@
+# -*- coding: utf-8 -*-
+
+from itertools import chain, product
+
+import pytest
+from eth_tester.constants import UINT256_MIN, UINT256_MAX
```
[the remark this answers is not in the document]
loredanacirstea: I would prefer this to them changing something and us not knowing.
I would go with the approach that we should know & control what we are testing.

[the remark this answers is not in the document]
err508: >I would prefer this to them changing something and us not knowing.
>I would go with the approach that we should know & control what we are testing.

Not clear if understand that correctly. We only know if they change something, if use their constants. If we use their constants, we cannot control them.

raiden_contracts/tests/fixtures/channel.py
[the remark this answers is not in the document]
err508: Cool, I'll change that

raiden_contracts/tests/test_unit_internals.py
loredanacirstea: You could also break `test_unit_internals` into `test_unit_internals_signatures` & others.  
If you see actual duplicated tests in another file, you can just remove them from there.
Otherwise, I think it is fine.

PR #135, opened 2018-06-19
query token addresses for raiden start up

nkbai:
how to tokens  already registered?


hackaugusto:
hey @nkbai , with the new smart contracts you will need query for the smart contract events using [an installed filter](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_newfilter) or [getting the logs directly](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getfilterlogs), we removed the getters to reduce gas usage (storage is quite expensive).

So I'm closing this, since the array was removed as an optimization.

nkbai:
@hackaugusto I think that events of TokenNetwork.sol   should add TokenNetworkRegistry address,
so raiden can filter and found Channel events much easier

PR #138, opened 2018-06-19
Include secret in SecretRevealed event

### LINKED ISSUE #137: Add `secret` as an argument in the `SecretRevealed` event
Related issue: https://github.com/raiden-network/raiden/issues/1588

It should not be indexed, as whoever looks it up would look it by hash and would then get the Secret that corresponds to the same event.

palango:
Fixes #137

### DIFF
--- a/raiden_contracts/contracts/SecretRegistry.sol
+++ b/raiden_contracts/contracts/SecretRegistry.sol
@@ -15,7 +15,7 @@ contract SecretRegistry {
      *  Events
      */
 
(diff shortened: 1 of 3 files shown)


loredanacirstea:
Please also add `secret` here:
- https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/utils/events.py#L1
- https://github.com/raiden-network/raiden-contracts/blob/a1592fc88768a0c04caf3fbf31d165ac3dadbbcd/raiden_contracts/tests/test_secret_registry.py#L74

PR #154, opened 2018-06-25
Resolve missing unlock tests from raiden/[...]/test_withdraw.

### LINKED ISSUE #144: Resolve unlock tests from raiden/test_withdraw.py
Go over old [raiden/test_withdraw](https://github.com/raiden-network/raiden/blob/991275c9e8e9e4a4ccfdccb3a66a9e20950dd9b5/raiden/tests/smart_contracts/netting_channel/test_withdraw.py)  and rebuild the testing in `raiden-contracts `.  Be aware: `nettingchannel.withdraw` corresponds to `token_network.unlock` with the current protocol. I.e. the currently available tests in raiden-contracts are found [here](https://github.com/raiden-network/raiden-contracts/blob/d0be6f5ccc4be6498db85c182d24b4e3a5b8f681/raiden_contracts/tests/test_channel_unlock.py#L15):  

- [x] decide which tests to rewrite
- [x] adapt tests to new functionality with `withdraw <-> unlock`

err508:
Resolves #144 .


raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -6,12 +6,203 @@
...
+    # Create channel and deposit
+    create_channel(A, B, settle_timeout)[0]
+    channel_deposit(A, values_A.deposit, B)
+    channel_deposit(B, values_B.deposit, A)
+
+    # Mock pending transfers data for A
+    pending_transfers_tree_A = get_pending_transfers_tree(web3, [1, 3, 5], [2, 4], settle_timeout)
+    locksroot_bytes = get_merkle_root(pending_transfers_tree_A.merkle_tree)
+    values_B.locksroot = '0x' + locksroot_bytes.hex()
+    values_B.locked = get_locked_amount(pending_transfers_tree_A.transfers)
```
loredanacirstea: So this uncovered a bug.
Here we should have `values_A.locksroot` and `values_A.locked`. At lines 151, 152 we should have `values_B.locksroot` and `values_B.locked`.

The problem is that tests pass because in the `TokenNetwork` smart contract, we have this mapping:
```
mapping (bytes32 => uint256) locksroot_to_locked_amount;
```
There is no way currently to see what participant provided what `locksroot`. Therefore, A can `unlock` its own merkle tree of pending transfers instead of the one provided by B, if this is more advantageous. This behavior is incorrect and leads to cheating. We need to fix this. Opened https://github.com/raiden-network/raiden-contracts/issues/157 for this.

So, for now, just do the swap between `values_A` and `values_B` and we will fix the bug later & add a test for it (with the logic that you actually used here).

raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -6,12 +6,203 @@
...
+    token_network.functions.unlock(
+        B,
+        A,
+        pending_transfers_tree_B.packed_transfers,
+    ).transact()
+
+    balance_A = custom_token.functions.balanceOf(A).call()
+    balance_B = custom_token.functions.balanceOf(B).call()
+    balance_contract = custom_token.functions.balanceOf(token_network.address).call()
+    unlockable_A = get_locked_amount(pending_transfers_tree_A.unlockable)
```
loredanacirstea: To get the true unlocked amount, better use `get_unlocked_amount` , because it also checks the registered secrets. It is possible to construct a bad mocked pending transfers example where not all of the secrets have time to be registered (too small expiration).

https://github.com/raiden-network/raiden-contracts/blob/d0be6f5ccc4be6498db85c182d24b4e3a5b8f681/raiden_contracts/tests/utils.py#L301

And just get the amount that will be returned by subtracting the above value from the total locked amount.

raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -70,136 +261,272 @@ def test_merkle_root(
...
+    # Calling unlock twice does not work
+    with pytest.raises(TransactionFailed):
+        token_network.functions.unlock(
+            A,
+            B,
+            pending_transfers_tree_1.packed_transfers,
+        ).transact({'from': A})
+
+
+def test_unlock_wrong_locksroot(
```
loredanacirstea: Again, this moving around does not make reviewing easy..

raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -70,136 +261,272 @@ def test_merkle_root(
...
+
+    with pytest.raises(TransactionFailed):
+        token_network.functions.unlock(
+            A,
+            B,
+            pending_transfers_tree_A_fake.packed_transfers,
+        ).call()
+
+
+def test_channel_unlock_bigger_locked_amount(
```
loredanacirstea: Please reorder the tests as they were, so I can actually see the new changes. I will continue the review afterwards.

raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -326,7 +652,76 @@ def test_channel_unlock(
...
+
+    balance_A = custom_token.functions.balanceOf(A).call()
+    balance_B = custom_token.functions.balanceOf(B).call()
+    balance_contract = custom_token.functions.balanceOf(token_network.address).call()
+    assert balance_A == pre_balance_A + 9
+    assert balance_B == pre_balance_B + 6
+    assert balance_contract == pre_balance_contract - values_B.locked
+
+
+def test_unlock_channel_event(
```
loredanacirstea: Consider placing the event test last in the file. I used this convention for each feature test file.

raiden_contracts/tests/test_channel_unlock.py
loredanacirstea: Yes, was thinking of adding the `participant_address` somewhere. E.g. `mapping (bytes32 => uint256) locksroot_to_locked_amount;` to be `keccak256(locksroot + participant_address) => locked_amount`.

But `keccak256(channel_identifier + participant_address) => UnlockData ` might also work.
I'm still thinking about what would be the best approach.

loredanacirstea: The key point here is to know who the locked tokens belong to. If you link a `locksroot` to a `participant_address` in the smart contract, you know that. This will be done at `closeChannel` & `updateNonClosingBalance`, where you already have signatures.

E.g. `locksroot_A` is the merkle root of `A`, meaning that A sent those pending transfers to B -> you know who the `unlocked_amount` and `not_unlocked_amount` belongs to. 

So I don't see the reason for adding yet another pair of signatures for `unlock` if the smart contract already knows who has what `locksroot`. Am I missing something? 

In general, I think it would be optimal if:
1. we make sure tokens go where they belong (as explained above)
2. allow `unlock` to be called by anyone willing to pay the gas cost, as long as 1. is respected. This would allow people to use some additional systems/services where you pay for gas with tokens and the service makes the transaction for you.

err508: Using the signatures from `closeChannel` and `updateNonClosingBalance`  to determine the locksroot ownership does the job, but mapping `locksroot -> participant_address` is ambiguous. I'd also think about using a signature to determine that at least for now, as changing `unlock data` or the mapping might require a lot more refactoring.

loredanacirstea: I would map `keccak256(locksroot + participant_address) => locked_amount` - what is ambiguous about this?

I still don't understand what would a signature achieve. You would still need to change the unlock data even with the signature. IOW the signature would just be an additional change on top of the data structure changes I mentioned - and I don't see the reason.

But, feel free to spec out some code here, so I can see exactly what you have in mind. It is possible I don't understand your idea entirely.

raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -449,3 +434,564 @@ def test_channel_settle_and_unlock(
...
+
+    close_and_update_channel(
+        A,
+        values_A,
+        B,
+        values_B,
+    )
+
+    # Settle channel after a "long" time
+    web3.testing.mine(settle_timeout + 1000)
```
loredanacirstea: Original `raiden` test was:
```
""" Withdraw must accept a lock that expires after the settlement period. """
[...]
# use a really large expiration
lock_expiration = tester_chain.block.number + settle_timeout * 5

[...]
tester_chain.mine(number_of_blocks=settle_timeout + 1)
```

This test indeed uses lock `expiration > settle_timeout`.

But `web3.testing.mine(settle_timeout + 1000)` is redundant and just takes up more testing time - because you wait `1000` more blocks.
I see no reason to wait this much - pertaining to the actual test.

loredanacirstea:
So you are duplicating the already existing https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/tests/test_recover_from_signature.py and also adding it to a `raiden_contracts/tests/unit` folder.

Please don't duplicate files. You can rename/move them.

Also, this should not be part of this PR.

raiden_contracts/tests/test_channel_unlock.py
err508: We have `expiration >= settle_timeout` tested in every successfull test. I agree it's annoying to wait for it, but I'd a) remove it entirely or b) let the "large" expiration time actually pass, otherwise this is doing nothing.

raiden_contracts/tests/test_channel_unlock.py
err508: It is last in the file

raiden_contracts/tests/test_channel_unlock.py
err508: As is it is now in the PR with `keccak256(channel_identifier, participant, locksroot)` it is clear.
Signature alone would't help. Yet might be good to have standalone authentication in `unlock`

raiden_contracts/tests/test_channel_unlock.py
loredanacirstea: > otherwise this is doing nothing.

The test is not for testing settlement after lock expiration. The Raiden client is not aware of the expiration and will settle the first chance it gets.
This test makes sure there are no conditions that would make an unlock fail if it contains secrets with big expirations - bigger than settlement window.

loredanacirstea: Measured mining time.
`web3.testing.mine(1000)` - `27.40565586090088` seconds
`web3.testing.mine(100)` - `0.2555418014526367` seconds
`web3.testing.mine(1)` - `0.003122091293334961` seconds

So `web3.testing.mine(100)` doesn't add that much now. It can remain like this.

PR #159, opened 2018-06-28
Add bumpversion for contracts

pcppcp:
- add bumpversion config for contracts and `setup.py`
- bump eth-tester

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,7 @@
+[bumpversion]
+current_version = 0.0.1
+commit = True
+tag = False
+
+[bumpversion:file:setup.py]
+serialize = {major}.{minor}.{patch}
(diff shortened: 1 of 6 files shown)


.bumpversion.cfg
```diff
@@ -0,0 +1,7 @@
+[bumpversion]
+current_version = 0.0.1
```
loredanacirstea: Why is this version different than the contracts version? Wouldn't it make sense to keep the package & contracts version in sync?

pcppcp: a simplest answer would be that you might want to update python code (i.e. contract manager) without changing on-chain contract's version

loredanacirstea: Ah, right. We can keep the `major` (maybe also `minor`) version synced. Thanks.

PR #162, opened 2018-06-29
Update settlement timeouts

palango:
Update the settlement timeouts to the decided values.

Part of raiden-network/spec#16


loredanacirstea:
The problem here is that we need to add at least the minimum settlement timeout as an argument for the `TokenNetworkRegistry` and `TokenNetwork` - otherwise settlement tests will take way too long.

palango:
> The problem here is that we need to add at least the minimum settlement timeout as an argument for the TokenNetworkRegistry and TokenNetwork - otherwise settlement tests will take way too long.

Because we'll have to mine 500 blocks for each test?

loredanacirstea:
> Because we'll have to mine 500 blocks for each test?

Yes. For the settlement window.
So for now, we have decided to set this variable at deployment time.

raiden_contracts/constants.py
```diff
@@ -29,8 +29,11 @@
 EVENT_ADDRESS_REGISTERED = 'AddressRegistered'
 
 # Timeouts
-SETTLE_TIMEOUT_MIN = 6
-SETTLE_TIMEOUT_MAX = 2700000
+TEST_SETTLE_TIMEOUT_MIN = 5
+TEST_SETTLE_TIMEOUT_MAX = 100000
```
palango: Just pulled the values out of my ass, we can revert to the old ones if they have meaning.

PR #163, opened 2018-07-02
Limit the total deposit to 100 tokens

### LINKED ISSUE #103: Raiden Smart Contract Value Limits for Mainnet Release
For the Red Eyes release we want to put artificial limits on the smart contracts for security reasons.

- [x] Figure out the amount of limits that should be put and how they should be. Kind of difficult due to different value of each token. For now let's go with 100 tokens.

- [x] Implement the limits in the contracts

- [x] Add some tests for the contract limits

palango:
Fixes #103


raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -27,6 +27,8 @@ contract TokenNetwork is Utils {
 
     uint256 constant public MAX_SAFE_UINT256 = 115792089237316195423570985008687907853269984665640564039457584007913129639935;
 
+    uint256 constant public DEPOSIT_LIMIT = 100 * (10 ** 18); // Assume 18 decimals for the token
+
```
err508: Why not use the `multiplier`here?

palango: what multiplier?

err508: In the `HumanStandardToken`, there is `_decimalUnits` in the constructor, in `customtoken`, we derive `        multiplier = 10**(uint256(decimal_units))` from that. I we might wanna use that - otherwise the our assumptions on the `DEPOSIT_LIMIT` might be really wrong.

palango: Yeah, but our `HumanStandardToken` is just an implementation of the `Token` interface. And the standard [1] doesn't require `_decimal_units`, so we can't use it.


[1]: https://github.com/ethereum/eips/issues/20

err508: How do we enforce this assumption? IMO if a token has `decimal_units` set, we should use it, else set  `DEPOSIT_LIMIT = 100` to be safe

palango: I think we can assume that most tokens have 18 digits

> IMO if a token has decimal_units set, we should use it, else set DEPOSIT_LIMIT = 100 to be safe.

How would that work in solidity code?

>  If a contract uses decimals without using the standard, that should not be our problem.

The problem is the there is no standard for this.

LefterisJP: @palango 

> The problem is the there is no standard for this.

There [is](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-20.md#decimals) but it's optional. It's the `decimasl()` function I linked above. It does not need to be present, but it's present in almost all of the tokens out there.

> How would that work in solidity code?

If they don't use decimals, I would default to 18 decimals, if they do I would use it.

It would work in the following way. Warning it's ugly:

```solidity

bool exists = addr(token).call(bytes4(sha3("decimals()")));
uint8 decimals = 18;
if (exists) {
    decimals = token.decimals();
}
```

For the mainnet we can also simply use the tokens list of mycrypto. It's updated all the time and contains decimals. I use it in rotkehlchen already. It's [here](https://github.com/MyCryptoHQ/MyCrypto/blob/2b60ba96500fe4445e605996e989936be29587df/common/config/tokens/eth.json).

palango: > It would work in the following way. Warning it's ugly:

I get the following warning:
```
E   /Users/paul/Work/raiden-contracts/raiden_contracts/contracts/TokenNetwork.sol:187:24: Error: Member "decimals" not found or not visible after argument-dependent lookup in contract Token
E       decimals = token.decimals();
E                  ^------------^
```

I have some minor changes to the code:
```sol
bool exists = address(token).call(bytes4(keccak256("decimals()")));
        uint8 decimals = 18;
        if (exists) {
            decimals = token.decimals();
        }
```

LefterisJP: Add decimals() to the token interface. If it's not there the `exists` would be false anyway

raiden_contracts/contracts/test/CustomToken.sol
```diff
@@ -84,4 +84,8 @@ contract CustomToken is StandardToken {
         owner_address.transfer(address(this).balance);
         assert(address(this).balance == 0);
     }
+
+    function decimals() public view returns (uint8 decimals) {
+        return _decimals;
+    }
```
loredanacirstea: This should not be necessary. The getter is added by the compiler if `uint8 decimals` is `public`. But you can leave it.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -27,6 +27,9 @@ contract TokenNetwork is Utils {
 
     uint256 constant public MAX_SAFE_UINT256 = 115792089237316195423570985008687907853269984665640564039457584007913129639935;
 
+    uint256 public deposit_limit;
+    uint8 public decimals;
```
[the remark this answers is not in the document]
LefterisJP: +1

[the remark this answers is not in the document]
palango: Ok, will remove it

Issue #149, opened 2018-06-25
Resolve channel_manager tests from raiden/[...]/test_channel_manager


err508:
Go over old [raiden/[...]/test_channel_manager](https://github.com/raiden-network/raiden/blob/991275c9e8e9e4a4ccfdccb3a66a9e20950dd9b5/raiden/tests/smart_contracts/test_channel_manager.py) and rebuild the tests that that are missing. 
Following tests need approval to remove or have to be rebuild. Pls edit issue in order to make transpaerent what happend to the test, i.e. add a link to where the functionality is tested already, a link to the new test or write a comment why it is not neccessary to be tested anymore:
- [ ]  netting_channel_settled
- [ ]  test_channelnew_event
- [ ]  test_channeldeleted_event
- [ ]  test_newchannel_fails_until_channel_is_settled
- [ ]  test_getchannelwith_must_return_zero_for_non_existing_channels
- [ ]  test_channelmanager_start_with_zero_entries
- [ ]  test_channelmanager
- [ ]  test_for_issue_892
- [ ]  test_reopen_channel
- [ ]  test_reopen_regression_bad_index_update
- [ ]  test_new_channel_state

err508:
sub #92

loredanacirstea:
`test_channelnew_event` - covered by https://github.com/raiden-network/raiden-contracts/blob/d6e7f6d08ff3a0c723b83600cae851bb479d137d/raiden_contracts/tests/test_channel_open.py#L99

`test_channeldeleted_event` - equivalent test now: https://github.com/raiden-network/raiden-contracts/blob/d6e7f6d08ff3a0c723b83600cae851bb479d137d/raiden_contracts/tests/test_channel_settle.py#L196

`test_newchannel_fails_until_channel_is_settled` - must be added

`test_getchannelwith_must_return_zero_for_non_existing_channels` - covered by https://github.com/raiden-network/raiden-contracts/blob/d6e7f6d08ff3a0c723b83600cae851bb479d137d/raiden_contracts/tests/test_channel_open.py#L62-L64 (the new way to check if a channel exists or not)

`test_channelmanager_start_with_zero_entries` - cannot replicate test with the current `TokenNetwork` contract 

`test_channelmanager` - outdated; we can say that for the new contracts, this should be enough: https://github.com/raiden-network/raiden-contracts/blob/d6e7f6d08ff3a0c723b83600cae851bb479d137d/raiden_contracts/tests/test_channel_open.py#L58

[...]

PR #165, opened 2018-07-02
Adds test_reopen_channel.

err508:
As described in #149.

### DIFF
--- a/raiden_contracts/tests/test_channel_open.py
+++ b/raiden_contracts/tests/test_channel_open.py
@@ -96,6 +96,83 @@ def test_open_channel_state(token_network, get_accounts):
     assert B_nonce == 0
 
 
+def test_reopen_channel(
+        web3,
+        token_network,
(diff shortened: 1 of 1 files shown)


raiden_contracts/tests/test_channel_open.py
```diff
@@ -96,6 +96,53 @@ def test_open_channel_state(token_network, get_accounts):
...
+        locksroot,
+        B,
+        0,
+        0,
+        locksroot,
+    ).transact({'from': A})
+
+    # Reopening the channel should work iff channel is settled
+    token_network.functions.openChannel(A, B, settle_timeout).transact()
+
```
loredanacirstea: Please also `check channel state to make sure we don't have lingering balance proofs from the previous one)`, as I mentioned here https://github.com/raiden-network/raiden-contracts/issues/149#issuecomment-400365496

PR #166, opened 2018-07-03
Register secret batch

jomuel:


raiden_contracts/utils/events.py
```diff
@@ -5,6 +5,21 @@ def get(event):
     return get
 
 
+def check_secrets_revealed(secrethashes, secrets):
```
LefterisJP: Isn't this a little bit complicated? Do you really need this closure style of construction?

Also, since this is only used in tests, it should be in the one of the utils under the `tests/` directory.

jomuel: Ok, I changed it. Checking that the correct number of events are emitted and each has reasonable parameters will be enough, we would only need the closure to verify each of the secrets was revealed.

Afaict everything in utils/ is used only in tests, except for utils/utils.py.

Issue #141, opened 2018-06-20
how to avoid replay attack to a channel?


nkbai:
for a reopened channel on a tokennetwork.sol,  how to detect that a participant is submitting the balance proof of last version?
1. v1 channel: A-B open a channel, channel id is 0x000001
 2. some transfers between A,B,
3. close and settle 0x000001
4. reopen v2 channel, this channel id is also 0x000001
5. A use V1's balance proof to close this v2 channel.

nkbai:
what about to use tokenetwork address, chain id, channel identifier and opened block number as the unique channel id?
user's signature of balance proof contains all these four arguments.

nkbai:
cooperate settle and withdraw has the same problem too.

loredanacirstea:
Correct, that is why I said `include it in the signed messages`. This block number should be included in all of them.

PR #170, opened 2018-07-05
Add channel counter - fix replay attack on reopened channels

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/141

The issue described a replay attack possible after reopening a settled channel. Old balance proofs could have been used to cheat, because there was way to differentiate between the old and new signed messages.

Two options were discussed:
1. Adding the block number at which the channel was opened in the signed messages
2. Making `channel_identifier` unique with either adding the `open_block_number` or an increasing nonce (counter)

**Option 1:**
- saving `open_block_number` in the channel state
- adding `open_block_number` to all signed messages
- adding `open_block_number` to all channel events
-> this also means changes to the Raiden client

**Option 2** (suggested by @hackaugusto):
- add a global channel counter that increases after each channel creation
- `channel_identifier` must include the participant addresses and the channel counter
-> minimal changes in the Raiden client

Tried both options, Option 2 seems the best one (minimal changes to `TokenNetwork` ABI). Raiden client changes are minimal. Gas costs are similar.

**Implementation:**
- added `uint256 public channel_counter;` - global counter for all channels, starts with `1`
- `channel_identifier` is `keccak256(keccak256(lexicographic order of participant addresses), channel_counter)`
- added a mapping `mapping (bytes32 => uint256) public participants_hash_to_channel_counter;` to check if a channel exists. 
- remove `participants_hash_to_channel_counter` entry for a channel after settlement
- `ChannelUnlocked` event - remove `channel_identifier`, add `locksroot` to arguments, because `channel_identifier` is invalid after settlement

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -29,17 +29,22 @@ contract TokenNetwork is Utils {
 
     uint256 public deposit_limit;
 
-    // channel_identifier => Channel, where the channel identifier is the keccak256 of the
-    // addresses of the two participants
+    uint256 public channel_counter;
+
+    // channel_identifier => Channel
+    // channel identifier is the keccak256(keccak256(lexicographic order of participant addresses), channel_counter)
     mapping (bytes32 => Channel) public channels;
+
+    // This is needed to enforce one channel per pair of participants
+    // The key is keccak256(participant1_address, participant2_address)
+    mapping (bytes32 => uint256) public participants_hash_to_channel_counter;
 
     // We keep the unlock data in a separate mapping to allow channel data structures to be
     // removed when settling uncooperatively. If there are locked pending transfers, we need to
     // store data needed to unlock them at a later time.
-    // The key is `keccak256(channel_identifier + participant_address + locksroot)`
+    // The key is `keccak256(participants_hash + participant_address + locksroot)`
     // The value is the total amount of tokens locked in the pending transfers corresponding to
-    // the `locksroot`, that were sent by `participant_address` to his partner from the channel
-    // represented by the `channel_identifier`.
+    // the `locksroot`, that were sent by `participant_address` to his channel partner.
     // Note that the assumption is that no two locksroots can be the same due to having different
     // values for the secrethash of each lock, combined with different `expiration` values.
     mapping(bytes32 => uint256) locksroot_identifier_to_locked_amount;
@@ -116,8 +121,9 @@ contract TokenNetwork is Utils {
     event ChannelClosed(bytes32 indexed channel_identifier, address indexed closing_participant);
 
     event ChannelUnlocked(
-        bytes32 indexed channel_identifier,
         address indexed participant,
+        address indexed partner,
+        bytes32 indexed locksroot,
         uint256 unlocked_amount,
         uint256 returned_tokens
     );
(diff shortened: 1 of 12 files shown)


raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -194,6 +200,18 @@ contract TokenNetwork is Utils {
         public
         returns (bytes32)
     {
+        channel_counter += 1;
```
palango: The global counter is used instead of a per channel counter to save gas by deleting entries from `participants_hash_to_channel_counter`, right?

loredanacirstea: Correct, but we would also have redundant data in storage, which is not a good practice (all per channel counters for settled channels, that will not be reopened).

raiden_contracts/tests/test_channel_open.py
```diff
@@ -55,16 +56,81 @@ def test_max_1_channel(token_network, get_accounts, create_channel):
...
 
 
+def test_participants_hash(token_network, get_accounts):
+    (A, B) = get_accounts(2)
+
+    AB_hash = get_participants_hash(A, B)
+    assert token_network.functions.getParticipantsHash(A, B).call() == AB_hash
+
+
+def test_counter(token_network, get_accounts, create_channel):
```
hackaugusto: do we have a test for the replay? I think this is the more important thing that this PR fixes, the nonce is just a implementation detail

loredanacirstea: Added replay tests for all the signed messages that we have - for reopened channels.

raiden_contracts/tests/test_channel_open.py
```diff
@@ -139,8 +209,12 @@ def test_reopen_channel(
         locksroot,
     ).transact({'from': A})
 
-    # Reopening the channel should work iff channel is settled
+    # Reopening the channel should work if channel is settled
     token_network.functions.openChannel(A, B, settle_timeout).transact()
```
err508: That second `f` actually has a meaning, pls readd.

hackaugusto: maths people

loredanacirstea: Next time, please explain the reasoning when making a comment or use `if and only if`. I spent 1 minute looking at the screen wondering if it's joke or not.

LefterisJP: You are not the first nor the last person to try and correct `iff` as a typo, no worries :P

err508: Ahhh angry engineers everywhere

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -205,6 +211,18 @@ contract TokenNetwork is Utils {
...
```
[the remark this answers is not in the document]
loredanacirstea: Good point, @LefterisJP - should I add an overflow check for the counter here just in case? So we limit the channel creation to `2^256 - 1`?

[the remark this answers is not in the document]
LefterisJP: yep -- no reason for a check here, that would be theoretically impossible.

Issue #132, opened 2018-06-18
Add tests for batch unlock - locked vs. unlocked varying amounts per participant


loredanacirstea:
A participant `P1`'s `locked_amount1` (from `P1`'s balance proof) actually contains:
- `amount_that_can_be_unlocked` - if the `secret` was revealed => `P2` must receive this amount
- `amount_that_cannot_be_unlocked` - if the `secret` was not revealed => `P1` must receive this amount back, because the transfer was never finalized.

We must add tests for the settlement calculations that will ensure we compute the locked amount properly, in a way that the batch unlock has enough tokens left in the contract for it to be resolved fairly.

Cases:
- a very high `amount_that_can_be_unlocked` with a very low `amount_that_cannot_be_unlocked`
- a very low `amount_that_can_be_unlocked` with a very high `amount_that_cannot_be_unlocked`

loredanacirstea:
Test cases:

A. Both balance proofs are valid and are the last known balance proofs.

```
Participant P1
deposit1 = 10
withdrawn1 = 0
transferred1 = 0
locked1 = 25

Participant P2
deposit2 = 5
withdrawn2 = 0
transferred2 = 0
locked2 = 15
```
1.
- all secrets are revealed for the pending transfers P1 -> P2, `locked1` must be transferred to `P2`
- all secrets are revealed for the pending transfers P2 -> P1, `locked2` must be transferred to `P1`
2. 
- none of the secrets are revealed for the pending transfers P1 -> P2, `locked1` must be transferred to `P1`
- all secrets are revealed for the pending transfers P2 -> P1, `locked2` must be transferred to `P1`
3. 
- all secrets are revealed for the pending transfers P1 -> P2, `locked1` must be transferred to `P2`
- none of the secrets are revealed for the pending transfers P2 -> P1, `locked2` must be transferred to `P2`

Note: both 2 & 3 are needed, due to the current settlement algorithm, which computes the participant amounts based on whether `transferred1 + locked1` is >= or <= than `transferred2 + locked2`

B. One of the balance proofs is old. We take `P1`s balance proof as being outdated.

```
Participant P1
deposit1 = 10
withdrawn1 = 0
transferred1 = 0
locked1 = 25

Participant P2
deposit2 = 5
withdrawn2 = 0
transferred2 = 0
locked2 = 100
```
Same 1, 2, 3 cases as for A.

loredanacirstea:
[...]

PR #177, opened 2018-07-06
Add edge test cases for locking and unlocking tokens in TokenNetwork

loredanacirstea:
Fixes https://github.com/raiden-network/raiden-contracts/issues/132

Note - I am trying a general approach here, to not construct the tests after the current implementation, but make sure the test cases are valid even with another implementation.
Feedback on the test cases - to see if they are general & make sense, will be appreciated.

Test cases:

A. Both balance proofs are valid and are the last known balance proofs.

```
Participant P1
deposit1 = 10
withdrawn1 = 0
transferred1 = 0
locked1 = 25

Participant P2
deposit2 = 5
withdrawn2 = 0
transferred2 = 0
locked2 = 15
```
1.
- all secrets are revealed for the pending transfers P1 -> P2, `locked1` must be transferred to `P2`
- all secrets are revealed for the pending transfers P2 -> P1, `locked2` must be transferred to `P1`
2.
- none of the secrets are revealed for the pending transfers P1 -> P2, `locked1` must be transferred to `P1`
- all secrets are revealed for the pending transfers P2 -> P1, `locked2` must be transferred to `P1`
3.
- all secrets are revealed for the pending transfers P1 -> P2, `locked1` must be transferred to `P2`
- none of the secrets are revealed for the pending transfers P2 -> P1, `locked2` must be transferred to `P2`
4.
- none of the secrets are revealed for the pending transfers P1 -> P2, `locked1` must be transferred to `P1`
- none of the secrets are revealed for the pending transfers P2 -> P1, `locked2` must be transferred to `P2`

Note: both 2 & 3 are needed, due to the current settlement algorithm, which computes the participant amounts based on whether `transferred1 + locked1` is >= or <= than `transferred2 + locked2`

B. One of the balance proofs is old. We take `P1`s balance proof as being outdated.

```
Participant P1
deposit1 = 10
withdrawn1 = 0
transferred1 = 0
locked1 = 25

Participant P2
deposit2 = 5
withdrawn2 = 0
transferred2 = 0
locked2 = 100
```
Same 1, 2, 3, 4 cases as for A.

C. Another test case where deposits are bigger than locked amounts, so we can check that all the locked locks go to the correct participants depending on the secrets being revealed.

### DIFF
--- a/raiden_contracts/tests/test_channel_settle_unlock_edge_cases.py
+++ b/raiden_contracts/tests/test_channel_settle_unlock_edge_cases.py
@@ -516,7 +516,7 @@ def test_unlock_valid_valid_notrevealed_notrevealed(
     assert custom_token.functions.balanceOf(A).call() == pre_balance_A + 15
     assert custom_token.functions.balanceOf(B).call() == pre_balance_B
     assert custom_token.functions.balanceOf(
-        token_network.address
+        token_network.address,
     ).call() == pre_balance_contract - 15
 
 
@@ -602,7 +602,7 @@ def test_unlock_valid_valid_revealed_notrevealed_locked_smaller_than_deposit(
     assert custom_token.functions.balanceOf(A).call() == pre_balance_A + values_A.deposit - 25
     assert custom_token.functions.balanceOf(B).call() == pre_balance_B + values_B.deposit - 15
     assert custom_token.functions.balanceOf(
-        token_network.address
+        token_network.address,
     ).call() == pre_balance_contract - values_A.deposit - values_B.deposit + 40
 
     pre_balance_A = custom_token.functions.balanceOf(A).call()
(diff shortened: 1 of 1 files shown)


raiden_contracts/tests/test_channel_settle_unlock_edge_cases.py
```diff
@@ -0,0 +1,1227 @@
...
+    # When doing the calculations, we simplify the logic by assuming that the pending transfers
+    # were finalized on-chain through a secret registration.
+    # Therefore, we assume a final balance of 75 tokens being owed by B to A
+    # In this case, A also has a deposit of 10 tokens ->  considering the total available deposit,
+    # A can receive a max amount of 15 tokens.
+    # We call it a max amount, because we don't know if the pending transfers were indeed
+    # finalized or not.
+    # These 15 tokens will be kept in the smart contract after settlement as a locked amount.
+
+    # The only person that might have something to loose at this point is A, in case the secrets
```
LefterisJP: Here and in the rest of the comments `loose` -> `lose`

loredanacirstea:
Closing this, because test cases are included in https://github.com/raiden-network/raiden-contracts/pull/189.

PR #178, opened 2018-07-06
Prevent needless recompilation of precompiled contracts

### LINKED ISSUE #179: Fix using precompiled contracts
Even if a precompiled contracts archive exists setup.py will try to re-compile on install.

PR: #178

ulope:
Previously `setup.py` would always trigger a recompilation of `contracts.json`. Even when installing from an `sdist` (where no `solc` might be available). This defeats the purpose of having the precompiled archive in the first place.

This adds a checksum to the precompiled `contracts.json` in order to prevent recompilation attempts in cases where the existing archive already exists.

Also changes the contracts storage to gzipped JSON to save space (~250k vs. 25k).

Also also switches Travis to Python 3.6 to be in line with the other Raiden repos.

Fixes #179

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ venv/
 # editors
 .vscode/
 
+raiden_contracts/data/contracts.json.gz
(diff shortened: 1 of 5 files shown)


ulope:
Right, thanks. I always forget that one.

PR #182, opened 2018-07-09
Deploy fixes / updates

ulope:
Some fixes / updates discovered / used while deploying the contracts

### DIFF
--- a/raiden_contracts/deploy/__main__.py
+++ b/raiden_contracts/deploy/__main__.py
@@ -1,6 +1,10 @@
 """
 A simple Python script to deploy compiled contracts.
 """
+import logging
+from logging import getLogger
+from getpass import getpass
+
 import click
 import json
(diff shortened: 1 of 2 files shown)


raiden_contracts/deploy/__main__.py
```diff
@@ -97,74 +110,86 @@ def main(**kwargs):
     print('Owner is', owner)
     assert web3.eth.getBalance(owner) > 0, 'Account with insuficient funds.'
 
-    transaction = {'from': owner}
-    if gas_price == 0:
-        transaction['gasPrice'] = gas_price
+    password = getpass(f'Enter password for {owner}:')
```
loredanacirstea: I assume this works well with a node that's running with an unlocked account - meaning, it doesn't prompt you for the password.

ulope: Hm, good question, I'll test it later.

Issue #188, opened 2018-07-13
Analysis of the settlement algorithm & attack vectors


loredanacirstea:
I have analyzed some edge cases that could have made attacks possible and demonstrated that there cannot be attacks in those specific cases (until proven wrong).

I cannot guarantee that all edge cases have been analyzed. The logic & calculations presented in my 6 comments below should also be verified by others.

TODO:
I have started by analyzing the current settlement algorithm. It would be worth to also analyze the edge cases irrespective of the current implementation, as mentioned in https://github.com/raiden-network/raiden-contracts/issues/188#issuecomment-405214043 and https://github.com/raiden-network/raiden-contracts/issues/188#issuecomment-407406686

loredanacirstea:
# Settlement algorithm

## Definitions:

- `valid last BP` = a balance proof that respects the official Raiden client constraints and is the last balance proof known.
- `valid old BP` = a balance proof that respects the official Raiden client constraints, but there are other newer balance proofs that were created after it (additional transfers happened)
- `invalid BP` = a balance proof that does not respect the official Raiden client constraints; E.g. someone else implements a malicious Raiden client and convinces people to use it.

```
Channel participants: P1, P2
D1 = P1's deposit
W1 = P1's withdrawn amount
T1 = P1's transferred amount to P2
L1 = P1's locked amount in pending transfers to P2
Lc1 = P1's locked amount in pending transfers to P2, that can be claimed by P2 (locks secrets have been registered on-chain, in the Secret Registry)
Lu1 = P1's locked amount in pending transfers to P2, that is unclaimable by P2 (locks secrets have not been registered on-chain), therefore belongs to P1
L1 = Lc1 + Lu1

TAD = total available deposit

B1 = total final amount that must be received by P1 after channel settlement and unlock
S1 = amount received by P1 at settlement
SL1 = amount locked in the TokenNetwork contract at settlement time, representing L1
B1 = S1 + SL1

AB1 = available balance for P1; determines if P1 can make additional transfers to P2 or not
```

## Constraints
```
(1) TAD = D1 + D2 - W1 - W2 ; TAD >= 0
(2) B1 = D1 - W1 + T2 - T1 + Lc2 - Lc1
(3) B2 = D2 - W2 + T1 - T2 + Lc1 - Lc2
(4) B1 + B2 = TAD
```

## Constraints that must be enforced by the Raiden client
```
(5) AB1 = D1 - W1 + T2 - T1 - L1; AB1 >= 0
[...]

PR #189, opened 2018-07-16
Settle unlock tests refactor

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/132

Linked issue: https://github.com/raiden-network/raiden-contracts/issues/188. This PR implements tests for old balance proofs. I manually created examples for the old balance proofs that can exist for a pair of last valid balance proofs.

- added values for balance proofs edge cases, mimicking old balance proofs
- removed the tests that only tried to mimic old balance proofs with the same `transferred + claimable_locked` sum (added manual values for this); these tests are redundant and they were not implemented correctly; it is hard to implement this correctly, because there are constraints on the maximum locked amount that can exist in a valid balance proof
- added balance tests to make sure that with an old balance proof, the attacker cannot receive more tokens that he is supposed to (at the end of the channel lifecycle)

### DIFF
--- a/raiden_contracts/tests/fixtures/channel.py
+++ b/raiden_contracts/tests/fixtures/channel.py
@@ -255,7 +255,7 @@ def create_settled_channel(
 @pytest.fixture()
 def reveal_secrets(web3, secret_registry_contract):
     def get(tx_from, transfers):
-        for (expiration, amount, secrethash, secret) in transfers:
+        for (expiration, _, secrethash, secret) in transfers:
             assert web3.eth.blockNumber < expiration
             secret_registry_contract.functions.registerSecret(secret).transact({'from': tx_from})
             assert secret_registry_contract.functions.getSecretRevealBlockHeight(secrethash).call() == web3.eth.blockNumber
--- a/raiden_contracts/tests/fixtures/channel.py
+++ b/raiden_contracts/tests/fixtures/channel.py
@@ -16,6 +16,7 @@ from raiden_contracts.utils.sign import (
 from raiden_contracts.tests.utils import (
     get_settlement_amounts,
     get_onchain_settlement_amounts,
+    get_expected_after_settlement_unlock_amounts,
     ChannelValues,
     get_participants_hash,
 )
@@ -470,6 +471,30 @@ def settle_state_tests(token_network, custom_token):
 
 
 @pytest.fixture()
+def after_settle_unlock_balance_tests(token_network, custom_token):
+    def get(
+            A,
+            values_A,
+            B,
+            values_B,
+            pre_settlement_balance_A,
+            pre_settlement_balance_B,
+            pre_settlement_contract
+    ):
+        # Calculate how much A and B should receive when knowing who will receive the locked amounts
+        (balance_A, balance_B) = get_expected_after_settlement_unlock_amounts(values_A, values_B)
+
+        # Make sure the correct amount of tokens has been transferred
+        account_balance_A = custom_token.functions.balanceOf(A).call()
+        account_balance_B = custom_token.functions.balanceOf(B).call()
+        balance_contract = custom_token.functions.balanceOf(token_network.address).call()
+        assert account_balance_A == pre_settlement_balance_A + balance_A
+        assert account_balance_B == pre_settlement_balance_B + balance_B
+        assert balance_contract == pre_settlement_contract - balance_A - balance_B
+    return get
+
+
+@pytest.fixture()
 def unlock_state_tests(token_network):
     def get(
             A,
@@ -494,6 +519,7 @@ def unlock_state_tests(token_network):
             locksroot_A
         ).call() == 0
     return get
+
 
 @pytest.fixture()
 def withdraw_state_tests(custom_token, token_network):
--- a/raiden_contracts/tests/fixtures/channel.py
+++ b/raiden_contracts/tests/fixtures/channel.py
@@ -19,6 +19,8 @@ from raiden_contracts.tests.utils import (
     get_expected_after_settlement_unlock_amounts,
     ChannelValues,
     get_participants_hash,
+    are_balance_proofs_valid,
+    is_balance_proof_old,
 )
 from raiden_contracts.utils.merkle import EMPTY_MERKLE_ROOT
 from .token_network import *  # flake8: noqa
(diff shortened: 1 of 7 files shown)


raiden_contracts/tests/test_channel_settle.py
```diff
@@ -85,124 +79,72 @@ def test_settle_channel_state(
...
-        channel_test_values,
 ):
-    number_of_channels = 5
-    accounts = get_accounts(2 * number_of_channels)
-    (vals_A0, vals_B0) = channel_test_values
-
-    # We mimic old balance proofs here, with a high locked amount and lower transferred amount
-    # We expect to have the same settlement values as the original values
-
-    def equivalent_transfers(balance_proof):
```
loredanacirstea: These mimicked balance proofs were removed, because they were not correct in the first place. There are additional constraints imposed on the locked amount, that should be `<=` total available deposit and `<=` than the available balance at a certain point in time. This is harder to mimic. Plus - they also took a lot of testing time (~10-15 min)

To replace them, I have added test values in `channel_settle_test_values` from `channel_test_values.py`. E.g.:
```
            # participant1 provides an old participant2 balance proof, but with the same
            # transferred + claimable_locked
            # This should have the same final participant balances as the valid last balance proofs
            ChannelValues(
                deposit=40,
                withdrawn=10,
                transferred=19994,
                claimable_locked=40,
                unclaimable_locked=2,
            ),
```

raiden_contracts/tests/test_channel_settle.py
```diff
@@ -75,7 +70,6 @@ def test_settle_no_bp_success(
     ).transact({'from': A})
 
 
-@pytest.mark.parametrize('channel_test_values', channel_settle_test_values)
 def test_settle_channel_state(
```
loredanacirstea: I left this test as a basic state test with 1 balance proof pair, following the pattern for all other tests.
Moved the more comprehensive tests for old balance proofs with various values in `test_channel_settle_unlock_state.py`, also testing the balances at the end of the channel lifecycle - so not only for `settleChannel`, but also for `unlock`.

raiden_contracts/tests/fixtures/channel_test_values.py
```diff
@@ -1,81 +1,475 @@
...
+# We must cover the edge cases documented in
+# https://github.com/raiden-network/raiden-contracts/issues/188
+# The scope is to make sure that if someone uses an old balance proof, this cannot be used as
+# an attack to steal tokens.
+# For invalid balance proofs (created and signed with an unofficial Raiden client),
+# we cannot determine and guarantee corectness. There are specific constraints that the
+# Raiden client must enforce that guarantee correctness.
+
+# For all valid last balance proofs provided we will programatically construct a suit of balance
+# proof pairs that could be used for attacks (documented edge cases).
```
loredanacirstea: `For all valid last balance proofs provided we will programatically construct a suit of balance` is outdated, I removed that test.

jomuel:
Reviewing this now

raiden_contracts/tests/test_channel_settle_unlock_state.py
```diff
@@ -0,0 +1,492 @@
...
+                    (vals_A, vals_B, 'invalid'),
+                    expected_settlement0,
+                    expected_settlement_onchain0,
+                    expected_final_balance_A0,
+                    expected_final_balance_B0,
+                )
+
+
+@pytest.mark.parametrize('channel_test_values', channel_settle_test_values)
+def test_channel_settle_old_old_balance_proof_values2(
```
jomuel: As far as I can see this and the last two tests are identical except for the subsets of results they check. Can't we avoid the code duplication by turning the whole setup into a fixture, or turning the three tests into one parametrized one?

raiden_contracts/tests/fixtures/channel.py
```diff
@@ -698,9 +704,11 @@ def get(
 
 
 def call_settle(token_network, channel_identifier, A, vals_A, B, vals_B):
-    assert vals_B.transferred + vals_B.locked >= vals_A.transferred + vals_A.locked
+    A_total_transferred = vals_A.transferred + vals_A.locked
+    B_total_transferred = vals_B.transferred + vals_B.locked
+    assert B_total_transferred >= B_total_transferred
```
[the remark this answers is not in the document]
loredanacirstea: Good catch!

raiden_contracts/tests/test_channel_settle_unlock_state.py
loredanacirstea: About `test_channel_settle_valid_old_balance_proof_values`, `test_channel_settle_old_old_balance_proof_values` and `test_channel_settle_old_old_balance_proof_values2`:

- all the balance proof value pairs cannot be run together under the same test because travis fails with `no output for 10 min` and looking into this, I did not find a way to make travis wait except for `travis_wait`, which hides the logs until the job finishes (really not useful because you cannot pick up on failed tests early), so I just split the tested values because I thought the remaining code is not that much and did not want to add too many fixtures (add more magic to this)
- they both use `test_settlement_outcome` as a fixture for the bulk of the tests
- tried by making a fixture with the balance proof pairs, but the tests ended up running for > 2h - so we cannot do a test/test_value.

But I am open to suggestions here - I had a bit of a hard time to do the setup.

jomuel: If we don't want to rewrite much or add more fixtures, what comes to my mind is a single test function with an additional argument `tested_range` that is parametrized like this:
```
@pytest.mark.parametrize('tested_range', ('valid_old', 'old_old1', 'old_old2'))

# ... setup

if tested_range is 'valid_old':
    # ... assertions from the first test
elif tested_range is 'old_old1':
    # ... assertions from the second test
else:
    # ... assertions from the third test
```
A bit unusual but also well readable... Otoh the duplicated code is really not that much and can stay as it is instead.

loredanacirstea: Is it ok for now to leave it as is and you can make a PR implementing the parametrized range after?

jomuel: Sure. I'll approve then.

raiden_contracts/tests/fixtures/channel_test_values.py
```diff
@@ -1,81 +1,475 @@
...
+            # participant2 provides an old participant1 balance proof, with a higher
+            # unclaimable locked amount can happen if expired transfers are removed
+            # from the merkle tree
+            # ChannelValues(
+            #     deposit=35,
+            #     withdrawn=5,
+            #     transferred=20020,
+            #     claimable_locked=3,
+            #     unclaimable_locked=12,
+            # ),
```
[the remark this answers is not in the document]
loredanacirstea: Opened https://github.com/raiden-network/raiden-contracts/issues/271

raiden_contracts/tests/fixtures/channel_test_values.py
loredanacirstea: Fixed in https://github.com/raiden-network/raiden-contracts/pull/272

raiden_contracts/tests/test_channel_settle_unlock_state.py
loredanacirstea: Opened https://github.com/raiden-network/raiden-contracts/issues/273 if you want to handle it.

PR #190, opened 2018-07-18
Adds wrapper for getSettleTransferAmounts.

err508:
We must decide if we wanna change the visibility of `getSettleTransferAmounts` to make it testable. 
Alternative would be to cp+pst it to `TokenNetworkInternalsTest` and keep it synced - danger to get green tests due testing an outdated function.


loredanacirstea:
Nevermind my comment from https://github.com/raiden-network/raiden-contracts/pull/190#event-1777930180, was made with a bad logic - thinking that deployed contracts can be inherited.
But this should nonetheless be accepted after tests pass.

Issue #197, opened 2018-07-30
Ideally, we should use modifiers for function input checks


loredanacirstea:
description is WIP

loredanacirstea:
As explained in the above description, we cannot use modifiers with the current contracts. 
I would move this issue to `Ithaca`. It will probably be solved after https://github.com/raiden-network/raiden-contracts/issues/215

@LefterisJP ?

LefterisJP:
Sounds good @loredanacirstea. So seems the splitting detailed in #215 is gonna be rather needed.

Issue #193, opened 2018-07-23
Use channel_identifier instead of the participant addresses pairs in the channel functions


loredanacirstea:
# Issues

As pointed out by @hackaugusto , the channel lifecycle is unclear, the `TokenNetwork` contract does not have a consistent interface for determining a channel's state.

One problem is that the two participants can reopen the channel after it was settled. The channel will have a different `channel_identifier` than the previous one. If this is not clear in the smart contract function calls and events, it introduces issues in the Raiden client, because the channel lifecycle is hard to track down in a clear & easy way.

(**update**: the following is not solved by neither proposed solutions from this description. It is solved by either refactoring data structures or ABI consistency : https://github.com/raiden-network/raiden-contracts/issues/201). Currently there is a lot of confusion when calling functions because we need to provide both participants, but usually one of the participants also has a role and the caller needs to pay attention at the participant addresses order. This has brought even myself a lot of pain when testing.

Another problem, as mentioned in https://github.com/raiden-network/raiden-contracts/pull/39#discussion_r183750230, the channel lifecycle is not clear also due to the `unlock` data that remains in the contract after the `settleChannel` transaction, while the channel data is removed.

## Current implementation 

Now we use the `participant`, `partner` addresses for determining a channel in the `TokenNetwork` functions and the `ChannelUnlocked` event (all events except `ChannelUnlocked` have the `channel_identifier` as an argument).
E.g.
```
function setTotalDeposit(address participant, uint256 total_deposit, address partner)

event ChannelUnlocked(
        address indexed participant,
        address indexed partner,
        bytes32 indexed locksroot,
        uint256 unlocked_amount,
        uint256 returned_tokens
    );
```
Current `channel_identifier` is defined as `channel_identifier = keccak256(keccak256(participant1, participant2), channel_counter)`, where `participant1, participant2` are in lexicographic order.

We currently have three getters for the channel state, which is confusing for getting the actual state from older channels that still have `unlock` data on chain. You can't actually check if a channel corresponding to a `channel_identifier` has an `unlock` to do or not.
```
# returns channel_identifier, settle_block_number, state (1 = open, 2 = closed, 0 = non-existent or settled)
function getChannelInfo(address participant1, address participant2)

# returns deposit, withdrawn_amount, is_the_closer, balance_hash, nonce
function getChannelParticipantInfo(address participant, address partner)

# returns locked_amount
function getParticipantLockedAmount(address participant1, address participant2, bytes32 locksroot)
```

[...]

PR #198, opened 2018-07-30
Channel identifier interface

loredanacirstea:
Main change discussed in https://github.com/raiden-network/raiden-contracts/issues/193

This PR does not replace the address pair. It adds the `channel_identifier` to the all the events and channel-related functions, except `openChannel`.

This should solve the issues from the Raiden client and make the interaction with the smart contracts clearer.
It also, makes it easier to use `modifier`s for all functions in the future, due to having less variables in the `stack`. More detailed explanation here: https://github.com/raiden-network/raiden-contracts/issues/197

Changes:
```
# Events

event ChannelUnlocked(
        address indexed participant,
        address indexed partner,
        bytes32 locksroot,
        uint256 unlocked_amount,
        uint256 returned_tokens
    );

# Getters

# ! returns (settle_block_number, state)
function getChannelInfo(uint256 channel_identifier) view external returns (uint256, uint8)

function getChannelParticipantInfo(uint256 channel_identifier, address participant)

# State changing transactions

function setTotalDeposit(
        uint256 channel_identifier,
        address participant,
        uint256 total_deposit,
        address partner
    )

function setTotalWithdraw(
        uint256 channel_identifier,
        address participant,
        uint256 total_withdraw,
        address partner,
        bytes participant_signature,
        bytes partner_signature
    )

function closeChannel(
        uint256 channel_identifier,
        address partner,
        bytes32 balance_hash,
        uint256 nonce,
        bytes32 additional_hash,
        bytes signature
    )

function updateNonClosingBalanceProof(
        uint256 channel_identifier,
        address closing_participant,
        address non_closing_participant,
        bytes32 balance_hash,
        uint256 nonce,
        bytes32 additional_hash,
        bytes closing_signature,
        bytes non_closing_signature
    )

function settleChannel(
        uint256 channel_identifier,
        address participant1,
        uint256 participant1_transferred_amount,
        uint256 participant1_locked_amount,
        bytes32 participant1_locksroot,
        address participant2,
        uint256 participant2_transferred_amount,
        uint256 participant2_locked_amount,
        bytes32 participant2_locksroot
    )

function unlock(
        uint256 channel_identifier,
        address participant,
        address partner,
        bytes merkle_tree_leaves
    )

function cooperativeSettle(
        uint256 channel_identifier,
        address participant1_address,
        uint256 participant1_balance,
        address participant2_address,
        uint256 participant2_balance,
        bytes participant1_signature,
        bytes participant2_signature
    )


```


loredanacirstea:
Note: if this is approved, I will open a separate issue and PR for adding a modifier/check to all functions that use `channel_identifier`:
```
modifier validChannelIdentifier(uint256 channel_identifier) {
        require(channel_identifier > 0 && channel_identifier <= channel_counter);
        _;
    }
```

loredanacirstea:
We decided to handle the 2rd party services for Ithaca, so for now:
I moved the Monitoring Service contracts to another folder to avoid the effort of changing them right now. MS test was skipped in a previous commit.
If this is merged, I will open an issue if necessary.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -748,18 +763,33 @@ contract TokenNetwork is Utils {
             require(token.transfer(partner, returned_tokens));
         }
 
-        emit ChannelUnlocked(participant, partner, computed_locksroot, unlocked_amount, returned_tokens);
-
-        // Channel must be settled and channel data deleted
-        require(channels[channel_identifier].state == 0);
```
loredanacirstea: I removed this because the `channel_identifier` argument can be anything, including `0` or `> channel_counter`, making this check irrelevant.
`unlock` data discussion here: https://github.com/raiden-network/raiden-contracts/issues/202

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -693,28 +707,29 @@ contract TokenNetwork is Utils {
...
     function unlock(
+        uint256 channel_identifier,
         address participant,
         address partner,
         bytes merkle_tree_leaves
     )
         public
     {
+        // Channel represented by channel_identifier must be settled and channel data deleted
+        require(channel_identifier != getChannelIdentifier(participant, partner));
```
loredanacirstea: This is essentially irrelevant for the current `unlock` implementation. `channel_identifier` can be any value that the client sets. We don't currently keep the `channel_identifier` in storage at this point in the channel state.
It is nice to have the `channel_identifier` in the `ChannelUnlock` event for the Raiden client, as discussed.

Therefore, we can continue the `unlock` discussion here: https://github.com/raiden-network/raiden-contracts/issues/202

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -243,22 +244,27 @@ contract TokenNetwork is Utils {
...
+    function setTotalDeposit(
+        uint256 channel_identifier,
+        address participant,
+        uint256 total_deposit,
+        address partner
+    )
+        isOpen(channel_identifier)
         public
     {
+        require(channel_identifier == getChannelIdentifier(participant, partner));
```
hackaugusto: okay, so if I understood correctly we cannot have just the channel identifier because we need to add the deposit of both participants and the addresses of each participant are not available, correct?

loredanacirstea: Correct.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -243,22 +244,27 @@ contract TokenNetwork is Utils {
...
+    /// @param channel_identifier Identifier for the channel on which this operation takes place.
     /// @param participant Channel participant whose deposit is being set.
     /// @param total_deposit The total amount of tokens that the participant will have
     /// as a deposit.
     /// @param partner Channel partner address, needed to compute the channel identifier.
-    function setTotalDeposit(address participant, uint256 total_deposit, address partner)
-        isOpen(participant, partner)
+    function setTotalDeposit(
+        uint256 channel_identifier,
+        address participant,
```
LefterisJP: So as Augusto also [commented](https://github.com/raiden-network/raiden-contracts/pull/198/files#r206564742) I find it a bit weird that we have this interface now.

I had assumed that with the introduction of the channel identifier we would no longer need the `address participant` in the function call interface for the contracts and thought that just passing `channel_identifier` would be enough.

Looking at the contract I see that the [Participant](https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/contracts/TokenNetwork.sol#L48) struct does not have the address and thus I guess it's an attempt to optimize for gas by not saving the participant address in the state and instead giving it as argument?

Is there any other reason that is blocking us from using only the single `channel_identifier` or could deoptimizing for gas for this case allow for a cleaner interface?

loredanacirstea: > I had assumed that with the introduction of the channel identifier we would no longer need the address participant in the function call interface for the contracts and thought that just passing channel_identifier would be enough.

My bad. We still need the addresses.

>  attempt to optimize for gas by not saving the participant address in the state and instead giving it as argument?

Yes, gas optimization was always a requirement as far as I know, unless it was about over-optimization (e.g. being concerned about `uint248` vs. `uint256` while in development). In the February retreat we were actually discussing having minimal storage (hashes) and giving the values to each function. The current state is a good compromise I think.

Also, this gas optimization is about data structure. If you change the contract's data structure, you need to refactor everything.

I would personally leave the mapping as it is.

LefterisJP: > Yes, gas optimization was always a requirement as far as I know, unless it was about over-optimization

You should optimize for a bottleneck that has been identified but not base the entire architecture around it.

>  you need to refactor everything.

In the contracts only, right?

> I would personally leave the mapping as it is

So just tell me, if you add the participant addresses in the storage can we have a cleaner contract interface for the cost of 2 SSTORE or is it more complicated than that and there is another reason apart from gas optimization?

loredanacirstea: I can think of at least one code complication with storing the participant addresses. You would have to add something like this every time you would need the other address or you need to differentiate between them:

```
function (channel_identifier, participant) {
    address partner;
    if(participant == storage.participant1_address) {
        partner = storage.participant2_address
    } else {
        partner = storage.participant1_address
    }
}
```
Something like the above would be needed in `setTotalDeposit`, `setTotalWithdraw`, `closeChannel`, `updateNonClosingBalanceProof`, `settleChannel` etc.

In my opinion, complicated code is worse than an interface with more arguments.

I remember taking things like this into consideration when I made the decision to implement the participant data as a mapping. Agreed, I should have documented it in writing better.

Also, for the record, this has not been changed since the first commit in this repo: https://github.com/raiden-network/raiden-contracts/commit/921c006a8fac5e556e238808667b48016b4c2bab#diff-1bdf2ff343918b98864fc26df16cc1ebR65

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -1028,39 +1060,31 @@ contract TokenNetwork is Utils {
...
     {
-        uint256 channel_identifier;
-
-        channel_identifier = getChannelIdentifier(participant1, participant2);
         Channel storage channel = channels[channel_identifier];
 
         return (
-            channel_identifier,
             channel.settle_block_number,
             channel.state
```
hackaugusto: We can have a state for settled channels, if `channel.state == 0 && channel_identifier <= channel_counter` the state settled can be returned.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -300,15 +308,13 @@ contract TokenNetwork is Utils {
     )
         external
```
hackaugusto: this is missing the `isOpen` modifier

hackaugusto: Additionally, shouldn't we move the checks `require(channel_identifier == getChannelIdentifier(participant, partner));` to the modifiers? (I didn't check the other PRs for this)

raiden_contracts/contracts/TokenNetwork.sol
loredanacirstea: Agreed and makes sense. I will use a `uint8` for this in the current PR. It can be changed to an `enum` state with https://github.com/raiden-network/raiden-contracts/pull/206.

For the record, we still have `settled_but_pending_unlock` not covered. It can be covered if the changes from https://github.com/raiden-network/raiden-contracts/pull/203 are merged, but it would require additional arguments in the `getChannelInfo` getter - e.g. `function getChannelInfo(uint256 channel_identifier, address participant1, address participant2)`, so the smart contract can check the stored `unlock` data. We can discuss this later I supposed, after this PR and the one above are clarified.

LefterisJP: +1.

With the check augusto suggests we can have a clear view of the channel state from the raiden client side using `getChannelInfo()`

raiden_contracts/contracts/TokenNetwork.sol
loredanacirstea: This is for `setTotalWithdraw`. There is an equivalent:
```
// Channel must be open
require(channel.state == 1);
```
in the code. An equivalent to `isOpen`.

Totally agree on the modifiers. This is why I opened https://github.com/raiden-network/raiden-contracts/issues/197. We currently cannot be consistent with them from the reasons I wrote in the issue. I want to tackle the modifiers separately and not in this PR - are you ok with this?

raiden_contracts/contracts/TokenNetwork.sol
LefterisJP: > Agreed, I should have documented it in writing better so I would not need to re-think this again now.

I never mentioned anything about lack of documentation from your side.

> Also, for the record, this has not been changed since the first commit in this repo: 

That's irrelevant.

> Something like the above would be needed 

I see. Yes that is true, you would need to differentiate between the addresses to figure our which participant is who.

I don't think that such an addition would make the code more complicated and I am sure you would find a nice way to optimize it.

---

But after reading again what I wrote and putting more thought into it I will no longer insist on this. My thinking is the following:

- You, loredana, have put a lot of thought into this, know this code well and your recommendation is it should stay as is.
- 40,000 gas less is not too much, but it's not an insigificant amount either.
- Our priority atm is to go to mainnet asap and refactoring this would add more work, thus take us more time to implement, test, pr review.
- The cleaner interface argument I described [above](https://github.com/raiden-network/raiden-contracts/pull/198#discussion_r206922595) sounded really important to me back then, especially as this was one of my expectations(the other being cleaner channel lifecycle differentiation through the channel identifier) from this PR. But in reality the only consumer of this interface right now is the Raiden client. In the Raiden client we got all the info so data availability is not an obstacle. We are gonna change the proxies anyway so I thought going for a clean interface is the best option but compared to all the other negatives I described above the interface argument sounds more like code purity than practicality which is not what we strive for.

raiden_contracts/contracts/TokenNetwork.sol
hackaugusto: > are you ok with this?

as longs as the code is correct and it's fixed, yes

raiden_contracts/contracts/TokenNetwork.sol
loredanacirstea: Added in https://github.com/raiden-network/raiden-contracts/pull/198/commits/388ff43356f619827691513ee12b301d95a4ab9f

raiden_contracts/contracts/TokenNetwork.sol
loredanacirstea: Ok.

> especially as this was one of my expectations

Yes, I also contributed to that with my mention from https://github.com/raiden-network/raiden-contracts/issues/193:
> Currently there is a lot of confusion when calling functions because we need to provide both participants, but usually one of the participants also has a role and the caller needs to pay attention at the participant addresses order. This has brought even myself a lot of pain when testing.

Added an update there. As I said, my bad.

Issue #196, opened 2018-07-30
Add `nonce` as argument to the `NonClosingBalanceProofUpdated` event


loredanacirstea:
Current event:
```
event NonClosingBalanceProofUpdated(
        uint256 indexed channel_identifier,
        address indexed closing_participant
    );
```

as @hackaugusto suggested:
The balance proof can be updated multiple times. Adding the `nonce` to the event helps the Raiden client & Monitoring Service to easily keep track of balance data updates and compare `nonce`s.

PR #199, opened 2018-07-30
Add nonce as argument to NonClosingBalanceProofUpdated event

loredanacirstea:
Fixes https://github.com/raiden-network/raiden-contracts/issues/196

Needed for keeping track of balance data updates more easily from the Raiden client & Monitoring Service
- add `nonce` as argument to `NonClosingBalanceProofUpdated` event
- fix & extend test for the event

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -130,7 +130,8 @@ contract TokenNetwork is Utils {
 
     event NonClosingBalanceProofUpdated(
         uint256 indexed channel_identifier,
-        address indexed closing_participant
+        address indexed closing_participant,
+        uint256 nonce
     );
 
     event ChannelSettled(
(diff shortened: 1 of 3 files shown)


raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -130,7 +130,8 @@ contract TokenNetwork is Utils {
 
     event NonClosingBalanceProofUpdated(
         uint256 indexed channel_identifier,
-        address indexed closing_participant
+        address indexed closing_participant,
```
hackaugusto: Why is this attribute named `closing_participant`? The closer is exactly the one node that cannot update the balance proof. Can't we name it just `participant`, `counterparty` or something else?

loredanacirstea: The event is
```
event NonClosingBalanceProofUpdated(
        address indexed closing_participant,
        uint256 nonce
    );
```
My thinking was: we are updating the balance data for the `closing_participant`. The `nonce` is also for the `closing_participant`'s balance proof.

About naming, please comment on https://github.com/raiden-network/raiden-contracts/issues/201 (`Additional Ideas` contains some naming suggestions; regardless of choosing simple names rather than more suggestive ones, we should have consistency -> we need to agree on something). 

So, your suggestion here is to have:
```
event NonClosingBalanceProofUpdated(
        address indexed participant,
        uint256 nonce
    );
```
As, it should be clear what role `participant` has. Correct?

hackaugusto: > As, it should be clear what role participant has. Correct?

I don't think it's necessarily clear who's address it is, but my first thought was "closing_participant is wrong, participant is less wrong".

But now I understand your rationale, this is a hard one to name, if I get a good idea I write in the linked issue.

PR #205, opened 2018-08-01
removed unnecessary comments

hackaugusto:


hackaugusto:
> Why do you not like this. I agree it's unnecessary, but it also isn't confusing I think.

These comments are only increasing signal to noise.

In general my opinion is that comments should express what is not expressed in the code, IOW take advantage of the natural language to express what cannot be done with the programming language, otherwise they don't add value only lines.

PR #206, opened 2018-08-01
Added enum type represent the channel state

hackaugusto:
Note: I did not check how much extra gas is being used for the conversion from the enum to uint8

- The first element of the enum starts with the value 0, therefor that
is used for the inexisting state.
- Because of storage optimization the state is stored in a uint8,
therefor the enum value must be converted to from uint8.

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -72,7 +72,7 @@ contract TokenNetwork is Utils {
         Opened,      // 1
         Closed,      // 2
         Settled      // 3; Note: The channel can be at the settled state and
-                     //          have pending unlocks or not
+                     //          may or may not have pending unlocks
     }
 
     struct Channel {
@@ -756,9 +756,9 @@ contract TokenNetwork is Utils {
             returned_tokens
         );
 
-        // After the channel is settled the storage is cleared, therefor the
-        // value will be Gone and not Settled. The value Settled is used for
-        // the external APIs
+        // After the channel is settled the storage is cleared, therefore the
+        // value will be NonExistent and not Settled. The value Settled is used
+        // for the external APIs
         require(channels[channel_identifier].state == ChannelState.NonExistent);
 
         require(computed_locksroot != 0);
@@ -1052,12 +1052,12 @@ contract TokenNetwork is Utils {
     function getChannelInfo(uint256 channel_identifier)
         view
         external
-        returns (uint256, uint8)
+        returns (uint256, ChannelState)
     {
         Channel storage channel = channels[channel_identifier];
-        ChannelState memory state = channel.state;  // This must **not** update the storage
+        ChannelState state = channel.state;  // This must **not** update the storage
 
-        if (state == ChannelState.Gone && channel_identifier > 0 && channel_identifier <= channel_counter) {
+        if (state == ChannelState.NonExistent && channel_identifier > 0 && channel_identifier <= channel_counter) {
             // The channel has been settled, channel data is removed
             // We might still have data stored for a future unlock operation
             state = ChannelState.Settled;
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -68,10 +68,10 @@ contract TokenNetwork is Utils {
     }
 
     enum ChannelState {
-        Gone,    // 0
-        Opened,  // 1
-        Closed,  // 2
-        Settled  // 3; note: the channel can be in a settled, but pending unlock state
+        NonExistent, // 0
+        Opened,      // 1
+        Closed,      // 2
+        Settled      // 3; note: the channel can be in a settled, but pending unlock state
     }
 
     struct Channel {
@@ -218,11 +218,10 @@ contract TokenNetwork is Utils {
         Channel storage channel = channels[channel_identifier];
 
         require(channel.settle_block_number == 0);
-        require(channel.state == ChannelState.Gone);
+        require(channel.state == ChannelState.NonExistent);
 
         // Store channel information
         channel.settle_block_number = settle_timeout;
-        // Mark channel as opened
         channel.state = ChannelState.Opened;
 
         emit ChannelOpened(channel_identifier, participant1, participant2, settle_timeout);
@@ -759,7 +758,7 @@ contract TokenNetwork is Utils {
         // After the channel is settled the storage is cleared, therefor the
         // value will be Gone and not Settled. The value Settled is used for
         // the external APIs
-        require(channels[channel_identifier].state == ChannelState.Gone);
+        require(channels[channel_identifier].state == ChannelState.NonExistent);
 
         require(computed_locksroot != 0);
         require(locked_amount > 0);
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -71,7 +71,8 @@ contract TokenNetwork is Utils {
         NonExistent, // 0
         Opened,      // 1
         Closed,      // 2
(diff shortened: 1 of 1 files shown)


raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -146,7 +149,7 @@ contract TokenNetwork is Utils {
 
     modifier isOpen(address participant, address partner) {
         uint256 channel_identifier = getChannelIdentifier(participant, partner);
-        require(channels[channel_identifier].state == 1);
+        require(channels[channel_identifier].state == uint8(ChannelState.Opened));
```
loredanacirstea: From what I know, `ChannelState` should actually be `uint8`. Did you have a compilation issue?

> "The integer type used is just large enough to hold all enum values" 
https://solidity.readthedocs.io/en/v0.4.24/types.html#enums

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -72,6 +72,12 @@ contract TokenNetwork is Utils {
         uint256 nonce;
     }
 
+    enum ChannelState {
+        Gone,    // 0 = non-existent or settled
```
loredanacirstea: Maybe `None` or `NonExistent` is better than `Gone`?

raiden_contracts/contracts/TokenNetwork.sol
hackaugusto: from the docs linked above:

> They are explicitly convertible to and from all integer types but implicit conversion is not allowed.

raiden_contracts/contracts/TokenNetwork.sol
hackaugusto: Sure, will change to NonExistent

raiden_contracts/contracts/TokenNetwork.sol
hackaugusto: We can remove the explicit casts from the assignments to the `Channel` struct if we change [`uint8 state` to `ChannelState state`](https://github.com/raiden-network/raiden-contracts/blob/53d3c10971a994e77dc5cf463bf89f4e61e118c1/raiden_contracts/contracts/TokenNetwork.sol#L81), but then we would have to add the casts on all the events that use the channel id.

loredanacirstea: I would prefer `ChannelState state;` inside the `Channel` struct with casting in `getChannelInfo` 

>  we would have to add the casts on all the events that use the channel id.

Why?

hackaugusto: > And you don't need casting in getChannelInfo, because you can have returns (uint256, ChannelState)

This is not support with the current ABI, only with the experimental.

loredanacirstea: We are talking about the channel state now.
Are you referring to https://github.com/raiden-network/raiden-contracts/issues/210?

`struct` as a function arguments is a recent addition. Until recently, only `ethers.js` supported this. I have not tested yet with `web3.py`

loredanacirstea: > This is not supported with the current ABI, only with the experimental.

Tested only with `0.4.24+commit.e67f0147.Emscripten.clang`, without an additional `pragma` and it worked.

hackaugusto: > but then we would have to add the casts on all the events that use the channel id.

My bad, I errd. the channel state is **not** used in any of the events that we have now, so for the events this is not a problem.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -763,6 +755,11 @@ contract TokenNetwork is Utils {
             returned_tokens
         );
 
+        // After the channel is settled the storage is cleared, therefor the
```
loredanacirstea: `therefor` -> `therefore`

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -763,6 +755,11 @@ contract TokenNetwork is Utils {
...
+        // value will be Gone and not Settled. The value Settled is used for
```
loredanacirstea: `Gone` -> `NonExistent`

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -763,6 +755,11 @@ contract TokenNetwork is Utils {
...
+        // the external APIs
+        require(channels[channel_identifier].state == ChannelState.NonExistent);
```
hackaugusto: @loredanacirstea This was removed by another PR [ https://github.com/raiden-network/raiden-contracts/commit/e8998ca1e66423b625755ac86e200acfebf2daaf#diff-2f6fb78ac7aaf29aaec1ed8ab59468f8L506 ], should I remove it from here too?

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -1058,12 +1055,12 @@ contract TokenNetwork is Utils {
         returns (uint256, uint8)
     {
         Channel storage channel = channels[channel_identifier];
-        uint8 state = channel.state;
+        ChannelState memory state = channel.state;  // This must **not** update the storage
 
-        if (state == 0 && channel_identifier > 0 && channel_identifier <= channel_counter) {
+        if (state == ChannelState.Gone && channel_identifier > 0 && channel_identifier <= channel_counter) {
```
loredanacirstea: `Gone` -> `NonExistent`

raiden_contracts/contracts/TokenNetwork.sol
loredanacirstea: You can leave it.

raiden_contracts/contracts/TokenNetwork.sol
LefterisJP: TIL: I thought that non-existent was a typo, and that only non-existant is correct but I was wrong as they are both acceptable spellings.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -1058,12 +1055,12 @@ contract TokenNetwork is Utils {
         returns (uint256, uint8)
     {
         Channel storage channel = channels[channel_identifier];
-        uint8 state = channel.state;
+        ChannelState memory state = channel.state;  // This must **not** update the storage
```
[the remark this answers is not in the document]
hackaugusto: why?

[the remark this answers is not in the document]
loredanacirstea: as the compiler error says:
```
TokenNetwork.sol:1058:9: Error: Storage location can only be given for array or struct types.
        ChannelState memory state = channel.state;  // This must **not** update the storage
        ^-----------------------^
```
The default for local variables is `memory`. You can set them to `storage` only if they are `array`s or `struct`s. Otherwise, it gives you an error.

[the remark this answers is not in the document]
LefterisJP: The docs suggest this only applies for complex types: http://solidity.readthedocs.io/en/v0.4.24/types.html#data-location

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -1058,12 +1055,12 @@ contract TokenNetwork is Utils {
         returns (uint256, uint8)
```
loredanacirstea: `returns (uint256, ChannelState)` should work.

raiden_contracts/contracts/TokenNetwork.sol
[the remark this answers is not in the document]
hackaugusto: Hummm, I asusmed that `ChannelState state` would be a reference to the storage, and not a local variable. Thanks

PR #207, opened 2018-08-01
removed redundant require

hackaugusto:
contractExists returns false for the address 0x0 therefore the check is redundant.

The only reason to not do this change is if we think the address 0x0 could have code in the future


loredanacirstea:
I agree with
> The only reason to not do this change is if we think the address 0x0 could have code in the future

My general approach is: if a check is not clearly redundant, I would rather have it.

hackaugusto:
> My general approach is: if a check is not clearly redundant, I would rather have it.

I'm fine with this

Issue #81, opened 2018-05-22
Use error strings in smart contracts


czepluch:
## Description 
Since solidity 0.4.22 error strings are supported for revert and require:
```
pragma solidity ^0.4.22;

contract VendingMachine {
    function buy(uint amount) payable {
        if (amount > msg.value / 2 ether)
            revert("Not enough Ether provided.");
        // Alternative way to do it:
        require(
            amount <= msg.value / 2 ether,
            "Not enough Ether provided."
        );
        // Perform the purchase.
    }
}
```
Adding this to the smart contract will make it easier to debug where things go wrong in a function. It also makes the terrible work flow of removing `require`s to figure out which require is failing obsolete. 

## Tasklist 
- [x] SecretRegistry - no `require`s
- [x] TokenNetworkRegistry (#1498)
- [ ] TokenNetwork (https://github.com/raiden-network/raiden-contracts/pull/1499)
- [x] Controllable (#1492)
- [x] MonitoringService
- [x] OneToN
- [x] ServiceRegistry
- [x] UserDeposit

LefterisJP:
@loredanacirstea Do you think you can add them in this contract change (red eyes testnet 4)? I would say it's good to take advantage of the redeployment to include them.

loredanacirstea:
Yes, added the milestone now.

loredanacirstea:
Just some notes:

- revert error messages are not currently supported by web3.py - https://github.com/ethereum/web3.py/issues/941. When they will be supported, in order to get them, one would need to `eth_call` the contract function and get the encoded error.
- there are EIPs in discussion about standardizing status codes for errors (similar to http), that can also be handled in the EVM - https://eips.ethereum.org/EIPS/eip-1066; This is worth following in my opinion, as it could improve the automatic interaction between smart contracts.

Aside from these mentions, I am ok with just using some error strings right now and re-evaluating them after standards are clarified.

pirapira:
P2 because I heard error strings would have saved our time today.

pirapira:
If we don't hit the bytecode length after enabling cooperativeSettle(), this will be definitely implemented.  So the next task is to try enabling the cooperativeSettle() locally and see if the compilation still works.

pirapira:
Even if we recover `cooperativeSettle()`, we can still add at least a two-character message to each `require`.

pirapira:
[...]

PR #213, opened 2018-08-06
Add revert error strings for TokenNetwork

loredanacirstea:
Fixes https://github.com/raiden-network/raiden-contracts/issues/81

This is in work. The contract currently fails with `Contract code size exceeds EIP170 limit of 24577.  Got code of size: 24781`

We can either remove some of the error messages and keep the very important ones or delay adding them until we have some cleaner code with modifiers (this would remove some of the duplicate error messages).

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -146,12 +146,12 @@ contract TokenNetwork is Utils {
     );
 
     modifier isOpen(uint256 channel_identifier) {
-        require(channels[channel_identifier].state == ChannelState.Opened);
+        require(channels[channel_identifier].state == ChannelState.Opened, 'Invalid channel state');
         _;
     }
 
     modifier settleTimeoutValid(uint256 timeout) {
-        require(timeout >= settlement_timeout_min && timeout <= settlement_timeout_max);
+        require(timeout >= settlement_timeout_min && timeout <= settlement_timeout_max, 'Invalid settlement timeout.');
         _;
     }
 
@@ -169,8 +169,8 @@ contract TokenNetwork is Utils {
         require(_chain_id > 0);
         require(_settlement_timeout_min > 0);
         require(_settlement_timeout_max > _settlement_timeout_min);
-        require(contractExists(_token_address));
-        require(contractExists(_secret_registry));
+        require(contractExists(_token_address), 'Invalid Token.');
+        require(contractExists(_secret_registry), 'Invalid SecretRegistry.');
 
         token = Token(_token_address);
 
@@ -180,7 +180,7 @@ contract TokenNetwork is Utils {
         settlement_timeout_max = _settlement_timeout_max;
 
         // Make sure the contract is indeed a token contract
-        require(token.totalSupply() > 0);
+        require(token.totalSupply() > 0, 'Invalid Token.');
 
         // Try to get token decimals, otherwise assume 18
         bool exists = address(token).call(bytes4(keccak256("decimals()")));
@@ -217,7 +217,7 @@ contract TokenNetwork is Utils {
         pair_hash = getParticipantsHash(participant1, participant2);
 
         // There must only be one channel opened between two participants at any moment in time.
-        require(participants_hash_to_channel_identifier[pair_hash] == 0);
+        require(participants_hash_to_channel_identifier[pair_hash] == 0, 'Channel already open.');
 
         participants_hash_to_channel_identifier[pair_hash] = channel_identifier;
 
@@ -251,9 +251,9 @@ contract TokenNetwork is Utils {
         isOpen(channel_identifier)
         public
     {
-        require(channel_identifier == getChannelIdentifier(participant, partner));
-        require(total_deposit > 0);
-        require(total_deposit <= deposit_limit);
+        require(channel_identifier == getChannelIdentifier(participant, partner), 'Invalid channel identifier.');
+        require(total_deposit > 0, 'Invalid total deposit.');
+        require(total_deposit <= deposit_limit, 'Over bug bounty limit.');
 
         uint256 added_deposit;
         uint256 channel_deposit;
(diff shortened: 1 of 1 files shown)


raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -146,12 +146,12 @@ contract TokenNetwork is Utils {
     );
 
     modifier isOpen(uint256 channel_identifier) {
-        require(channels[channel_identifier].state == ChannelState.Opened);
+        require(channels[channel_identifier].state == ChannelState.Opened, 'Invalid channel state');
```
hackaugusto: I think the expected state should be mentioned in the string, e.g. 'channel state is not open'

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -169,8 +169,8 @@ contract TokenNetwork is Utils {
         require(_chain_id > 0);
         require(_settlement_timeout_min > 0);
         require(_settlement_timeout_max > _settlement_timeout_min);
-        require(contractExists(_token_address));
-        require(contractExists(_secret_registry));
+        require(contractExists(_token_address), 'Invalid Token.');
```
hackaugusto: This uses the same string to the `require(token.totalSupply() > 0)`, to differentiate them I would mention the contract doesnt have code, e.g. 'Token address doesnt have code'

hackaugusto: And if you change the string for the token address, please also update the one for the secret registry

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -301,8 +301,8 @@ contract TokenNetwork is Utils {
     )
         external
     {
-        require(channel_identifier == getChannelIdentifier(participant, partner));
-        require(total_withdraw > 0);
+        require(channel_identifier == getChannelIdentifier(participant, partner), 'Invalid channel identifier.');
+        require(total_withdraw > 0, 'Invalid total withdraw.');
```
hackaugusto: this could also work as a documentation string, why not 'withdraw cannot be zero'? (since the type is uint a negative value is not possible)

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -320,8 +320,8 @@ contract TokenNetwork is Utils {
         participant_state.withdrawn_amount += current_withdraw;
 
         // Do the tokens transfer
-        require(token.transfer(participant, current_withdraw));
-        require(channel.state == ChannelState.Opened);
+        require(token.transfer(participant, current_withdraw), 'Token transfer failed.');
+        require(channel.state == ChannelState.Opened, 'Invalid channel state.');
```
hackaugusto: Here too I would mention the expected channel state.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -332,14 +332,14 @@ contract TokenNetwork is Utils {
...
+        require(current_withdraw > 0, 'Invalid withdrawn amount.');
 
-        // Underflow check
+        // Overflow check
         require(participant_state.withdrawn_amount >= current_withdraw);
         require(participant_state.withdrawn_amount == total_withdraw);
 
         // Entire withdrawn amount must not be bigger than the entire channel deposit
-        require(participant_state.withdrawn_amount <= (total_deposit - partner_state.withdrawn_amount));
+        require(participant_state.withdrawn_amount <= (total_deposit - partner_state.withdrawn_amount), 'Invalid withdrawn amount.');
```
hackaugusto: This string is equal to the one used for the `0` check, I would differentiate among them, e.g. 'withdraw larger than channel deposit'

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -465,16 +465,16 @@ contract TokenNetwork is Utils {
 
         emit NonClosingBalanceProofUpdated(channel_identifier, closing_participant, nonce);
 
-        require(channel.state == ChannelState.Closed);
+        require(channel.state == ChannelState.Closed, 'Invalid channel state.');
```
hackaugusto: Here too I would mention the expected channel state, e.g. 'channel state is not closed'

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -850,19 +850,20 @@ contract TokenNetwork is Utils {
...
         if (participant2_balance > 0) {
-            require(token.transfer(participant2, participant2_balance));
+            require(token.transfer(participant2, participant2_balance), 'Token transfer 2 failed.');
         }
 
         // The provided addresses must be the same as the recovered ones
-        require(participant1 == participant1_address);
-        require(participant2 == participant2_address);
+        require(participant1 == participant1_address, 'Invalid signature 1.');
+        require(participant2 == participant2_address, 'Invalid signature 1.');
```
hackaugusto: typo, `s/1/2/`

loredanacirstea:
@hackaugusto , my initial version was very similar to what you are suggesting above. That version did not compile, due to the mentioned error from my description, so I cut out every channel state etc. in order to make the error message as small as possible, but still trackable.

Still, compilation fails because of the size.

This is why I discontinued work on this PR for now.

hackaugusto:
> This is why I discontinued work on this PR for now.

My bad, I didn't pay attention to the WIP tag.

> my initial version was very similar to what you are suggesting above

Since we are here, we could either

- Use number IDs instead of proper strings to identify the errors. Which makes the code a bit less readable but may fix the problem for now
- Split the TokenNetwork in libraries

loredanacirstea:
> Use number IDs instead of proper strings

There are some discussions about status codes, as mentioned in https://github.com/raiden-network/raiden-contracts/issues/81#issuecomment-410663218
We can have our own. But we should remember that we cannot even use error strings at this point, so there is no rush to include them.

> Split the TokenNetwork in libraries

Yes, I am thinking about this. I will open a discussion about this and the new cheap delegatecall proxy (no need to be upgradable), so we can discuss it.

pirapira:
There was a second attempt.  The bottleneck is the runtime code size.

PR #214, opened 2018-08-07
Add docstring for getChannelInfo params

LefterisJP:
After https://github.com/raiden-network/raiden-contracts/pull/212 there was two additional parameters added to `getChannelInfo` but the docstring was not updated. This PR updates the docstring.

Also it adds a check that the participant addresses are not the same.

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -1058,6 +1058,8 @@ contract TokenNetwork is Utils {
 
     /// @dev Returns the channel specific data.
     /// @param channel_identifier Identifier for the channel on which this operation takes place.
+    /// @param participant1 Address of one of the channel's participants
+    /// @param participant2 Address of one of the other channel participant
     /// @return Channel settle_block_number and state.
     function getChannelInfo(
         uint256 channel_identifier,


raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -1070,6 +1072,7 @@ contract TokenNetwork is Utils {
     {
         bytes32 unlock_key1;
         bytes32 unlock_key2;
+        require(participant1 != participant2);
```
loredanacirstea: I specifically did not add any checks for the `getChannelInfo` getter, to make it able to retrieve info for any combination of identifier and addresses.
It does not change the state.
It can be used to make sure we do not have channel data that is not supposed to be there (e.g. some checks fail in state changing functions).

But I am open to suggestions.

loredanacirstea: I also prefer to have a `getChannelIdentifier` getter without checks, but at the moment this was the best solution I had to integrate those checks while not triggering the compiler error about the local variables limit...

loredanacirstea: > But giving the two same addresses make no sense as a combination, right?

Correct, but just because we enforce this with checks in other state-changing functions. With this, we can do random checks (contract calls) to see if anything is out of place - if needed.

> I have no strong opinions on this and can remove it if you want me to and move it to the raiden client.

I would prefer this, unless there is a better reason to add them.

LefterisJP: No there isn't. Amending the commit.

PR #221, opened 2018-08-08
Reorder functions after Solidity docs guidelines

loredanacirstea:
Reorder functions for better readability, after the Solidity docs guidelines:
https://solidity.readthedocs.io/en/v0.4.24/style-guide.html#order-of-functions

- constructor
- fallback function (if exists)
- external
- public
- internal
- private

We still have some methods that are `external` and have not been moved to the top. Such as:
- state changing: `setTotalWithdraw` , `updateNonClosingBalanceProof`
- getters: `getChannelInfo`, `getChannelParticipantInfo`

I wouldn't move the channel-related state changing functions to the top because we actually have not discussed the visibility modifiers on the channel functions + the functions are ordered by channel state logic, which makes more sense to me.

Open to ideas about the getters. I chose to have them after all the `public` & `external` state changing functions.

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -602,96 +602,6 @@ contract TokenNetwork is Utils {
         );
     }
 
-    /// @dev Function that calculates the amount of tokens that the participants will
-    /// receive when calling settleChannel.
-    function getSettleTransferAmounts(
-        Participant storage participant1_state,
-        uint256 participant1_transferred_amount,
-        uint256 participant1_locked_amount,
-        Participant storage participant2_state,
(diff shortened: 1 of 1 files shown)


czepluch:
I actually saw this in the solidity doc guidelines some time back too. But I always likes that reading the contract also kind of explained the life cycle of a channel. Nonetheless if we are set on being totally compliant with the solidity guidelines, I am also fine with that.

LefterisJP:
I am neutral on the changes but: 

>  Nonetheless if we are set on being totally compliant with the solidity guidelines, I am also fine with that.

No we are not set on that. The solidity guidelines are not a law and they are also opinionated. They are made by people like you and me who just commit on the solidity repo docs.

PR #232, opened 2018-08-10
setTotalWithdraw - use assert for deposit overflow

loredanacirstea:
To be merged later if accepted.

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -380,8 +380,9 @@ contract TokenNetwork is Utils {
         // deposit
         require(participant_state.withdrawn_amount <= (total_deposit - partner_state.withdrawn_amount));
 
-        require(total_deposit >= participant_state.deposit);
-        require(total_deposit >= partner_state.deposit);
+        // This should never happen, as we have an overflow check in setTotalDeposit
+        assert(total_deposit >= participant_state.deposit);
+        assert(total_deposit >= partner_state.deposit);
 
         // A withdraw should never happen if a participant already has a
         // balance proof in storage


raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -378,8 +378,9 @@ contract TokenNetwork is Utils {
         // deposit
         require(participant_state.withdrawn_amount <= (total_deposit - partner_state.withdrawn_amount));
 
-        require(total_deposit >= participant_state.deposit);
-        require(total_deposit >= partner_state.deposit);
+        // This should never happen, as we have an overflow check in setTotalDeposit
+        assert(total_deposit >= participant_state.deposit);
```
LefterisJP: From a quick look `setTotalDeposit()` is not called anywhere inside of `setTotalWithdraw()`. Which makes sense.

But then, I think this change should not happen. As shown [here](http://solidity.readthedocs.io/en/v0.4.23/control-structures.html?highlight=assert#error-handling-assert-require-revert-and-exceptions) it should never be possible to reach an assert. Here would it not be possible to reach the assert by just calling `setTotalWithdraw()` without ever having deposited anything?

loredanacirstea: Those `assert`s should never be triggered. `participant.deposit` is only set in `setTotalDeposit`, where we have these checks already:
https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/contracts/TokenNetwork.sol#L311-L312

So, in `setTotalWithdraw` the overflow should never be the case.

> Here would it not be possible to reach the assert by just calling setTotalWithdraw() without ever having deposited anything?

It would be:
```
total_deposit = participant_state.deposit + partner_state.deposit = 0 + 0;
```

Asserts will never be triggered:
```sh
# equivalent with 0 >= 0
assert(total_deposit >= participant_state.deposit);

# equivalent with 0 >= 0
assert(total_deposit >= partner_state.deposit);
```

PR #240, opened 2018-08-16
Removed address_by_endpoint from the endpoint registry

hackaugusto:
The smart contract function address_by_endpoint allowed a malicious node
to set its address for an endpoint which it did not control. The
functionality was not used and buggy, this removes it.

### DIFF
--- a/raiden_contracts/contracts/EndpointRegistry.sol
+++ b/raiden_contracts/contracts/EndpointRegistry.sol
@@ -7,11 +7,7 @@ contract EndpointRegistry {
     string constant public contract_version = "0.3._";
 
     event AddressRegistered(address indexed eth_address, string endpoint);
-
-    // Mapping of Ethereum addresses => Endpoints
     mapping (address => string) private address_to_endpoint;
-    // Mapping of Endpoints => Ethereum addresses
-    mapping (string => address) private endpoint_to_address;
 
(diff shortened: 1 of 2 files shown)


loredanacirstea:
Yes, thank you @err508 for pointing it out.

@hackaugusto , please rebase.

raiden_contracts/contracts/EndpointRegistry.sol
```diff
@@ -7,11 +7,7 @@ contract EndpointRegistry {
     string constant public contract_version = "0.3._";
 
     event AddressRegistered(address indexed eth_address, string endpoint);
-
-    // Mapping of Ethereum addresses => Endpoints
-    mapping (address => string) private address_to_endpoint;
-    // Mapping of Endpoints => Ethereum addresses
-    mapping (string => address) private endpoint_to_address;
+    mapping (address => string) address_to_endpoint;
```
loredanacirstea: After rebasing the `private` state visibility was lost. It is good practice to explicitly keep it even if the default is indeed `private`. It was added in https://github.com/raiden-network/raiden-contracts/commit/dd0b258d79622211a8605012038d39ef710b0f67

hackaugusto: my bad, I didn't see it while fixing the conflicts.

PR #254, opened 2018-08-23
Reorders assert statements 

err508:
Asserts after all state changes were made, as in #47.

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -281,34 +281,35 @@ contract TokenNetwork is Utils {
         isOpen(channel_identifier)
         public
     {
-        require(channel_identifier == getChannelIdentifier(participant, partner));
-        require(total_deposit > 0);
-        require(total_deposit <= deposit_limit);
-
         uint256 added_deposit;
         uint256 channel_deposit;
 
+        // Check deposit constraints
+        require(total_deposit > 0);
+        require(total_deposit <= deposit_limit);
+
+        // Check that address data is correct
+        require(channel_identifier == getChannelIdentifier(participant, partner));
+
+        // Get channel deposit state
         Channel storage channel = channels[channel_identifier];
         Participant storage participant_state = channel.participants[participant];
         Participant storage partner_state = channel.participants[partner];
 
-        // Calculate the actual amount of tokens that will be transferred
+        // Check for underflow:
+        require(total_deposit > participant_state.deposit);
+        // Calculate the actual amount of tokens that will be transferred:
         added_deposit = total_deposit - participant_state.deposit;
-        // The actual amount of tokens that will be transferred must be > 0
-        require(added_deposit > 0);
-        // Underflow check; we use <= because added_deposit == total_deposit for the first deposit
-        require(added_deposit <= total_deposit);
+
         // This should never fail at this point. Added check for security, because we directly set
         // the participant_state.deposit = total_deposit, while we transfer `added_deposit` tokens.
         assert(participant_state.deposit + added_deposit == total_deposit);
 
+        // Check that overall channel deposit does not overflow
+        require((participant_state.deposit + partner_state.deposit) >= participant_state.deposit);
+
         // Update the participant's channel deposit
         participant_state.deposit = total_deposit;
-
-        // Calculate the entire channel deposit, to avoid overflow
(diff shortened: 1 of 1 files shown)


loredanacirstea:
I do not like this approach, because it makes the code less readable and auditable (we had complaints already). This is why I moved them in the first place here: https://github.com/raiden-network/raiden-contracts/pull/248. Explanations can be found in both PR & issue. 

So, I am not going to approve this. If anyone else wants to approve, I am not against it - reordering the checks in itself does not affect security. This is only a matter of opinion.

err508:
I updated the PR with the goal to incorperate both comments from here and in #248.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -281,35 +281,36 @@ contract TokenNetwork is Utils {
...
+        uint256 added_deposit;
+        uint256 channel_deposit;
+
+        // Check deposit constraints
         require(total_deposit > 0);
         require(total_deposit <= deposit_limit);
 
-        uint256 added_deposit;
-        uint256 channel_deposit;
+        // Check that address data is correct
```
loredanacirstea: Here we check that:
- participant and partner addresses have a current channel opened
- the channel that they have open is the same as the channel identifier given as input

Otherwise, `address data` does not really explain what we are checking.

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -281,35 +281,36 @@ contract TokenNetwork is Utils {
...
+        require(channel_identifier == getChannelIdentifier(participant, partner));
 
+        // Get channel deposit state
         Channel storage channel = channels[channel_identifier];
         Participant storage participant_state = channel.participants[participant];
         Participant storage partner_state = channel.participants[partner];
 
-        // Calculate the actual amount of tokens that will be transferred
+        // Check for underflow:
+        require(total_deposit > participant_state.deposit);
```
loredanacirstea: Checking for underflows before the actual state change is confusing in my opinion. 

As discussed in https://github.com/raiden-network/raiden-contracts/issues/47#issuecomment-413898659,  I think that a computation  should be `followed by the validation of the computation`.

> But I would strive for a contract-based approach, where the beginning of a function starts with the requirements, first validating/requiring the value inputs, then the code is followed by the computation, which is followed by the validation of the computation and external side-effects, and the code finishes with the asserts/invariants that must be true after the execution of the code.

raiden_contracts/contracts/TokenNetwork.sol
err508: I described it as general `address data` to suggest to an auditor, that the actual requires are not in that part of the code but inside of `getChannelIdentifier`. I can add that to the comment to make that more explicit.

raiden_contracts/contracts/TokenNetwork.sol
err508: I followed the approach described in the comment, for me overflow checks should be part of 
> first validating/requiring the value inputs
as we must not perform any compuations on overflowing input values.

err508: > Validating input would be something like if say an integer should not have the value 0 or something like that.

I'd say that is pretty much what we do here. We validate, that `total_deposit` is monotonically increasing. Additionally to saving gas, it makes the code easier to read/audit to me, when we first formulate constraints on the inputs/variables, then directly do computation. However, we can not do this directly after the function is called, but only after reading from the state, so I see why you'd classify this as validation of computation.

pirapira:
@err508 are you still pursuing this PR?

err508:
No, currently not, feel free to pick it up/close it.

pirapira:
As discussed offline, I'll go through this PR, pick up good ideas, and close it.

pirapira:
Since this PR was created, more comments were added on `master`.  Making most additional comments in this PR obsolete.   ` // Check deposit constraints` is still missing, but this is not particularly informative.

As for different places to assert and require conditions,

The proposed structure
```
check(inputs);

do_stuff();
```
makes sense when `do_stuff();` causes some irreversible (or externally observable) changes.  In Solidity, everything is reverted at a `require(...)` failure at the last line.

In Solidity we are more concerned with reentrance attacks.  The current structure
```
do_everything();
require(outer_call());
/// ???
```
is a good practice.  If we want to introduce invariant checks, at  `///???`, read-only `require()` checks can be added.

PR #256, opened 2018-08-30
Sync dependency with raiden-libs

andrevmatos:
Raiden-libs updated its `coincurve` dependency. No breaking changes on this update.


andrevmatos:
The later, from the changelog only some new features from libsecp256k1, packaging and tests changes.

PR #260, opened 2018-09-06
Fix contract compilation during initial install

ulope:
The order in which packages are installed is not guranteed by pip / setuptools therefore we need to include `py-solc` in `setup_requires` to make sure it's available during the initial installation / wheel building in a fresh (v)env. Otherwise it can lead to situations where the `contracts.json.gz` can't be built even though all dependencies are available.

Required by: raiden-network/raiden#2349


CosminNechifor:
Thanks for the quick fix @ulope , this was needed for https://github.com/raiden-network/raiden/pull/2349

PR #267, opened 2018-09-17
Check for solc when doing a release

### LINKED ISSUE #266: Wrong precompiled contracts code in 0.3 release
I ran into some strange issues when trying to decode `ChannelClosed` events in the status page backend inside docker. After some debugging with @rakanalh we found out that the issue comes from including the old 0.2 release ABI in the precompiled ABI file (`contracts.json.gz`).

We need to:
- [ ] Fix the packaging code to generate the proper ABI for a release
- [ ] See if we can replace the 0.3 package or release a 0.3.1 release with the proper precompiled ABI

pcppcp:
- adds `make release` target - solc is required for building a release
  if RAIDEN_SOLC_REQUIRED environment variable is set
- removed check for `py-solc` module (it's a `setup.py` dependency now)

Fixes #266

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,4 +11,7 @@ lint:
 	flake8 raiden_contracts/
 
 clean:
-	rm -rf build/ *egg-info/ raiden_contracts/data/contracts.json dist
+	rm -rf build/ *egg-info/ raiden_contracts/data/contracts.json.gz dist .eggs
+
+release: clean
+	RAIDEN_SOLC_REQUIRED=1 python setup.py sdist bdist_wheel upload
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -11,4 +11,7 @@ lint:
 	flake8 raiden_contracts/
 
 clean:
-	rm -rf build/ *egg-info/ raiden_contracts/data/contracts.json dist
+	rm -rf build/ *egg-info/ raiden_contracts/data/contracts.json.gz dist .eggs
+
+release: clean
+	RAIDEN_SOLC_REQUIRED=1 python setup.py sdist bdist_wheel upload
```
ulope: It's recommended to use [`twine`](https://pypi.org/project/twine/) to upload packages instead of `python setup.py upload`

pcppcp: haha I've noticed this warning when doing test upload, and I wanted create a separate issue for that. I don't have `twine` installed at the moment and I don't want to mess with keyring for passwordless auth, so I guess I'll leave it as it is now.

Issue #263, opened 2018-09-07
Implement the deprecation switch for Red Eyes


LefterisJP:
Implement a deprecation switch in the contracts. Only a specific address given at the token network registry's creation will be able to invoke the switch.

Once the switch is invoked and the contracts are marked as deprecated then users are no longer able to open new channels and not able to deposit any more tokens. The contract is going to stay on chain and user can close the channels to get access to their tokens, as well as continue to transfer tokens within the channels.

Edit for clarity: Once the switch has been activated it can not be de-activated.

loredanacirstea:
@LefterisJP please confirm that the deprecation switch MUST NOT be toggable. So, if you set it from `false` to `true`, it cannot be undone anymore.

LefterisJP:
Confirmed as per our offline discussions.

LefterisJP:
Number (2) -- we can live with this possibility.

PR #275, opened 2018-09-19
Deprecation switch for Bug Bounty release

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/263

To be merged after https://github.com/raiden-network/raiden-contracts/pull/270

### DIFF
--- a/raiden_contracts/contracts/TokenNetwork.sol
+++ b/raiden_contracts/contracts/TokenNetwork.sol
@@ -40,6 +40,10 @@ contract TokenNetwork is Utils {
 
     string public constant signature_prefix = '\x19Ethereum Signed Message:\n';
 
+    // Only for the limited Bug Bounty release
+    address public deprecation_executor;
+    bool public safety_deprecation_switch = false;
+
     // channel_identifier => Channel
     // channel identifier is the channel_counter value at the time of opening
     // the channel
@@ -178,6 +182,16 @@ contract TokenNetwork is Utils {
         uint256 participant2_amount
     );
 
+    modifier onlyDeprecationExecutor() {
+        require(msg.sender == deprecation_executor);
+        _;
+    }
+
+    modifier isSafe() {
+        require(safety_deprecation_switch == false);
+        _;
+    }
+
     modifier isOpen(uint256 channel_identifier) {
         require(channels[channel_identifier].state == ChannelState.Opened);
         _;
@@ -194,12 +208,14 @@ contract TokenNetwork is Utils {
         address _secret_registry,
         uint256 _chain_id,
         uint256 _settlement_timeout_min,
-        uint256 _settlement_timeout_max
+        uint256 _settlement_timeout_max,
+        address _deprecation_executor
     )
         public
     {
         require(_token_address != address(0x0));
         require(_secret_registry != address(0x0));
+        require(_deprecation_executor != address(0x0));
         require(_chain_id > 0);
         require(_settlement_timeout_min > 0);
         require(_settlement_timeout_max > _settlement_timeout_min);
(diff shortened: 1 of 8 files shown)


LefterisJP:
reviewing

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -224,6 +240,13 @@ contract TokenNetwork is Utils {
         }
 
         deposit_limit = 100 * (10 ** uint256(decimals));
+
+        administrator = _administrator;
+    }
+
+    function setSafetyDeprecationSwitch(bool status) onlyAdmin public {
```
LefterisJP: This looks more like a `toggleSafetyDeprecationSwitch()` function. Seems that it can `toggle` between `true` and `false` values for the switch. If we keep with this behaviour use `toggle` in the name.

But also from what I understood from our conversation we want this to be set only once and to not be settable again. Once it's deprecated that's a final action which should not be undoable. 

Let's discuss this also offline.

raiden_contracts/contracts/TokenNetworkRegistry.sol
```diff
@@ -39,13 +47,16 @@ contract TokenNetworkRegistry is Utils {
...
+
+        administrator = msg.sender;
     }
 
     /// @notice Deploy a new TokenNetwork contract for the Token deployed at
     /// `_token_address`.
     /// @param _token_address Ethereum address of an already deployed token, to
     /// be used in the new TokenNetwork contract.
     function createERC20TokenNetwork(address _token_address)
+        onlyAdmin
```
LefterisJP: I had a different idea about how this should be done.
The idea would be to allow only one token network to exist instead of allowing only the administrator to deploy one.

The end result would indeed be the same but would at least agree with our limitations and give less power to the administrator.

raiden_contracts/tests/test_deprecation_switch.py
```diff
@@ -0,0 +1,243 @@
...
+    channel_identifier = create_channel(A, B)[0]
+    channel_deposit(channel_identifier, A, deposit, B)
+    channel_deposit(channel_identifier, B, deposit, A)
+
+    token_network.functions.setSafetyDeprecationSwitch(True).transact({
+        'from': administrator,
+    })
+    assert token_network.functions.safety_deprecation_switch().call() is True
+
+    # Now we cannot deposit in existent channels
```
LefterisJP: nitpick: /s/existent/existing both here and in the rest of the tests below

[source ref](https://ell.stackexchange.com/a/161670)

raiden_contracts/contracts/TokenNetworkRegistry.sol
```diff
@@ -16,11 +16,20 @@ contract TokenNetworkRegistry is Utils {
     uint256 public settlement_timeout_min;
     uint256 public settlement_timeout_max;
 
+    // Only for the limited Bug Bounty release
+    address public deprecation_executor;
+    bool public bug_bounty_token_network_created = false;
```
LefterisJP: please change to `token_network_created` or if you want to include the release name use `red_eyes`.

raiden_contracts/contracts/TokenNetworkRegistry.sol
```diff
@@ -16,11 +16,20 @@ contract TokenNetworkRegistry is Utils {
     uint256 public settlement_timeout_min;
     uint256 public settlement_timeout_max;
 
+    // Only for the limited Bug Bounty release
```
LefterisJP: Add red eyes here to make it clear for which release it is.

Issue #273, opened 2018-09-19
Simplify tests for settle&unlock


loredanacirstea:
As mentioned in https://github.com/raiden-network/raiden-contracts/pull/189#discussion_r218134369,
the tests may be simplified by using parametrized ranges.
Suggestion: https://github.com/raiden-network/raiden-contracts/pull/189#discussion_r218443021

hackaugusto:
Just one note: Usually parametrized tests are way slower to run.

loredanacirstea:
Fixed in https://github.com/raiden-network/raiden-contracts/pull/278

PR #278, opened 2018-09-20
Avoid code duplication in settle&unlock tests

jomuel:
Turn three similiar tests into one parametrized one to avoid a bit of code duplication in the setup. 

Despite our worries in the issue discussion, the tests do not seem to run any slower this way. I ran both versions a couple of times.

Fixes issue #273

### DIFF
--- a/raiden_contracts/tests/test_channel_settle_unlock_state.py
+++ b/raiden_contracts/tests/test_channel_settle_unlock_state.py
@@ -258,13 +258,16 @@ def test_settlement_outcome(
 
 
 @pytest.mark.parametrize('channel_test_values', channel_settle_test_values)
-def test_channel_settle_valid_old_balance_proof_values(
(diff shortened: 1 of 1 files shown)


raiden_contracts/tests/test_channel_settle_unlock_state.py
```diff
@@ -258,13 +258,16 @@ def f(
 
 
 @pytest.mark.parametrize('channel_test_values', channel_settle_test_values)
-def test_channel_settle_valid_old_balance_proof_values(
+@pytest.mark.parametrize('tested_range', ('valid', 'old1', 'old2'))
```
loredanacirstea: From the context of the test, `valid`, `old1` and `old2` are not clear enough. I suggest something like `one_old`, `both_old_1`, `both_old2`.

jomuel: Done.

PR #286, opened 2018-09-25
Added nonce=0 closeChannel test, clarified state test for closeChannel

loredanacirstea:
- added `closeChannel` test, to check that if `nonce = 0`, then no balance data is stored in the `TokenNetwork` contract
- cleaned up `test_close_channel_state` test for `closeChannel` and added ETH & token balance checks

### DIFF
--- a/raiden_contracts/tests/test_channel_close.py
+++ b/raiden_contracts/tests/test_channel_close.py
@@ -209,6 +209,18 @@ def test_close_nonce_zero(
         vals_B.locksroot,
     )
 
+    (
+        _, _,
+        B_is_the_closer,
+        B_balance_hash,
+        B_nonce,
+        _,
+        _,
+    ) = token_network.functions.getChannelParticipantInfo(channel_identifier, B, A).call()
+    assert B_is_the_closer is False
+    assert B_balance_hash == EMPTY_BALANCE_HASH
+    assert B_nonce == 0
+
     token_network.functions.closeChannel(
         channel_identifier,
         B,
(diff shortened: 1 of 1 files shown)


raiden_contracts/tests/test_channel_close.py
```diff
@@ -181,6 +181,56 @@ def test_close_wrong_sender(
...
+        vals_B.locksroot,
+    )
+
+    token_network.functions.closeChannel(
+        channel_identifier,
+        B,
+        *balance_proof_B,
+    ).transact({'from': A})
+
+    # Even though we somehow provide valid values for the balance proof, they are not taken into
```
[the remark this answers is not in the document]
loredanacirstea: We added a `if (nonce > 0 ) then update balance proof` to avoid wrongly constructed balance proofs to be stored on-chain & lower gas cost. Added a test to check this behavior.

raiden_contracts/tests/test_channel_close.py
```diff
@@ -181,6 +181,56 @@ def test_close_wrong_sender(
...
+        _, _,
+        B_is_the_closer,
+        B_balance_hash,
+        B_nonce,
+        _,
+        _,
+    ) = token_network.functions.getChannelParticipantInfo(channel_identifier, B, A).call()
+    assert B_is_the_closer is False
+    assert B_balance_hash == EMPTY_BALANCE_HASH
+    assert B_nonce == 0
```
palango: You might want to duplicate these checks before the `close` call as well, but not required IMO.

loredanacirstea: Added them.

PR #289, opened 2018-09-26
Add more tests for TokenNetwork unlock

loredanacirstea:
- Replace `EMPTY_MERKLE_ROOT` with `EMPTY_LOCKSROOT` from constants.py (we use this now)
- Add unlock test for different channel, same participants
- Add `test_channel_unlock_no_locked_amount_fail` - unlock fails if there is no amount locked in the `TokenNetwork` contract
- Add more checks for unlock tests


palango:
Reviewing this.

PR #293, opened 2018-10-01
More tests merkle tree

loredanacirstea:
To be merged after https://github.com/raiden-network/raiden-contracts/pull/289

Add tests for `getMerkleRootAndUnlockedAmount`
- check merkle tree components length
- check odd & even number of merkle tree components
- check wrong order of lockhashes - compare resulted merkle roots when the order of lockhashes is changed.

Add test for `getLockDataFromMerkleTree`

### DIFF
--- a/raiden_contracts/tests/test_channel_unlock.py
+++ b/raiden_contracts/tests/test_channel_unlock.py
@@ -14,7 +14,7 @@ from raiden_contracts.utils.utils import (
 )
 from raiden_contracts.constants import TEST_SETTLE_TIMEOUT_MIN
 from raiden_contracts.tests.utils import ChannelValues
-from raiden_contracts.tests.fixtures.config import fake_bytes
+from raiden_contracts.tests.fixtures.config import fake_bytes, TestLockIndex
 from raiden_contracts.tests.fixtures.channel import call_settle
 from raiden_contracts.constants import ParticipantInfoIndex
 from raiden_contracts.tests.fixtures.config import EMPTY_LOCKSROOT
@@ -39,10 +39,10 @@ def test_merkle_root_1_item_unlockable(
     pending_transfers_tree = get_pending_transfers_tree(web3, [6])
 
     secret_registry_contract.functions.registerSecret(
-        pending_transfers_tree.unlockable[0][3],
+        pending_transfers_tree.unlockable[0][TestLockIndex.SECRET],
     ).transact({'from': A})
     assert secret_registry_contract.functions.getSecretRevealBlockHeight(
-        pending_transfers_tree.unlockable[0][2],
+        pending_transfers_tree.unlockable[0][TestLockIndex.SECRETHASH],
     ).call() == web3.eth.blockNumber
 
     (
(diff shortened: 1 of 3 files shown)


raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -57,7 +57,40 @@ def test_merkle_root_1_item_unlockable(
...
+        network_utils.functions.getMerkleRootAndUnlockedAmountPublic(packed[0:-32]).call()
+    # last merkle tree component only contains expiration
+    with pytest.raises(TransactionFailed):
+        network_utils.functions.getMerkleRootAndUnlockedAmountPublic(packed[0:-64]).call()
+
+    assert len(packed) % 96 == 0
+    network_utils.functions.getMerkleRootAndUnlockedAmountPublic(packed).call()
+    network_utils.functions.getMerkleRootAndUnlockedAmountPublic(packed[0:-96]).call()
+
+
```
palango: Seems like something went wrong here

palango: Ah, I should look at the whole diff, not just commits. But might be nice to have the individual commits working.

raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -57,16 +57,66 @@ def test_merkle_root_1_item_unlockable(
...
+        get_accounts,
+        token_network_test_utils,
+        secret_registry_contract,
+):
+    network_utils = token_network_test_utils
+    A = get_accounts(1)[0]
+    pending_transfers_tree = get_pending_transfers_tree(web3, [2, 3, 6], [5])
+
+    secret_registry_contract.functions.registerSecret(
+        pending_transfers_tree.unlockable[0][3],
```
palango: Can you make clearer what the `[0][3]` refers to?

raiden_contracts/tests/test_channel_unlock.py
loredanacirstea: Argh, yes. Something went wrong there. Fixing.

raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -131,6 +131,111 @@ def test_merkle_root_odd_even_components(
...
+        EMPTY_LOCKSROOT,
+    )
+
+    # Merkle tree lockhashes are ordered lexicographicaly.
+    # If we change the order, we change the computed merkle root.
+    # However, the getMerkleRootAndUnlockedAmount orders neighbouring lockhashes
+    # lexicographicaly, so simple item[i], item[i + 1] swap
+    # will still result in the same merkle root.
+    wrong_order = pending_transfers_tree.transfers
+    wrong_order[1], wrong_order[0] = wrong_order[0], wrong_order[1]
```
palango: Should we test this for all pairs? IOW, is this an implementation detail or something that people can/will rely on?

raiden_contracts/tests/test_channel_unlock.py
loredanacirstea: Done, I added a `TestLockIndex` enum to make it explicit.

raiden_contracts/tests/test_channel_unlock.py
loredanacirstea: It is an implementation detail. We order all the locks lexicographically, so we have a convention of how to order them in the merkle tree to get the proper merkle root. I assume this is better than relying on the time when the transfer was initiated / received etc., which would probably end up in race conditions.

I can test all the pairs if you suggest to do so. I thought the current tests were enough, but it's not a big example anyway, so it is doable.

palango: Makes sense, and I'm ok with the tests that are there.

PR #294, opened 2018-10-01
Reenable automated PyPI uploading

ulope:
[skip ci]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,14 +34,14 @@ script:
   - coverage run --source raiden_contracts/ -m py.test -Wd --travis-fold=always -n 2 -v $TEST_TYPE
 
 deploy:
-    provider: pypi
-    skip_cleanup: true
-    user: raiden_pypi_automated
-    password:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -34,14 +34,14 @@ script:
...
-      tags: true
-      repo: raiden-network/raiden-contracts
-      branch: master
+  provider: pypi
+  skip_cleanup: true
+  user: raiden_pypi_automated
+  password:
+    secure: "Lzq5jhpMQukBKLk+cdGS2JjJFwfmhdJT8tF2ZpHwaOLR5EDFtbdmAYu43Tg+GUB0AUaFrdaZ388opZHQjEBNmGbWrJRpbls7yfBjIksNAFQcwTUzk1AhQ1qwUZbeMATy3IE6VVrWka0ovO0Ey3eReOstCVYVrYvaa4ee0uQm3+Vrw8dIrKSgKH8OqaGFC0tH2ycFJlWk2i60fIDqsW+XlLzEVxL8oPzEcW6JtUPgSmkHP3Su/egyfHwckB/NvtS3qG/lPBhzHZKbH6BYyqucatX3e/gZRb0QYnM80nrBzMu7wo/J89WP6ZI6QQwgvW5VumbSbNEBNPqhf1qAuOE1UwgRk6qYuf8EXAhM2gjyWzvCZVDGu0vuBCaj+fscXkIN1wlLZyS1ocGnjyc3jJlG6mgCD8GlQska0uEsgEgrxJbAz3RIEb5GQ2a4jhcsHH5qHrBWGwAGKlYNnIbBHDZoP7cc2vBzHxYwPPizLG81NrvgQVTpaFOZu0+f5tWL3yAHNLiuwDktbj9hg+QtqVEBWE7efWkvw7/FRSMumX3TfFBeAW2UrG5gY2KJNPRmiL4MO1csM6c9gtYiVFAV5ByQjrZfpBmdcxXnItuQbZ731uKWLJZK+cbkDpbsbMAKkrUrpspoz0R+7YHEO6r6/uNgayFVNDQjEBEEg6rqGjrapZI="
+  python: 3.6
+  distributions: sdist bdist_wheel
```
loredanacirstea: Does this only run `python setup.py sdist bdist_wheel` or use `make release` ? 
In https://github.com/raiden-network/raiden-contracts/pull/295 I have:
```
release: clean verify_contracts
	python setup.py sdist bdist_wheel upload
```
Note the `verify_contracts` step. Would this also be included in the automated deployment?

ulope: This is using travis' `dpl` pypi provider which is running [`python setup.py <distributions>`](https://github.com/travis-ci/dpl/blob/2f13fcaff5293d7463660321c0a197fa1e62ef58/lib/dpl/provider/pypi.rb#L124-L127)

So if you want to make sure that the verify step is being run you should hook into the build command (as was done previously). 

But OTOH verify will have been run just before during the tests, so do we really need it to run again?

loredanacirstea: > But OTOH verify will have been run just before during the tests, so do we really need it to run again?

Yes, this seems like overdoing it, so it should be ok not add `verify_contracts` in the build step.

Should we merge this without `[skip ci]` after we are ready for a release? E.g. after merging https://github.com/raiden-network/raiden-contracts/pull/295.

Issue #280, opened 2018-09-23
Installation fails with wrong solidity version installed


schmir:
I'm trying to install raiden-contracts with `pip install raiden-contracts` and get the following error.
```
...
  Running setup.py install for raiden-contracts ... error
    Complete output from command /home/ralf/local/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-install-cpwuj17j/raiden-contracts/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-record-cgvifanb/install-record.txt --single-version-externally-managed --compile --install-headers /home/ralf/local/include/site/python3.6/raiden-contracts:
    running install
    running build
    running build_py
    running compile_contracts
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-cpwuj17j/raiden-contracts/setup.py", line 118, in <module>
        setup(**config)
      File "/home/ralf/local/lib/python3.6/site-packages/setuptools/__init__.py", line 131, in setup
        return distutils.core.setup(**attrs)
      File "/usr/lib/python3.6/distutils/core.py", line 148, in setup
        dist.run_commands()
      File "/usr/lib/python3.6/distutils/dist.py", line 955, in run_commands
        self.run_command(cmd)
      File "/usr/lib/python3.6/distutils/dist.py", line 974, in run_command
        cmd_obj.run()
      File "/home/ralf/local/lib/python3.6/site-packages/setuptools/command/install.py", line 61, in run
        return orig.install.run(self)
      File "/usr/lib/python3.6/distutils/command/install.py", line 589, in run
        self.run_command('build')
      File "/usr/lib/python3.6/distutils/cmd.py", line 313, in run_command
        self.distribution.run_command(command)
      File "/usr/lib/python3.6/distutils/dist.py", line 974, in run_command
        cmd_obj.run()
      File "/usr/lib/python3.6/distutils/command/build.py", line 135, in run
        self.run_command(cmd_name)
      File "/usr/lib/python3.6/distutils/cmd.py", line 313, in run_command
        self.distribution.run_command(command)
      File "/usr/lib/python3.6/distutils/dist.py", line 974, in run_command
        cmd_obj.run()
      File "/tmp/pip-install-cpwuj17j/raiden-contracts/setup.py", line 32, in run
[...]

PR #295, opened 2018-10-02
Contracts transparent compile

### LINKED ISSUE #291: Store contract source hashes
As suggested by @hackaugusto 

1) in `setup.py compile_contracts` perform two actions - compile the json abi+bytecode + store hash of contract source files in the json
2) in `setup.py build` compare hashes from the json abi. Abort if they don't match.

This should prevent bugs like #280 or situations where the compiled abi differs from the updated source.

loredanacirstea:
Fixes #291 #280

Built on top of https://github.com/raiden-network/raiden-contracts/pull/292

- `contracts.json` is now included in `master` and updated on every source code change
- `contracts.json` contains the ABIs, bytecode & source code checksums, in order to verify that the existing compiled file matches the existing source code
- `raiden-contracts` package installation will not compile the contracts, but it will do the above verification
- add a CI contracts verification test
- run the verification test before uploading the package to pypi
- Removed global `CONTRACT_MANAGER`:
  - `raiden-contract` tests should use `ContractManager(CONTRACTS_SOURCE_DIRS)` - the contracts source code; we verify the source code checksums in the compiled contracts file, so we should be safe
  - `raiden` client & tests should use `ContractManager(CONTRACTS_PRECOMPILED_PATH)`, because the precompiled file is the actual output of the package
- added more specific errors: `ContractManagerCompilationError`, `ContractManagerLoadError`, `ContractManagerVerificationError`

Note:
- I did not include the `compile_contracts` step in the `release` process, because I want compilation to be done on demand, to avoid cases where we change the source code by mistake, forget about it and then release it.


raiden_contracts/contract_manager.py
```diff
@@ -142,20 +142,30 @@ def get_event_abi(self, contract_name: str, event_name: str) -> Dict:
         contract_abi = self.get_contract_abi(contract_name)
         return find_matching_event_abi(contract_abi, event_name)
 
-    def _checksum_contracts(self):
-        if self._contracts_source_dirs is None:
-            raise TypeError("Can't checksum when using precompiled contracts.")
-        checksums = []
-        for contracts_dir in self._contracts_source_dirs.values():
+    @staticmethod
+    def checksum_contracts(contracts_source_dirs: list):
```
LefterisJP: Can you add the return type here?

raiden_contracts/contract_manager.py
```diff
@@ -142,20 +142,30 @@ def get_event_abi(self, contract_name: str, event_name: str) -> Dict:
...
-                checksums.append(
-                    '{}:{}'.format(
-                        file.name,
-                        hashlib.sha256(file.read_bytes()).hexdigest(),
-                    ),
+                checksums[file.name] = hashlib.sha256(file.read_bytes()).hexdigest()
+        return checksums
+
+    @staticmethod
+    def verify_contracts(contracts_source_dirs: list, contracts_precompiled: str):
```
LefterisJP: can you add return type `None` here?

setup.py
```diff
@@ -29,22 +28,29 @@ def _get_single_requirement(requirements: List[str], package: str) -> List[str]:
...
         except SystemExit:
             pass
         build_py.run(self)
 
 
-class SdistCommand(sdist):
-    def run(self):
-        from raiden_contracts.contract_manager import CONTRACTS_PRECOMPILED_PATH
+class VerifyContracts(Command):
+    description = 'compile contracts to json'
```
LefterisJP: Perhaps `'Compile contracts and add ABI and checksums to a json file'` ?

raiden_contracts/contract_manager.py
```diff
@@ -142,20 +142,30 @@ def get_event_abi(self, contract_name: str, event_name: str) -> Dict:
...
 
-    def _checksum_contracts(self):
-        if self._contracts_source_dirs is None:
-            raise TypeError("Can't checksum when using precompiled contracts.")
-        checksums = []
-        for contracts_dir in self._contracts_source_dirs.values():
+    @staticmethod
+    def checksum_contracts(contracts_source_dirs: list) -> dict:
+        checksums = {}
+        for contracts_dir in contracts_source_dirs.values():
```
hackaugusto: The type annotation says `contracts_source_dirs` is a list, but the code seems to expect a dict.

raiden_contracts/contract_manager.py
```diff
@@ -142,20 +142,30 @@ def get_event_abi(self, contract_name: str, event_name: str) -> Dict:
...
-                checksums.append(
-                    '{}:{}'.format(
-                        file.name,
-                        hashlib.sha256(file.read_bytes()).hexdigest(),
-                    ),
+                checksums[file.name] = hashlib.sha256(file.read_bytes()).hexdigest()
+        return checksums
+
+    @staticmethod
+    def verify_contracts(contracts_source_dirs: list, contracts_precompiled: str) -> None:
```
hackaugusto: here too the type annotation says `list`, but this will be used by `checksum_contracts` above as a `dict`

raiden_contracts/contract_manager.py
```diff
@@ -142,20 +142,30 @@ def get_event_abi(self, contract_name: str, event_name: str) -> Dict:
...
+        checksummed_sources = ContractManager.checksum_contracts(contracts_source_dirs)
+        contracts_precompiled = ContractManager(contracts_precompiled)._contracts_checksum
```
hackaugusto: ugh, this looks weird, can't you make the checksum public?

raiden_contracts/tests/test_contracts_compilation.py
```diff
@@ -1,17 +1,21 @@
...
+    CONTRACTS_PRECOMPILED_PATH,
 )
 from raiden_contracts.constants import (
     CONTRACT_TOKEN_NETWORK,
     ChannelEvent,
 )
 
 
+def test_verification():
+    ContractManager.verify_contracts(CONTRACTS_SOURCE_DIRS, CONTRACTS_PRECOMPILED_PATH)
```
hackaugusto: This is not testing much. Could you please add test to ensure that it raises when the contract checksum doesn't match? This is the case we are interest in

loredanacirstea: Argh, I tested this in the terminal, but I forgot to write a test. Thank you.

raiden_contracts/contract_manager.py
LefterisJP: by public you mean remove the preceding underscore right?

raiden_contracts/contract_manager.py
```diff
@@ -85,77 +91,99 @@ def _compile_all_contracts(self) -> None:
...
+    def verify_precompiled_checksums(self, contracts_precompiled: str) -> None:
+        """ Compare source code checksums with those from a precompiled file. """
+
+        # We get the precompiled file data
+        contracts_precompiled = ContractManager(contracts_precompiled)
+
+        # Compare each contract source code checksum with the one from the precompiled file
+        for contract, checksum in self.contracts_checksums.items():
+            try:
+                precompiled_checksum = contracts_precompiled.contracts_checksums[contract]
```
LefterisJP: To avoid a regression of what you mentioned in Rocketchat, perhaps add an assertion for the contract names here? They are also kept in the contracts.json file so you can also check for them here, right?

hackaugusto: how would an assertion here avoid that regression?

LefterisJP: Hmm correct. I did not word that correctly.

It would crash, but with a nice message explaining what it was expecting and we could immediately deduce what we did wrong.

hackaugusto: This would not fix the problem. There was no key missing in the dictionary, the problem is that the order of the keys is not stable. The existing code `hashlib.sha256(':'.join(checksums).encode()).hexdigest()` was not stable if the type of checksums is a dictionary.

loredanacirstea: Side note: this made me ask myself if we also check if all the contracts were compiled and checksummed.
We check that they were all checksummed by `overall_checksum`, which would fail if a contract was missing.
Should I also check here if `contracts.json[contracts]` contains the compiled info for all contracts?  E.g. `contracts_precompiled.contracts[contract] is not None` ?

loredanacirstea: Opened https://github.com/raiden-network/raiden-contracts/issues/301 to see if there are other things to be checked for, so you can comment there. I wouldn't expand this PR more..

raiden_contracts/contract_manager.py
```diff
@@ -85,77 +91,99 @@ def _compile_all_contracts(self) -> None:
...
             for file in contracts_dir.glob('*.sol'):
-                checksums.append(
-                    '{}:{}'.format(
-                        file.name,
-                        hashlib.sha256(file.read_bytes()).hexdigest(),
-                    ),
+                checksums[file.name] = hashlib.sha256(file.read_bytes()).hexdigest()
+
+        self.overall_checksum = hashlib.sha256(
+            ':'.join([checksums[key] for key in sorted(checksums)]).encode(),
```
hackaugusto: btw, join accepts an iterable, and `(.. for ... in ...)` is a generator, which is an iterable, so the inner list is unnecessary, the following has the same effect and does one allocation less.

> ':'.join(checksums[key] for key in sorted(checksums)).encode()

Issue #296, opened 2018-10-02
Package and contracts versions with bumpversion


loredanacirstea:
We currently have two config files for `bumpversion`:

- package version: https://github.com/raiden-network/raiden-contracts/blob/de291393bcbc8372a3d3e9b2cbe2bfd4effc6993/.bumpversion.cfg
- smart contracts versions: https://github.com/raiden-network/raiden-contracts/blob/de291393bcbc8372a3d3e9b2cbe2bfd4effc6993/.bumpversion_contracts.cfg

Current status is:
- package: `current_version = 0.4.0`
- smart contracts: `current_version = 0.3.0` (in reality, in the source code this is `0.3._`)

I think both should be in sync on the `major` & `minor` parts.

I would propose:
- `major` as an ABI breaking change; this needs contracts redeployment
- `minor` as a contracts source code change that does not imply an ABI change; this needs contracts redeployment; 
- `patch` any fix that does not touch the contracts source code (tests etc.)

About changes in the exports (e.g. the constants from https://github.com/raiden-network/raiden-contracts/blob/de291393bcbc8372a3d3e9b2cbe2bfd4effc6993/raiden_contracts/constants.py):
- if we decide to change a constant name, how should this be treated? `minor/patch`?
- if we add new exports - `minor/patch`?

If you agree with *both should be in sync on the `major` & `minor` parts.*, then I suggest we trigger the contracts version bump automatically if one of these are used, otherwise we can forget.

loredanacirstea:
@hackaugusto , @LefterisJP , opinions?

loredanacirstea:
> I believe we will need support to multiple versions

But this should not change the old versions, right? At least not the smart contracts code.

So then, what do you propose @hackaugusto ?

PR #297, opened 2018-10-02
Remove MonitoringService and RaidenServiceBundle from bumpversion

loredanacirstea:
They are temporarily removed from the contracts releases, due the development being delayed to another milestone.

### DIFF
--- a/.bumpversion_contracts.cfg
+++ b/.bumpversion_contracts.cfg
@@ -7,14 +7,6 @@ tag = False
 parse = contract_version = "(?P<major>\d+)\.(?P<minor>\d+)\._";
 serialize = contract_version = "{major}.{minor}._";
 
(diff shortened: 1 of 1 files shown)


loredanacirstea:
@LefterisJP , `bumpversion` failed, because the contracts are in another folder. At this point it made more sense for me to just remove this entirely.

LefterisJP:
I don't really have an opinion here. Only thing I see as annoying is that if these contracts fall out of sync due to not being bumped they will need to be brought back in sync manually, before reapplying bumpversion.

loredanacirstea:
Yeah, agree .. then I would leave this to be handled when https://github.com/raiden-network/raiden-contracts/issues/296 is handled. So, closing this for now.

Issue #299, opened 2018-10-02
Add deployed contract addresses


loredanacirstea:
We need to add the addresses of the deployed contracts for all chains (Mainnet + testnets).

We can add them in https://github.com/raiden-network/raiden-contracts/blob/de291393bcbc8372a3d3e9b2cbe2bfd4effc6993/raiden_contracts/constants.py

The Raiden client should use these addresses when running.

LefterisJP:
We already [do](https://github.com/raiden-network/raiden/blob/89190f86cbe9b7cbfa78e10314ba1aafc3e9c663/raiden/constants.py#L35) that in the Raiden client, [categorize them depending on the network](https://github.com/raiden-network/raiden/blob/89190f86cbe9b7cbfa78e10314ba1aafc3e9c663/raiden/constants.py#L80) and also have [an argument](https://github.com/raiden-network/raiden/blob/89190f86cbe9b7cbfa78e10314ba1aafc3e9c663/raiden/ui/cli.py#L408) that's used to switch between the different configurations [when raiden starts](https://github.com/raiden-network/raiden/blob/89190f86cbe9b7cbfa78e10314ba1aafc3e9c663/raiden/ui/cli.py#L707-L776).

I would like to keep it that way. Don't see a reason to have the deployed addresses here.

loredanacirstea:
I consider official deployed addresses as part of the package deliverables. 

Also, aside from the Raiden client, they will also be used by the 3rd party services (e.g. Monitoring Service using the deployed addresses from the MS contract etc.) and by other client implementations or extensions that anyone might build.

LefterisJP:
> I consider the officially deployed addresses as part of the package deliverables.

I have not considered this as part of this package especially since we never had it here so far.

> Also, aside from the Raiden client, they will also be used by the 3rd party services

But this is a valid reason to have them here.

I then agree to moving them here. No need to duplicate code/addresses across packages.

But then the deployment block of the contracts should also be moved. Do you agree we should then move the entire network configuration as seen [here](https://github.com/raiden-network/raiden/blob/89190f86cbe9b7cbfa78e10314ba1aafc3e9c663/raiden/constants.py#L80) in the client?

PR #303, opened 2018-10-04
Move deployed addressed and network configs from client to constants

LefterisJP:
Fix https://github.com/raiden-network/raiden-contracts/issues/299

### DIFF
--- a/raiden_contracts/constants.py
+++ b/raiden_contracts/constants.py
@@ -43,14 +43,13 @@ ROPSTEN = 'ropsten'
 RINKEBY = 'rinkeby'
 KOVAN = 'kovan'
 SMOKETEST = 'smoketest'
-RAIDENTEST_CHAINID = 627
 
 ID_TO_NETWORKNAME = {
-    1: MAINNET,
-    3: ROPSTEN,
-    4: RINKEBY,
-    42: KOVAN,
-    RAIDENTEST_CHAINID: SMOKETEST,
+    ChainId.MAINNET: MAINNET,
+    ChainId.ROPSTEN: ROPSTEN,
+    ChainId.RINKEBY: RINKEBY,
+    ChainId.KOVAN: KOVAN,
+    ChainId.SMOKETEST: SMOKETEST,
 }
 
 
@@ -60,7 +59,7 @@ class NetworkType(Enum):
 
 
 ID_TO_NETWORK_CONFIG = {
-    3: {
+    ChainId.ROPSTEN: {
         NetworkType.TEST: {
             'network_type': NetworkType.TEST,
             'contract_addresses': {
--- a/raiden_contracts/constants.py
+++ b/raiden_contracts/constants.py
@@ -16,18 +16,42 @@ CONTRACT_RAIDEN_SERVICE_BUNDLE = 'RaidenServiceBundle'
 # Deployed contract information
 # Deployed to Ropsten revival on 2018-09-03 from
 # raiden-contracts@fc1c79329a165c738fc55c3505cf801cc79872e4
-ROPSTEN_REGISTRY_ADDRESS = '0xf2a175A52Bd3c815eD7500c765bA19652AB89B30'
-ROPSTEN_DISCOVERY_ADDRESS = '0xEEADDC1667B6EBc7784721B123a6F669B69Eb9bD'
+ROPSTEN_TOKEN_NETWORK_REGISTRY_ADDRESS = '0xf2a175A52Bd3c815eD7500c765bA19652AB89B30'
+ROPSTEN_ENDPOINT_REGISTRY_ADDRESS = '0xEEADDC1667B6EBc7784721B123a6F669B69Eb9bD'
 ROPSTEN_SECRET_REGISTRY_ADDRESS = '0x16a25511A92C5ebfc6C30ad98F754e4c820c6822'
 # Deployed to Ropsten revival on 2018-09-21 from
 # raiden-contracts@bfb24fed3ebda2799e4d11ad1bb5a6de116bd12d
-ROPSTEN_LIMITS_REGISTRY_ADDRESS = '0x6cC27CBF184B4177CD3c5D1a39a875aD07345eEb'
-ROPSTEN_LIMITS_DISCOVERY_ADDRESS = '0xcF47EDF0D951c862ED9825F47075c15BEAf5Db1B'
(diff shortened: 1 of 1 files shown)


raiden_contracts/constants.py
```diff
@@ -11,6 +13,57 @@
 CONTRACT_MONITORING_SERVICE = 'MonitoringService'
 CONTRACT_RAIDEN_SERVICE_BUNDLE = 'RaidenServiceBundle'
 
+# Deployed contract information
+# Deployed to Ropsten revival on 2018-09-03 from
+# raiden-contracts@fc1c79329a165c738fc55c3505cf801cc79872e4
+ROPSTEN_REGISTRY_ADDRESS = '0xf2a175A52Bd3c815eD7500c765bA19652AB89B30'
```
loredanacirstea: Can we make this `ROPSTEN_TOKEN_NETWORK_REGISTRY_ADDRESS`?
Raiden client can rename it internally, but I wanted consistency with the contract names.

raiden_contracts/constants.py
```diff
@@ -11,6 +13,57 @@
...
+ROPSTEN_DISCOVERY_ADDRESS = '0xEEADDC1667B6EBc7784721B123a6F669B69Eb9bD'
```
loredanacirstea: Same as above: `ROPSTEN_ENDPOINT_REGISTRY_ADDRESS`?

raiden_contracts/constants.py
```diff
@@ -11,6 +13,57 @@
...
+ROPSTEN_SECRET_REGISTRY_ADDRESS = '0x16a25511A92C5ebfc6C30ad98F754e4c820c6822'
+# Deployed to Ropsten revival on 2018-09-21 from
+# raiden-contracts@bfb24fed3ebda2799e4d11ad1bb5a6de116bd12d
+ROPSTEN_LIMITS_REGISTRY_ADDRESS = '0x6cC27CBF184B4177CD3c5D1a39a875aD07345eEb'
```
loredanacirstea: `ROPSTEN_LIMITS_TOKEN_NETWORK_REGISTRY_ADDRESS` ?

raiden_contracts/constants.py
```diff
@@ -11,6 +13,57 @@
...
+ROPSTEN_LIMITS_DISCOVERY_ADDRESS = '0xcF47EDF0D951c862ED9825F47075c15BEAf5Db1B'
```
loredanacirstea: `ROPSTEN_LIMITS_ENDPOINT_REGISTRY_ADDRESS` ?

raiden_contracts/constants.py
```diff
@@ -11,6 +13,57 @@
...
+START_QUERY_BLOCK_KEY = 'DefaultStartBlock'
+
+
+class NetworkType(Enum):
+    MAIN = 1
+    TEST = 2
+
+
+ID_TO_NETWORK_CONFIG = {
+    3: {
```
loredanacirstea: Can we have an additional `Enum` with the chain id?
E.g.
```
class ChainId(Enum):
    ROPSTEN = 3
    MAINNET = 1
   ...
```

raiden_contracts/constants.py
LefterisJP: Sure, it's just a bit big :)

raiden_contracts/constants.py
LefterisJP: We are using `Discovery` in the client. The class is named like that too. But yes since we moved it here I will make it endpoint registry.

raiden_contracts/constants.py
LefterisJP: yes

raiden_contracts/constants.py
LefterisJP: Yep!

loredanacirstea: Ah, I was suggesting to have `ChainId. ROPSTEN` instead of `3` here.

raiden_contracts/constants.py
```diff
@@ -11,6 +13,87 @@
 CONTRACT_MONITORING_SERVICE = 'MonitoringService'
 CONTRACT_RAIDEN_SERVICE_BUNDLE = 'RaidenServiceBundle'
 
+# Deployed contract information
+# Deployed to Ropsten revival on 2018-09-03 from
+# raiden-contracts@fc1c79329a165c738fc55c3505cf801cc79872e4
+ROPSTEN_TOKEN_NETWORK_REGISTRY_ADDRESS = '0xf2a175A52Bd3c815eD7500c765bA19652AB89B30'
```
hackaugusto: Why don't we group these contracts in a struct?

```python
class DeployedContract(typing.namedtuple):
    address: str
    deployed_at_block: int

class ContractsBundle(typing.namedtuple):
    version: str
    chain_id: ChainId
    token_network_registry_address: DeployedContract 
    secret_registry: DeployedContract
    endpoint_registry_address: DeployedContract

ROPSTEN = DeployedContracts(
    'Ropsten',
    ChainId.ROPSTEN,
    DeployedContract('0xf2a175A52Bd3c815eD7500c765bA19652AB89B30', 3604000),
    DeployedContract('0x16a25511A92C5ebfc6C30ad98F754e4c820c6822', 3604000),
    DeployedContract('0xEEADDC1667B6EBc7784721B123a6F669B69Eb9bD', 3604000),
)

ROPSTEN_LIMITED = DeployedContracts(
    'Ropsten - Upper limit deposits',
    ChainId.ROPSTEN,
    DeployedContract('0x6cC27CBF184B4177CD3c5D1a39a875aD07345eEb', 3604000),
    DeployedContract('0x8167a262Fa3Be92F05420675c3b409c64Be3d348', 3604000),
    DeployedContract('0xcF47EDF0D951c862ED9825F47075c15BEAf5Db1B', 3604000),
)
```

Then the mapping bellow `ID_TO_NETWORK_CONFIG` is not required, we can reference the smart contracts through the name of structure

hackaugusto: I wonder if the transaction hash would be useful information o have here too.

hackaugusto:
Actually, I think we should move this to a JSON file instead of having it in a python file (not necessarily the one with the precompiled source code). The main argument for this is that the values above ATM must be updated by hand, and that is both burdensome and susceptible to human errors, I'd like to have a tool that deploy the smart contracts and makes the data available in a JSON file. (I'm not proposing to change this PR to this behavior)

raiden_contracts/constants.py
LefterisJP: Yes we should. I missed it during the change. Let me add it now.

Issue #312, opened 2018-10-05
Add post-deployment checks


loredanacirstea:
Add post-deployment checks: test if contracts were deployed with correct arguments, do state checks for the variables (e.g. see if `TokenNetwork` uses the correct `SecretRegistry` etc.)

~~print encoded arguments data for easy contracts verification on etherscan; we can get the ABI, `solc` version & optimizer value from the compiled file.~~ made a new issue with this: https://github.com/raiden-network/raiden-contracts/issues/319

Issue #306, opened 2018-10-05
Automated contracts deployment & changing of deployed addresses


loredanacirstea:
As suggested by @hackaugusto in https://github.com/raiden-network/raiden-contracts/pull/303#issuecomment-427297755:

> Actually, I think we should move this to a JSON file instead of having it in a python file (not necessarily the one with the precompiled source code). The main argument for this is that the values above ATM must be updated by hand, and that is both burdensome and susceptible to human errors, I'd like to have a tool that deploy the smart contracts and makes the data available in a JSON file. (I'm not proposing to change this PR to this behavior)

PR #320, opened 2018-10-10
Store deployment info

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/306 (deployment script should write the deployment info in a json file)
fixes https://github.com/raiden-network/raiden-contracts/issues/312 (post-deployment checks)

Store deployment information for the current contracts version, per chain:
- addresses
- block numbers
- transaction hashes
- constructor parameters
- deployment gas cost

Add post-deployment verification
- check that the `deployment_[CHAIN_NAME].json` file created post deployment contains correct information, compared to the chain
- check that deployed bytecode matches the one from the compiled `contracts.json` file for each contract
- add Readme information to run the verification separate from the deployment: `python -m raiden_contracts.deploy verify --rpc-provider http://127.0.0.1:8545`

Add Rinkeby deployment information. (Kovan & Ropsten will be added in another PR, as there are some more issues to solve)


LefterisJP:
Reviewing this

raiden_contracts/constants.py
```diff
@@ -98,7 +98,7 @@ class MessageTypeId(IntEnum):
 START_QUERY_BLOCK_KEY = 'DefaultStartBlock'
 
 
-class ChainId(Enum):
+class ChainId(IntEnum):
```
LefterisJP: As discussed in rocketchat perhaps let's remove this `IntEnum` and replace it with constants. Semantically this is only the known chain IDs. The user may provide a custom chain ID and then an enum no longer works. This already has bitten us in the client and had to go around it.

raiden_contracts/deploy/__main__.py
```diff
@@ -377,5 +452,152 @@ def register_token_network(
...
+        f' has been updated at {deployment_file_path}.',
+    )
+
+
+def verify_deployed_contracts(web3, contract_manager, deployment_file_path: Path):
+    try:
+        with deployment_file_path.open() as deployment_file:
+            deployment_data = json.load(deployment_file)
+    except (JSONDecodeError, UnicodeDecodeError) as ex:
+        raise DeploymentVerificationError(f'Cannot load deployment data file: {ex}') from ex
```
LefterisJP: should this be `{str(ex)}`?

raiden_contracts/deploy/__main__.py
```diff
@@ -377,5 +452,152 @@ def register_token_network(
...
+    endpoint_registry = web3.eth.contract(
+        abi=endpoint_registry_abi,
+        address=endpoint_registry_address,
+    )
+    endpoint_registry = PrivateContract(endpoint_registry)
+
+    # Check that the deployed bytecode matches the precompiled data
+    blockchain_bytecode = web3.eth.getCode(endpoint_registry_address).hex()
+    compiled_bytecode = contract_manager.contracts[CONTRACT_ENDPOINT_REGISTRY]['bin']
+    # Compiled code contains some additional initial data compare to the blockchain bytecode
```
LefterisJP: compare -> compared

raiden_contracts/deploy/__main__.py
andrevmatos: @LefterisJP `{ex}` == `{ex!s}` == `{str(ex)}`, afaict

raiden_contracts/constants.py
loredanacirstea: I left only `ID_TO_NETWORKNAME` and `NETWORKNAME_TO_ID`. As far as I see, these should be enough.

raiden_contracts/deploy/__main__.py
LefterisJP: @andrevmatos Is this an f-string thing? Or do we always wrap exceptions in `str()` in the client due to the struct logger? Cause in the client we do a lot of:

```
                except RaidenRecoverableError as e:
                    log.error(str(e))
```

If we don't wrap it in `str()` there, then we get the `TypeError: unsupported operand type(s) for +=: 'RaidenRecoverableError' and 'str'`

But perhaps seems like it is a structlogger thing.

raiden_contracts/constants.py
```diff
@@ -97,30 +97,14 @@ class MessageTypeId(IntEnum):
...
-KOVAN = 'kovan'
-SMOKETEST = 'smoketest'
-
 ID_TO_NETWORKNAME = {
-    ChainId.MAINNET: MAINNET,
-    ChainId.ROPSTEN: ROPSTEN,
-    ChainId.RINKEBY: RINKEBY,
-    ChainId.KOVAN: KOVAN,
-    ChainId.SMOKETEST: SMOKETEST,
+    1: 'mainnet',
```
LefterisJP: But can you make them also constants?

e.g.:

NETWORKID_MAINNET = 1
NETWORKID_ROPSTEN = 3
...
```

e.t.c.

raiden_contracts/deploy/__main__.py
loredanacirstea: Not sure about `structlogger`, but: https://www.python.org/dev/peps/pep-0498/#s-r-and-a-are-redundant, also supports lambdas etc.

raiden_contracts/constants.py
LefterisJP: It would allow us in the client to refer to ropsten's (or any other network's) ID programmatically without going through the `ID_TO_NETWORKNAME` whenever we want network specific behavior. It's just a convenience to avoid a mapping lookup.

And constant variables are not redundant in constants.py :)

loredanacirstea: > And constant variables are not redundant in constants.py :)

My problem is defining constants in `raiden-contracts` that are not used in `raiden-contracts`. If they are used **only** in `raiden`, then `raiden` should define them. Otherwise, we can break things by mistake.
To be honest, I would not even use `raiden_contracts.ID_TO_NETWORKNAME` in `raiden`... because of this. I would only stick to the `raiden_contracts` specific deliverables: contracts source, compiled files, `ContractManager` and the specific contracts-related constants.

But if you really want them added, I will.

loredanacirstea: By the way, the entire `ID_TO_NETWORK_CONFIG` will be removed in the PR that solves https://github.com/raiden-network/raiden-contracts/issues/321
We will have `json` files with deployment data. So, any `raiden` specific network configs will be handled by `raiden` anyway

LefterisJP: > My problem is defining constants in raiden-contracts that are not used in raiden-contracts.

Last time we discussed I had understood you wanted contract deployment related data to be in contracts. It used to be in raiden constants.py but then got moved here.

I really have no strong opinion on this. Just finalize these as quick as possible so that the client code has time to adapt. Whatever you remove from here, we will have in the client.

> The contracts will just provide a way to get deployment info for a specific version & chain.

So will the contracts package provide a method for us to read this json file or will we have to implement it in the client?

LefterisJP: >  a method for reading the json file based on version (default=current version) and the chain id, to make it easier for clients. Does this sound ok?

I think it does. These 2 pieces of information is what the client will have. At the moment we also have this `development | release` configuration as additional data. I guess we can incorporate that in the version?

loredanacirstea: > development | release

Not sure what this is about. Are you talking about the pre-limits version and current version of the contracts? For this, I will have 2 versioned `data` dirs with the compiled + deployment info. E.g. `data_0.4.1` for current one, and `data_[PRE_LIMITS_VERSION]` (I'll check what version that was)

Issue #331, opened 2018-10-12
Fix Manifest.in to include specific files or fix deployment test artifact


loredanacirstea:
Deployment test leaves behind a `json` file that is included in the pypi release: https://github.com/raiden-network/raiden-contracts/blob/560b45a82c18dcf891d7d705daeb98ff436bef1e/raiden_contracts/tests/test_deploy_script.py

We should handle this artifact and even better, restrict `MANIFEST.in` to use specific files and not `/data/*`, as introduced in https://github.com/raiden-network/raiden-contracts/commit/7166db83501fc382afdf399f3a1b39e79bf550e0

PR #333, opened 2018-10-15
Fix deployment test & restrict release files

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/331

- added deployment info file save & verification as a separate step from the actual deployment
- used the above verification script in the deployment script test & added some more checks
- this removed the fake deployment info file that was created during the test & could be wrongly included in a package release
- restrict what files can go in the release, to avoid cases where we have private network deployment files created in the `data` directory.
- add option to not store deployment info when running the deploy script

### DIFF
--- a/raiden_contracts/deploy/__main__.py
+++ b/raiden_contracts/deploy/__main__.py
@@ -198,9 +198,16 @@ def raiden(
     gas_limit,
 ):
     setup_ctx(ctx, private_key, rpc_provider, wait, gas_price, gas_limit)
-    deployed_contracts = deploy_raiden_contracts(
-        ctx.obj['deployer'],
-    )
+    deployer = ctx.obj['deployer']
+    deployed_contracts_info = deploy_raiden_contracts(deployer)
+    deployed_contracts = {
+        contract_name: info['address']
+        for contract_name, info in deployed_contracts_info['contracts'].items()
+    }
+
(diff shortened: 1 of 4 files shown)


raiden_contracts/deploy/__main__.py
```diff
@@ -317,85 +339,61 @@ def verify(ctx, rpc_provider, contracts_version):
...
-            },
-            CONTRACT_TOKEN_NETWORK_REGISTRY: {
-                'address': deployed_contracts[CONTRACT_TOKEN_NETWORK_REGISTRY],
-                'transaction_hash': encode_hex(token_network_registry_receipt['transactionHash']),
-                'block_number': token_network_registry_receipt['blockNumber'],
-                'gas_cost': token_network_registry_receipt['gasUsed'],
-                'constructor_arguments': token_network_constructor_arguments,
-            },
-        },
+    deployed_contracts['contracts'][CONTRACT_TOKEN_NETWORK_REGISTRY] = {
```
LefterisJP: The deployment data dict format remains the same here, right?

loredanacirstea: Yes. 
I only changed how it is built, after moving `store_deployment_info` and `verify_deployed_contracts` out off this deployment function.

Issue #16, opened 2018-03-21
Create script to save transaction gas cost estimations to a file


loredanacirstea:
tl;dr
Gas cost estimations are calculated in `test_print_gas.py`. This should not be ran as a test, but should be a script that generates a file with the information, which should be added in the docs.

## Problem Definition

There is no way for (advanced developer) users to easily see a rundown of how much it would cost at any verion of raiden to perform common channel operations.

## Solution

Add a table, probably in [spec.rst](https://github.com/raiden-network/raiden/blob/master/docs/spec.rst) with gas costs of all raiden common channel operations which should be updated every time we change the logic of the contracts or if there is any considerable contract revamp.

This way we will be able to efficiently have a versioned tracking of gas costs optimizations and also a way to identify bottlenecks that require optimization.

loredanacirstea:
Update: the values should not be in the docs, but as package deliverables. We can make this script run at deployment time for the testnet and save the addresses & gas estimation values together, in a json file.

Related: https://github.com/raiden-network/raiden-contracts/issues/306

loredanacirstea:
This should have not been closed, sorry. https://github.com/raiden-network/raiden-contracts/pull/335 only adds the gas limits to the `constants.py` file. We should still have this calculations done automatically, maybe after we compile the `contracts.json` file, but I am moving the issue to anther milestone.

LefterisJP:
An idea mentioned by @hackaugusto [here](https://github.com/raiden-network/raiden-contracts/pull/508#issuecomment-460440862) is to not auto-generate a `constants.py` but instead a `.json` file so that it can be processed by non-python consumers.

pirapira:
* [ ] is there a standard format for storing gas-cost estimations? (Maybe `solc` output)?

pirapira:
`solc` does not like `--combined-json abi,bin,bin-runtime,gas,metadata,ast`.

pirapira:
Here is one option:
* If we switch to `standard-input`, we can get gas estimates.  https://solidity.readthedocs.io/en/v0.5.3/using-the-compiler.html#compiler-input-and-output-json-description
    * but, this will not cover some special needs like "how much gas does it take to unlock() six hashlocks?"

pirapira:
But, the first task is probably https://github.com/raiden-network/raiden-contracts/issues/547 .

After that the test can be turned into a script.

pirapira:
This issue is about adding the gas information in the document.  I don't know if that's so urgent.

PR #335, opened 2018-10-16
Adding the transactions estimated gas limits in constants.py

loredanacirstea:
We can fix https://github.com/raiden-network/raiden-contracts/issues/16 (have a script that generates the gas limit values when compiling the contracts) later.

For now, I just manually added the values to the `constants.py` file, so they can be used in the Raiden client.
Differences from https://github.com/raiden-network/raiden/blob/bc573fc984ce532a85c3ccff96ed733b851f148c/raiden/constants.py are:
- `GAS_REQUIRED_FOR_DISCOVERY_REGISTER = 76000` is now `GAS_REQUIRED_FOR_ENDPOINT_REGISTER = 49183` (not sure where the `76000` value came from)


raiden_contracts/constants.py
```diff
@@ -25,6 +25,17 @@
...
+GAS_REQUIRED_FOR_OPEN_CHANNEL = 112922
+GAS_REQUIRED_FOR_SET_TOTAL_DEPOSIT = 60110
+GAS_REQUIRED_FOR_CLOSE_CHANNEL = 112715
+GAS_REQUIRED_FOR_UPDATE_BALANCE_PROOF = 96348
+GAS_REQUIRED_FOR_SETTLE_CHANNEL = 125052
+GAS_REQUIRED_FOR_UNLOCK_1_LOCKS = 33515
+GAS_REQUIRED_FOR_UNLOCK_6_LOCKS = 72944
+
+GAS_REQUIRED_FOR_REGISTER_SECRET = 46161
+GAS_REQUIRED_FOR_ENDPOINT_REGISTER = 49183
```
LefterisJP: This will require client changes.

hackaugusto:
These constants most likely change with the version of the smart contract, in the future I believe it makes sense to add them in the json file (which depends on the network and contract versions)

LefterisJP:
@hackaugusto good point.

loredanacirstea:
@hackaugusto , agreed. This is why I kept the original issue open https://github.com/raiden-network/raiden-contracts/issues/16 and explained the state there.

PR #339, opened 2018-10-16
Bumpversion contracts 0.4.0

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/296

- bumpversion for contracts from `0.3._` to `0.4.0`
- replaced `{major}.{minor}._` version with `{major}.{minor}.{patch}`
- recompiled contracts after version change & redeployed them + updated `contracts.json` & deployment info files.
- updated version tests
- added some comments on failed `assert`s due to the block numbers not being correctly returned in the transaction receipt for Kovan, to make it easier to manually fix (this should be checked and reported upstream)


raiden_contracts/deploy/__main__.py
```diff
@@ -504,9 +504,15 @@ def verify_deployed_contracts(web3: Web3, contract_manager: ContractManager, dep
     receipt = web3.eth.getTransactionReceipt(
         contracts[CONTRACT_ENDPOINT_REGISTRY]['transaction_hash'],
     )
-    assert receipt['blockNumber'] == contracts[CONTRACT_ENDPOINT_REGISTRY]['block_number']
-    assert receipt['gasUsed'] == contracts[CONTRACT_ENDPOINT_REGISTRY]['gas_cost']
-    assert receipt['contractAddress'] == contracts[CONTRACT_ENDPOINT_REGISTRY]['address']
+    assert receipt['blockNumber'] == contracts[CONTRACT_ENDPOINT_REGISTRY]['block_number'], \
```
LefterisJP: Just an FYI for the future -- no need to change this here: 

In raiden we do multiline asserts like this:

```python
assert a == b , (
    "A very big"
    "string here"
)
```

PR #360, opened 2018-12-04
Update monitoring service to contract changes

karlb:
The monitoring service contracts did not stay in sync with signature
changes in other contracts. This should bring it up to date.


loredanacirstea:
Please also re-enable the MS tests, to make sure things work. E.g. https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/tests/test_monitoring_service.py#L40 (MS tests need some love soon)

palango:
@loredanacirstea Should we do this stuff in a development/Ithaca branch for now or is master ok going forward?

loredanacirstea:
@palango , seems `raiden` is sticking to `master` = development and releases = stable. So, I went for the same approach.
As long as we don't include the service contracts in the released package, I say we can stick with developing them on `master`.

palango:
> seems raiden is sticking to master == development and releases == stable. So, I went for the same approach.

I think Raiden master is still blocked for red eyes

loredanacirstea:
@palango tl;dr if it is easier for you to make an Ithaca branch to start development, feel free to do so. We can realign after Red Eyes.

loredanacirstea:
My comment here https://github.com/raiden-network/raiden-contracts/pull/360#issuecomment-444178113 is wrong. The service contracts can be included in the compiled data (otherwise you cannot use them), and I think we can also deploy them on the test nets. 
The problem is that they are currently in development and we do not want users to think they are reliable. Not deploying them on mainnet should be enough to make this point. A separate Ithaca branch is also fine.

palango:
Let's merge stuff to master then, no need to create more branches.

karlb:
> Please also re-enable the MS tests, to make sure things work. E.g. https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/tests/test_monitoring_service.py#L40 (MS tests need some love soon)

Tests are enabled and pass, now.

PR #361, opened 2018-12-04
Enable service contracts

karlb:
These contracts are required to put the monitoring and path finding services to work. Merge https://github.com/raiden-network/raiden-contracts/pull/360 first, to avoid compilation errors.

Does this change make sense as it is? It does make some MS tests run, but it's hard for me to tell if this is the proper way to enable the contracts.

Can something like this be merged to master, or do we need a separate branch for the foreseeable future?


karlb:
I will include this in https://github.com/raiden-network/raiden-contracts/pull/360, so that the tests actually pass.

PR #372, opened 2018-12-10
Prefix reward hash with "Ethereum Signed Message"

karlb:
`raiden_libs.utils.signing.eth_sign` includes the `"\x19Ethereum Signed
Message"` prefix while `raiden_contracts.utils.sign_utils.sign` doesn’t. This inconsistency causes failures in the MS tests (in the [raiden-monitoring-service](https://github.com/raiden-network/) repo). @pirapira suggested that including the prefix is usually a good idea because some clients can only create signatures including it.

We might want to add these inside the sign and verify functions, but let's make the behaviour consistent first and the refactor, if we really want the prefix in all cases.


pirapira:
I think this choice fits with the current signatures on [balance proofs](https://github.com/raiden-network/spec/blob/bcb3614c4c5e59c9779cd8ca993957627d7db503/smart_contracts.rst#balance-proof), [withdraw proofs](https://github.com/raiden-network/spec/blob/bcb3614c4c5e59c9779cd8ca993957627d7db503/smart_contracts.rst#withdraw-proof) and so on, which are all signed with the prefix `"\x19Ethereum Signed Message"`.

raiden_contracts/utils/sign.py
```diff
@@ -105,12 +105,14 @@ def hash_reward_proof(
...
         nonce):
     return Web3.soliditySha3([
+        'string',
         'uint256',
         'uint256',
         'address',
         'uint256',
         'uint256',
     ], [
+        "\x19Ethereum Signed Message:\n148",
```
loredanacirstea: Would be good to reuse `eth_sign_hash_message`, to make sure the signature format used is the same (e.g. if we change it, we should do that in 1 place).

PR #373, opened 2018-12-10
Add fields to NewBalanceProofReceived event

karlb:
as suggested in
https://github.com/raiden-network/raiden-contracts/pull/360#discussion_r240005564 .


loredanacirstea:
Some notes:
1. About commit messages - referencing the comment here is good. But please also add a small description, as the PRs and issues are kept separately by github and are more difficult to transfer.
@hackaugusto made this point a couple of times and he is right.
2. I see there are a lot of event arguments. It would be nice to document why we need each of them, for multiple reasons:
- they add gas cost, especially if they are `indexed`
- we need to see what we need to actually index (indexing makes event filtering faster)
- we need to see if we have redundant arguments
You can document this directly in the code or start adding these contracts in the `spec` repo (if so, they should be in a different file than the core ones). I would recommend that you start with the spec first, so you have an overview of what needs to be built.
3. Please add a test for the event - see https://github.com/raiden-network/raiden-contracts/blob/ce4c328157eae5b8bc61202e00796301b44f318b/raiden_contracts/tests/test_token_network_registry.py#L182

karlb:
>     2. I see there are a lot of event arguments. It would be nice to document why we need each of them, for multiple reasons.

Yes, good idea. My current focus is to get the MS working again, at all. I've just kept the events and parameters without checking the reason for their existence. I'll do so, but it's not something I'll do right now.

karlb:
>  3. Please add a test for the event - see [raiden-contracts/raiden_contracts/tests/test_token_network_registry.py](https://github.com/raiden-network/raiden-contracts/blob/ce4c328157eae5b8bc61202e00796301b44f318b/raiden_contracts/tests/test_token_network_registry.py#L182)

Done in f5d81a4.

raiden_contracts/constants.py
```diff
@@ -92,6 +92,14 @@ class MessageTypeId(IntEnum):
     COOPERATIVE_SETTLE = 4
 
 
+# Message types used my MonitoringService contract
```
LefterisJP: /s/my/by

PR #386, opened 2018-12-17
Document getUnlockIdentifier()

pirapira:
because when I saw this function first, I couldn't tell the roles of the two address arguments of this function.


raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -1086,6 +1086,16 @@ contract TokenNetwork is Utils {
         }
     }
 
+    /// @dev Get the hash of the channel identifier and the participant
+    /// addresses (whose orderings matter). The hash might be useful for
```
loredanacirstea: `matter` -> `matters`

raiden_contracts/contracts/TokenNetwork.sol
```diff
@@ -1086,6 +1086,16 @@ contract TokenNetwork is Utils {
...
+    /// looking up an UnlockData in
```
loredanacirstea: `looking up an UnlockData in` -> `looking up the appropriate UnlockData for the participant`

raiden_contracts/contracts/TokenNetwork.sol
pirapira: will change into `whose ordering matters`.

raiden_contracts/contracts/TokenNetwork.sol
pirapira: will change into `... useful for the partner to look up the appropriate UnlockData to claim.`

Issue #357, opened 2018-11-26
Add docstrings to tests


loredanacirstea:
Docstrings should be used for most tests. We should have a human readable explanation of what the test does.
I have mostly tried to give relevant names to the test itself, but this is obviously not enough.

pirapira:
I'll first do this on the unit tests.

pirapira:
I completely missed `tests/test_channel_update_transfer.py`.

Of the skipped tests, I would only document the Ithaka functionalities in this issue.

I guess `test_print_gas.py` should be explained in the release procedure.  Maybe there should be a docstring with a link to that document.

pirapira:
So I think this one is done for Ithaka.

PR #387, opened 2018-12-20
Docstring tests

pirapira:
This is a part of #357


raiden_contracts/tests/test_channel_close.py
```diff
@@ -439,6 +454,7 @@ def test_close_channel_event_no_offchain_transfers(
         create_channel,
         event_handler,
 ):
+    """ closeChannel succeeds and emits an event even with no balance proofs """
```
loredanacirstea: I would suggest:
`closeChannel succeeds when called no balance proof and emits an event even with nonce 0`

raiden_contracts/tests/test_channel_deposit.py
```diff
@@ -131,6 +144,7 @@ def test_deposit_notapproved(
         create_channel,
         get_accounts,
 ):
+    """ Calling setTotalDeposit() without approving transfers on the token contract """
```
loredanacirstea: `Calling setTotalDeposit() fails ...`

raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -440,6 +453,10 @@ def test_channel_unlock_smaller_locked_amount(
         get_accounts,
         reveal_secrets,
 ):
+    """ Test an unlock() call that claims too many tokens
+
+    When settleChannel() call computes a smaller amount of locked tokens than
+    the following unlock() call,  """
```
loredanacirstea: This docstring is not complete.
```
When settleChannel() call computes a smaller amount of locked tokens than
    the following unlock() call, the settleChannel() computation is stronger and
    the participant receives less tokens. Stealing tokens from other channels is then prevented.
```

raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -598,6 +617,7 @@ def test_channel_unlock(
         reveal_secrets,
         event_handler,
 ):
+    """ unlock() on pending transfers should split the locked amount into both parties """
```
loredanacirstea: `unlock() on pending transfers with unlockable and expired locks should split the locked amount accordingly, to both parties`

PR #392, opened 2019-01-03
Document tests about TokenNetworkRegistry contract

pirapira:
Moreover, added some useful cases to try invalid constructor arguments
containing wrong challenge period durations.


pirapira:
This PR is part of #357.

PR #395, opened 2019-01-03
Add comments to test_token.py

pirapira:
This is a part of #357.

### DIFF
--- a/raiden_contracts/tests/test_token.py
+++ b/raiden_contracts/tests/test_token.py
@@ -3,6 +3,8 @@ from eth_tester.exceptions import TransactionFailed
 
 
 def test_token_mint(web3, custom_token, get_accounts):
+    """ Use the mint() function of the custom token contract """
+
     (A, B) = get_accounts(2)
     token = custom_token
     multiplier = custom_token.functions.multiplier().call()
@@ -25,6 +27,8 @@ def test_token_mint(web3, custom_token, get_accounts):
 
 
 def test_approve_transfer(web3, custom_token, get_accounts):
+    """ Use the approve() function of the custom token contract """
+
     (A, B) = get_accounts(2)
     token = custom_token
     token.functions.mint(50).transact({'from': A})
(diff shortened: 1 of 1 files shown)


raiden_contracts/tests/test_token.py
```diff
@@ -43,6 +47,8 @@ def test_approve_transfer(web3, custom_token, get_accounts):
 
 
 def test_token_transfer_funds(web3, custom_token, get_accounts, txn_gas):
+    """ transferFuhnds() should fail when the ETH balance of the contract is zero """
```
loredanacirstea: typo `transferFunds`

LefterisJP: /s/Fuhnds/Funds

raiden_contracts/tests/test_token.py
```diff
@@ -60,8 +67,12 @@ def test_token_transfer_funds(web3, custom_token, get_accounts, txn_gas):
 
 
 def test_custom_token(custom_token, web3):
-    assert web3.eth.getCode(custom_token.address)
+    """ See custom_token.address contains some big code """
+
+    assert len(web3.eth.getCode(custom_token.address)) > 100
```
LefterisJP: Why care about the size of the code?

pirapira: In case of empty, '0x' is returned.  `assert '0x'` silently succeeds.

pirapira:
Waiting for CI.

raiden_contracts/tests/test_token.py
LefterisJP: Ah good point @pirapira

LefterisJP: In that case you should also change the docstring though -- as anyone who reads the code will have the same question I had.

hackaugusto: shouldn't it decode the hex representation and check the length is larger then 0?

pirapira: Maybe we also have the exact expected hex string.

PR #405, opened 2019-01-07
Check the exact runtime code in test_token

pirapira:
Also factor out a utility function for computing the runtime code.

This commit comes from this conversation
https://github.com/raiden-network/raiden-contracts/pull/395#discussion_r245040959


pirapira:
I have a sense of getting lost here.  Thinking.

pirapira:
@loredanacirstea  #408 was not because of the calculation here, but because of #414 .

pirapira:
@loredanacirstea merge?

pirapira:
There is a better way to compute the expected onchain code: #419 .

Issue #408, opened 2019-01-08
Fix runtime bytecode check when deploying a contract


loredanacirstea:
From: https://github.com/raiden-network/raiden-contracts/issues/397
The runtime bytecode calculation is wrong here and **might** fail for other systems on which `solc` in stalled:
https://github.com/raiden-network/raiden-contracts/blob/343b9a9255eaad977a4988ac92b50fb174dd0ced/raiden_contracts/deploy/__main__.py#L519-L525

Needs to be fixed for all subsequent occurrences.

@pirapira started working on this in https://github.com/raiden-network/raiden-contracts/pull/405.

pirapira:
@loredanacirstea  Do you mean
a. we should use `solc --runtime-bin` to get the runtime code, or
b. there is a mistake in those few lines that should be changed?

pirapira:
Maybe I can see this problem with `solc 0.5.*`.

loredanacirstea:
@pirapira I suspect there might be a mistake in `compiled_bytecode = compiled_bytecode[-len(blockchain_bytecode):]` if this verification fails for some systems.
We need a clear understanding of how the `compiled_bytecode` is produced and what it contains (length, meaning).

Otherwise, determine the external failure reason. In the initial complaint (https://github.com/raiden-network/raiden-contracts/issues/397), there was also another bug that recompiled the contracts when deploying, fixed by https://github.com/raiden-network/raiden-contracts/pull/407.
So, the above bytecode `assert` was between that newly recompiled data and the locally deployed contract.

pirapira:
Now I can reproduce the issue.
```
  File "/home/yh/src/raiden-contracts/raiden_contracts/deploy/__main__.py", line 520, in verify_deployed_contracts
    assert blockchain_bytecode == compiled_bytecode, f"different\n{blockchain_bytecode}\n{compiled_bytecode}"
AssertionError: different
[...]

PR #414, opened 2019-01-08
Do not store '/home/yh/...' in contracts.json

pirapira:
Before this commit, `contracts.json` contained some absolute paths with user names
and so on. This caused verification failures after a user compiles a
contract locally. The absolute paths were used in the contract metadata
and were included in the metadata hash at the end of the deployed code.

This was one cause of
https://github.com/raiden-network/raiden-contracts/issues/408


loredanacirstea:
Just a note, I would hold off merging this just because it changes the precompiled data and this will not match the deployed contracts. Also, this fix is not urgent. We might do a python-code fix release first (finish with removing `raiden-libs`), as this might be useful for the Raiden client.

pirapira:
See https://github.com/raiden-network/raiden-contracts/commit/f16c4e4359509ae86ee601b105cdf36f50dbcb72#r31897816 for the problem this caused, and https://github.com/raiden-network/raiden-contracts/pull/429 for the reversion.

Issue #389, opened 2019-01-02
Deployment script - cannot import name 'get_settlement_amounts'


loredanacirstea:
```
Meet an error : Traceback (most recent call last):
File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
"main", mod_spec)
File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
exec(code, run_globals)
File "/home/a/sym/raiden-contract-git/raiden-contracts/raiden_contracts/deploy/main.py", line 34, in <module>
from raiden_contracts.tests.utils.transaction import check_succesful_tx
File "/home/a/sym/raiden-contract-git/raiden-contracts/raiden_contracts/tests/utils/init.py", line 5, in <module>
from .contracts import 
File "/home/a/sym/raiden-contract-git/raiden-contracts/raiden_contracts/tests/utils/contracts.py", line 6, in <module>
from raiden_contracts.tests.fixtures.base.web3 import FAUCET_ALLOWANCE
File "/home/a/sym/raiden-contract-git/raiden-contracts/raiden_contracts/tests/fixtures/init.py", line 11, in <module>
from .channel import 
File "/home/a/sym/raiden-contract-git/raiden-contracts/raiden_contracts/tests/fixtures/channel.py", line 14, in <module>
from raiden_contracts.tests.utils import (
ImportError: cannot import name 'get_settlement_amounts'
```
This might be an unintended effect of the PR that removed part of the `raiden-libs` dependencies

pirapira:
How do I see this error message?

loredanacirstea:
@pirapira this was an issue posted in gitter, that I did not have time to check if this indeed happens: https://gitter.im/raiden-network/raiden?at=5c263073f6166a302781eeb2. Unfortunately, the op did not open an issue or respond afterwards.

pirapira:
Now I can reproduce this issue.  I was trying to fix #380 and saw
```
(env) ~/src/raiden-contracts((45181b0...)|BISECTING) $ python raiden_contracts/tests/deprecation_switch_testnet.py
```

causes this circular import.

pirapira:
My circular import is like this:
```
(env) ~/src/raiden-contracts(389) $ python raiden_contracts/tests/deprecation_switch_testnet.py 
Traceback (most recent call last):
  File "raiden_contracts/tests/deprecation_switch_testnet.py", line 7, in <module>
    from raiden_contracts.deploy.__main__ import (
[...]

PR #415, opened 2019-01-08
Fix circular dependency #389

pirapira:
This PR
* adds a test to observe the circular dependency #389 in the Travis runs, and
* fixes the circular dependency


pirapira:
#417 does it in a better way.

PR #422, opened 2019-01-09
Add github highlighting for .sol

konradkonrad:
This adds nice syntax highlighting for solidity (`*.sol`) to the github web interface :)
See: ethereum/solidity#2681

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sol linguist-language=Solidity


konradkonrad:
Or doesn't it? I am confused now, https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/contracts/EndpointRegistry.sol already has highlighting, while https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/contracts/TokenNetwork.sol has not.

loredanacirstea:
I thought maybe the size of the contract was an issue, but `TokenNetworkRegistry.sol` & `Utils.sol` also don't have highlighting.

LefterisJP:
Works. All of them are now highlighted for me. That's funny btw. In the past this was a requirement as github did not automatically detect solidity. Guess the automatic detection is not perfect yet.

PR #425, opened 2019-01-10
Versionup

pirapira:
I'm following https://github.com/raiden-network/raiden-contracts/blob/master/RELEASE.rst to make a release.


loredanacirstea:
@pirapira the contracts version should not be updated if no changes are made to the Solidity source code or precompiled data. If the release guide is not clear on that, please suggest a fix.

pirapira:
There are new comments in the Solidity sources.

loredanacirstea:
Ah, I forgot about them. So we are changing the precompiled data anyway. Thinking.

loredanacirstea:
@pirapira I would suggest we delay the release until the team has a discussion about how we move forward - either support yet another contracts version (current development version, 0.4.0, 0.3._) or remove deployed data (which isn't too helpful for the fixes we added to the deploy script).

What is your opinion on this?

pirapira:
Let's decide on that first.

Issue #424, opened 2019-01-09
Make sure the development pre_limits version can be deployed on a private chain


loredanacirstea:
Currently, the development `pre_limits` version cannot be deployed on a private chain. We need to add support for this until https://github.com/raiden-network/raiden-contracts/issues/391 is ready.

PR #426, opened 2019-01-10
Add support for deploying the testnet pre_limits version

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/424


raiden_contracts/contract_manager.py
```diff
@@ -202,6 +202,10 @@ def contracts_data_path(version: Optional[str] = None):
     if version is None or version == CONTRACTS_VERSION:
         return _BASE.joinpath('data')
     else:
+        # Temporary fix for the pre_limits version
+        # to avoid breaking changes
+        version = 'pre_limits' if version == '0.3._' else version
```
pirapira: Can you also describe the non-temporary version of the fix?

raiden_contracts/deploy/__main__.py
```diff
@@ -73,10 +74,12 @@ def __init__(
         self.precompiled_path = contracts_precompiled_path(self.contracts_version)
         self.contract_manager = ContractManager(self.precompiled_path)
 
-        # Check that the precompiled data is correct
-        contract_manager_source = ContractManager(contracts_source_path())
-        contract_manager_source.checksum_contracts()
-        contract_manager_source.verify_precompiled_checksums(self.precompiled_path)
+        # Check that the precompiled data matches the source code
+        # Only for current version, because this is the only one with source code
```
pirapira: Why do we have old precompiled data (without sources) around?

loredanacirstea:
In light of https://github.com/raiden-network/raiden-contracts/pull/425#issuecomment-453055778, I renamed `data_pre_limits` to `data_0.3._`, to be consistent with versions + added a Readme note.

From commit comment:
This makes the support for multiple versions more consistent, by only using the contracts version and not an arbitrary version descriptor.
This is a breaking change in API for retrieving the version, but the next release will nonetheless contain other breaking changes.

In order to use this version:
- the deploy script to use `-- contracts-version "0.3._"`
- `contracts_precompiled_path("0.3._")`

raiden_contracts/deploy/__main__.py
pirapira: I guess this question is is out of scope of this PR.

raiden_contracts/deploy/__main__.py
```diff
@@ -73,10 +74,12 @@ def __init__(
...
+        if self.contracts_version in [None, CONTRACTS_VERSION]:
+            contract_manager_source = ContractManager(contracts_source_path())
+            contract_manager_source.checksum_contracts()
+            contract_manager_source.verify_precompiled_checksums(self.precompiled_path)
```
pirapira: Maybe
```
else:
    warn('Skipped checks against the source code becaust it's not available.')
```

raiden_contracts/deploy/__main__.py
loredanacirstea: Because the old precompiled data was added on the fly and it was considered to be a temporary addition just for the Raiden client tests with no limits.

We need to solve this for following versions.

raiden_contracts/deploy/__main__.py
loredanacirstea: done.

Issue #430, opened 2019-01-10
Do not store absolute paths in contracts.json


pirapira:
Currently, in `contracts.json` we see absolute paths like `/home/yh/...` but we we should not.

#414 was one attempt to fix this, but in a wrong way.  #414 used the working directory of the process, and that caused https://github.com/raiden-network/raiden-contracts/commit/f16c4e4359509ae86ee601b105cdf36f50dbcb72#r31897816.

pirapira:
The correct fix would involve `cd` ing to the top of the package before calling `solc` with relative paths.

pirapira:
This will solve https://github.com/raiden-network/raiden-contracts/issues/408 .

PR #434, opened 2019-01-10
Use relative paths when calling Solidity

pirapira:
`contracts.json` used to contain absolute paths because `solc` was called
with absolute paths. After this commit, `contract_manager.py` changes the
current directory to the top of raiden_contracts package and calls solc
with relative paths (it even comes back to the previous current
directory).

This fixes https://github.com/raiden-network/raiden-contracts/issues/430 .

This time it won't break the monitoring-service Travis builds  :crossed_fingers:


pirapira:
https://travis-ci.org/palango/raiden-monitoring-service/builds/477922793 goes through.

PR #435, opened 2019-01-10
Use bin-runtime in compilation and post-deploy bytecode check

loredanacirstea:
fixes https://github.com/raiden-network/raiden-contracts/issues/419

Adds `bin-runtime` computed by `solc` to the precompiled data.
Uses `bin-runtime` from the precompiled data to check the post-deployment bytecode.

PR #439, opened 2019-01-11
Add bin-runtime to pre-limits 0.3._ version

loredanacirstea:
Source code: https://github.com/raiden-network/raiden-contracts/tree/fc1c79329a165c738fc55c3505cf801cc79872e4

We added `bin-runtime` to the precompiled data in https://github.com/raiden-network/raiden-contracts/pull/435. This PR updates the precompiled data for `0.3._`


LefterisJP:
Looks good as far as the README is concerned. The `bin-runtime` is only checked by the deployment script, right after it was deployed, right? So this needs no changes on the client side?

loredanacirstea:
@LefterisJP 

> The bin-runtime is only checked by the deployment script, right after it was deployed, right? So this needs no changes on the client side?

Correct.

PR #441, opened 2019-01-11
Upgrade solc from ^0.4.23 to ^0.5.2 and implement version changes

loredanacirstea:
Upgrade from `^0.4.23` to `0.5.2`.

- specify data location for function arguments `memory` / `storage` / `calldata`
- update MS/RSB contracts to `0.5.2`
- use `payable` for addresses that are transferred ETH
- test tokens - replace `constant`, fix abi encoding for `call()`
- Remove empty remapping prefix from compilation. Not sure why it was added, but it is not allowed in solc `0.5.2`.

Reasons:
- we might use `struct`s in external functions (available `^0.5.0`)
- generally good to go with the new solc stable minor version considering all the changes they did & the fact that we have some code to refactor + additional contracts to write
- there is a compiler bug for `0.4.25`: `ExpExponentCleanup` (https://etherscan.io/solcbuginfo), which does not apply to the current code, but maybe will be used.
- `0.5.2` release is already 1 month old. Considering that we are in development phase, I think this is fine.


raiden_contracts/contract_manager.py
```diff
@@ -76,8 +76,10 @@ def _compile_all_contracts(self) -> None:
 
         def relativise(path):
             return path.relative_to(_BASE)
-        import_dir_map = ['=.'] + ['%s=%s' % (k, relativise(v))
-                                   for k, v in self.contracts_source_dirs.items()]
+        import_dir_map = [
+            '%s=%s' % (k, relativise(v))
+            for k, v in self.contracts_source_dirs.items()
+        ]
```
loredanacirstea: Explanation:
From https://solidity.readthedocs.io/en/v0.5.2/using-the-compiler.html?highlight=remappings#using-the-commandline-compiler:

> An empty remapping prefix is not allowed.

Not sure why `['=.']` was added in the first place, it is enough to compute `['raiden=contracts', 'test=contracts/test', 'services=contracts/services']`

Issue #447, opened 2019-01-18
User Deposit Contract


karlb:
The Raiden services will ask for payment in RDN. The Monitoring Service and potentially the Pathfinding Service require deposits to be made in advance of service usage. These deposits will be handled by the User Deposit Contract.
Usage of the deposit for payments is not safe from double spending, but measures can be taken to reduce the likelihood to acceptable levels. This is a good trade off as long as the money lost on double spending is less than the savings in gas cost. 

## Requirements
- [x] Users can deposit and withdraw tokens.
- [x] Tokens can be deposited to the benefit of other users. This could facilitate onboarding of new raiden users and allow a MS to defer the monitoring to another MS.
- [x] Tokens can't be withdrawn immediately, but only after a certain delay. This allows services to claim their deserved payments before the withdraw takes place.
- [x] Services can read the effective balance of a user (current balance - planned withdrawals)
- [x] Service contracts are trusted and can claim tokens for the service providers.
- [x] Services can listen to events which notify them of decreasing user balances. A service can then claim payments before double spending becomes too likely.

### Monitoring Service rewards

The MS is promised a reward for each settlement in which it took part on behalf of the non-closing participant. Before accepting a monitor request, the MS checks if enough tokens are deposited in the UDC. The MS that has submit the latest BP upon settlement will receive the promised tokens on it's UDC balance.

### 1-n payments
[This is just a suggestions so far. No decision made.]
The PFS will be paid with signed IOUs, roughly a simplified uRaiden adapted to 1-n payments. The IOU contains the amount of tokens that can be claimed from the signer's UDC balance. See https://github.com/raiden-network/raiden-services/issues/38 for details.

palango:
Nice overview, no requirements I'd add. Some comments below.

> Users can deposit and withdraw tokens.

The token will always be the RDN token, right?

> Service contracts have permission to transfer tokens in the UDC between users.

What do you mean by that? I'd also add that the service contracts are trusted and can claim tokens for the providers.

> Before accepting a monitor request, the MS checks if enough tokens are deposited in the UDC

The MS might also want to check the deposit before sending the update transaction.

karlb:
Specification from this ticket is added to the spec repo in https://github.com/raiden-network/spec/pull/202.
Implementation in https://github.com/raiden-network/raiden-contracts/pull/448.

PR #448, opened 2019-01-18
First UDC implementation suggestion

karlb:
No handling of trusted contracts, no testing of error cases. Should be
enough to see if this is a workable approach.

Implements #447


raiden_contracts/contracts/services/UserDeposit.sol
```diff
@@ -0,0 +1,142 @@
+pragma solidity ^0.5.2;
+
+import "raiden/Token.sol";
+import "raiden/Utils.sol";
+
+contract UserDepositContract is Utils {
+    string constant public contract_version = "0.1.0";
```
loredanacirstea: Note that we currently go with the same version for all contracts and we also need to write a test for this - e.g. https://github.com/raiden-network/raiden-contracts/blob/d08a3abaffdd9ebd4b6cd0a4209d34a1a1328715/raiden_contracts/tests/test_token_network_registry.py#L17-L20 (can be done in another PR, an issue can be opened).
Also, the overall contracts version is written in the precompiled + deployment data only once per all contracts.

Having multiple contract version within the same package can complicate things -> we went with redeployment of all contracts when changing the overall `CONTRACTS_VERSION`. You may want individual versions at some point, but I think we should be consistent until the decision to change this is reached.

raiden_contracts/contracts/services/UserDeposit.sol
```diff
@@ -0,0 +1,142 @@
...
+    // Keep track of balances
+    mapping(address => uint256) public balances;
+    mapping(address => WithdrawPlan) public withdraw_plans;
+
+    /*
+     *  Structs
+     */
+    struct WithdrawPlan {
+        uint256 amount;
+        uint256 earliest_withdraw;  // earliest block at which withdraw is allowed
```
loredanacirstea: Consider something like `withdraw_block` - to be more specific that is a `block` and not a `timestamp`, for example.

raiden_contracts/contracts/services/UserDeposit.sol
```diff
@@ -0,0 +1,142 @@
...
+
+    /*
+     *  Constructor
+     */
+
+    /// @notice Set the default values for the smart contract
+    /// @param _token_address The address of the token to use for rewards
+    constructor(
+        address _token_address
+    )
```
loredanacirstea: Generally, we keep it in 1 line if < 80/99 characters.

raiden_contracts/contracts/services/UserDeposit.sol
```diff
@@ -0,0 +1,142 @@
...
+        uint256 added_deposit = total_deposit - balances[beneficiary];
+
+        balances[beneficiary] += added_deposit;
+        require(token.transferFrom(msg.sender, address(this), added_deposit));
+    }
+
+    /// @notice Internally transfer deposits between two addresses
+    /// The amount will be deducted from the msg sender's balance
+    /// @param receiver Account to which the amount will be credited
+    /// @param amount Amount of tokens to be withdrawn
```
loredanacirstea: Good to also document the `@return`.

raiden_contracts/contracts/services/UserDeposit.sol
```diff
@@ -0,0 +1,142 @@
...
+    /// @param amount Maximum amount of tokens to be withdrawn
+    function planWithdraw(uint256 amount) public {
+        require(amount > 0);
+        require(balances[msg.sender] >= amount);
+
+        withdraw_plans[msg.sender] = WithdrawPlan({
+            amount: amount,
+            earliest_withdraw: block.number + withdraw_delay
+        });
+        emit WithdrawPlanned(msg.sender, balances[msg.sender] - amount);
```
loredanacirstea: Just a note for `planWithdraw` - this function enables users to overwrite their previous planned withdraws and emits an event each time this is overwritten. I assume this is the intention.

raiden_contracts/tests/test_user_deposit_contract.py
```diff
@@ -0,0 +1,79 @@
+import pytest
+
+from raiden_contracts.constants import CONTRACT_USER_DEPOSIT
+
+
+@pytest.fixture
+def user_deposit_contract(
```
loredanacirstea: We place fixtures in `tests/fixtures`, as they are generally used across multiple test files.

raiden_contracts/tests/test_user_deposit_contract.py
```diff
@@ -0,0 +1,79 @@
...
+    custom_token.functions.mint(30).transact({'from': A})
+    custom_token.functions.approve(user_deposit_contract.address, 30).transact({'from': A})
+
+    # deposit to A's own balance
+    user_deposit_contract.functions.deposit(A, 10).transact({'from': A})
+    assert user_deposit_contract.functions.balances(A).call() == 10
+
+    # increase A's deposit
+    user_deposit_contract.functions.deposit(A, 20).transact({'from': A})
+    assert user_deposit_contract.functions.balances(A).call() == 20
```
loredanacirstea: Another check can be added here to ensure that `deposit(A, 15)` fails.

raiden_contracts/contracts/services/UserDeposit.sol
```diff
@@ -0,0 +1,142 @@
...
+    function deposit(address beneficiary, uint256 total_deposit) public
+    {
+        require(total_deposit > balances[beneficiary]);
+
+        // Calculate the actual amount of tokens that will be transferred
+        uint256 added_deposit = total_deposit - balances[beneficiary];
+
+        balances[beneficiary] += added_deposit;
+        require(token.transferFrom(msg.sender, address(this), added_deposit));
+    }
```
loredanacirstea: Note - some more invariant checks will be needed, but I understand that the current version is in work and they will probably be added later.

pirapira:
@karlb ping me when this PR is ready.

raiden_contracts/contracts/services/UserDeposit.sol
karlb: And I think we can even skip the safe failsafe_addition. We know that the total token supply is a uint256 (ERC20 uses `uint` to return `totalSupply`), so we can't get large enough numbers to overflow.

loredanacirstea: Looked over this now and I don't think it needs other checks.
Regarding `balances[beneficiary] += added_deposit;`, even the overflow here is protected against by `require(total_deposit > balances[beneficiary]);` -> edge case would be `total_deposit == 2^256 - 1` -> after `balances[beneficiary] += added_deposit;`, we have `balances[beneficiary] == 2^256 - 1`, so it is fine, as the rest of the checks will be done in `token.transferFrom`

> Do we have to expect malicious behavior of the token contract or will they be checked and approved in some way?

In general in Raiden, we assume malicious behavior up to a point (meaning, we always have state checks before the transfer, as you also have here). However, in this case, the deployer of UDC will choose the token & make sure the implementation is not malicious.

karlb:
@pirapira, I'm ready for more feedback!

raiden_contracts/contracts/services/UserDeposit.sol
```diff
@@ -0,0 +1,171 @@
...
+    /// @return true if transfer has been done successfully, otherwise false
+    function transfer(
+        address sender,
+        address receiver,
+        uint256 amount
+    )
+        canTransfer()
+        external
+        returns (bool success)
+    {
```
pirapira: Have you considered whether to emit an event when `sender == receiver`?

raiden_contracts/contracts/services/UserDeposit.sol
```diff
@@ -0,0 +1,171 @@
...
+    /// @param total_deposit The amount tokens in the user's deposit after the call
+    function deposit(address beneficiary, uint256 total_deposit)
+        external
+    {
+        require(total_deposit > balances[beneficiary]);
+
+        // Calculate the actual amount of tokens that will be transferred
+        uint256 added_deposit = total_deposit - balances[beneficiary];
+
+        require(token.transferFrom(msg.sender, address(this), added_deposit));
```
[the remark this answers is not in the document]
loredanacirstea: Hmm, with my comment from https://github.com/raiden-network/raiden-contracts/pull/448/files#r250609427 `(meaning, we always have state checks before the transfer, as you also have here)` I actually meant `state changes before the transfer` - so, exactly what @pirapira is saying above. Did I say something that prompted you to change the order?

raiden_contracts/tests/test_user_deposit_contract.py
```diff
@@ -0,0 +1,119 @@
...
+    get_accounts,
+):
+    (A, B) = get_accounts(2)
+    custom_token.functions.mint(30).transact({'from': A})
+    custom_token.functions.approve(user_deposit_contract.address, 30).transact({'from': A})
+
+    # deposit to A's own balance
+    user_deposit_contract.functions.deposit(A, 10).transact({'from': A})
+    assert user_deposit_contract.functions.balances(A).call() == 10
+    assert custom_token.functions.balanceOf(A).call() == 20
```
pirapira: Also check `custom_token.functions.balanceOf(user_deposit_contract.address)`.

raiden_contracts/tests/test_user_deposit_contract.py
```diff
@@ -0,0 +1,119 @@
...
+        user_deposit_contract.functions.deposit(A, 19).transact({'from': A})
+
+    # A deposits to the benefit of B
+    user_deposit_contract.functions.deposit(B, 10).transact({'from': A})
+    assert user_deposit_contract.functions.balances(B).call() == 10
+    assert custom_token.functions.balanceOf(A).call() == 0
+
+    # can't deposit if not enough tokens are approved
+    with pytest.raises(TransactionFailed):
+        user_deposit_contract.functions.deposit(A, 21).transact({'from': A})
```
pirapira: Not sure if this is failing for lack of balance or insufficient approval.

raiden_contracts/contracts/services/UserDeposit.sol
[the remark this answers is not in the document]
karlb: While changing that I was thinking about first executing the negative effects for the user so that he didn't gain any balance when weird things happen inside of `transferFrom`. But this is not helpful because everything is done in a transactional context anyway and it leads to the problem you both pointed to. So my temporary confusion is cleared up now and I will fix this, again.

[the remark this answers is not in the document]
pirapira: Yeah this environment is a bit different from, say, OS programming (where you should often care about the intermediate states being visible from outside).

raiden_contracts/contracts/services/UserDeposit.sol
karlb: Maybe it's best to `require(sender != receiver)`. That is confusing and I don't see a use for it.

pirapira: Or treat similarly to `amount == 0`.

Anyway, this function is for MSC only.  So it's up to you to decide.

raiden_contracts/tests/test_user_deposit_contract.py
karlb: Good point! Done in f07f76d.

raiden_contracts/tests/test_user_deposit_contract.py
karlb: I've increased the minted balance to make that unambiguous.

I didn't add a separate test for failure due to insufficient funds, since this distinction is not really relevant to the UDC and would rather be a test for the token implementation.

raiden_contracts/contracts/services/UserDeposit.sol
```diff
@@ -0,0 +1,172 @@
...
+        require(contractExists(_msc_address));
+        msc_address = _msc_address;
+    }
+
+    /// @notice Deposit tokens. Idempotent function that sets the
+    /// total_deposit of tokens of the beneficiary.
+    /// Can be called by anyone several times and on behalf of other accounts
+    /// @param beneficiary The account benefiting from the deposit
+    /// @param total_deposit The amount tokens in the user's deposit after the call
+    function deposit(address beneficiary, uint256 total_deposit)
```
pirapira: I see an opportunity to explain another kind of common attacks.

1. a user sends a `deposit()` transaction
2. the transaction is seen worldwide in the pending transaction pool
3. MSC (or somebody operating MSC) decides to move tokens out of `balances[beneficiary]`, and sends a `transfer()` call
4. a miner includes MSC's `transfer()` call.  Now `balances[beneficiary]` is smaller than the user thought.
5. a miner includes the user's `deposit()` transaction.  The user pays more than expected.

What is the reason `total_deposit` is used here, instead of just `added_deposit`?

raiden_contracts/tests/test_user_deposit_contract.py
pirapira: The comment currently only talks about `not enough tokens are approved`.  But I'm not sure if the comment is correct.  Maybe `depositing more than the token contract allows`.

raiden_contracts/contracts/services/UserDeposit.sol
```diff
@@ -0,0 +1,172 @@
...
+
+        balances[beneficiary] += added_deposit;
+        require(token.transferFrom(msg.sender, address(this), added_deposit));
+    }
+
+    /// @notice Internally transfer deposits between two addresses.
+    /// @param sender Account from which the amount will be deducted
+    /// @param receiver Account to which the amount will be credited
+    /// @param amount Amount of tokens to be transferred
+    /// @return true if transfer has been done successfully, otherwise false
```
pirapira: When I read this, I might believe all failures return false.  Document the throwing behavior as well.

raiden_contracts/contracts/services/UserDeposit.sol
```diff
@@ -0,0 +1,172 @@
...
+
+    /// @notice Announce intention to withdraw tokens.
+    /// Sets the planned withdraw amount and resets the withdraw_block.
+    /// There is only one planned withdrawal at a time, the old one gets overwritten.
+    /// @param amount Maximum amount of tokens to be withdrawn
+    function planWithdraw(uint256 amount)
+        external
+    {
+        require(amount > 0);
+        require(balances[msg.sender] >= amount);
```
pirapira: I wonder if this check is necessary.  `deposit()` and `transfer()` can happen during the waiting period anyway.

Is there a document describing how UDC will be used?

raiden_contracts/contracts/services/UserDeposit.sol
loredanacirstea: The history is related to the reason why went for an idempotent deposit for the core contracts in the first place:  
- https://github.com/raiden-network/raiden-contracts/issues/76
- https://docs.google.com/document/d/1y6jW7yg5Cn6Z5QShQbVDhyVAa4Ogp29k5c45VleZr0I/edit?disco=AAAABp-Q2jo

pirapira: @karlb I reopened https://github.com/raiden-network/raiden-contracts/issues/76 . Will you drive the decision making there?

raiden_contracts/contracts/services/UserDeposit.sol
karlb: I don't think this check is strictly necessary. It might be confusing though to get events about a user planning to withdraw much more than his current balance. On the other hand, removing this limit would provide a nice way to withdraw all tokens, including potentially incoming ones, by setting a high withdraw amount.

karlb: The UDC is meant to be used both for https://github.com/raiden-network/raiden-services/issues/38 and as a replacement of the current deposit handling inside the MSC. By unifying this, we only need a single deposit for paying both the PFS and MS.

pirapira: > removing this limit would provide a nice way to withdraw all tokens, including potentially incoming ones, by setting a high withdraw amount.

Then I would always be planning withdraws with very big numbers. What kind of game is being played here between the user and MSC? That's why I asked for a context.

pirapira: Now I see.  These withdraw plans prompt the services to claim their IOUs.  Then, actually the current implementation makes sense to me!

karlb: The effectiveBalance is used by services to check that the user will be able to pay the service with a high likelyhood. Withdraws are immediately deducted from the effectiveBalance when they are planned, so services can claim outstanding payments during the `withdraw_delay` before the user is able to withdraw his deposit.

This is also the reason why a user might not want to plan withdraws with arbitrarily high amounts. If he does, the effectiveBalance will be zero and services will not act upon his requests.

raiden_contracts/tests/test_user_deposit_contract.py
karlb: To me the "not enough approved" and "more than allowed" have the same meaning in this context. Am I missing a subtle difference?

pirapira: "more than allowed" == ("'more than the available balance" or "more than approved").

raiden_contracts/contracts/services/UserDeposit.sol
pirapira: I'm not seeing any comments for the possible `require()` failure.  In the end people's fund might be locked up because of this.

raiden_contracts/utils/logs.py
```diff
@@ -89,6 +89,16 @@ def wait(self, seconds):
                 raise Exception(message + ' waiting_events ' + str(waiting_events),
                                 ' len(self.event_unkown) ' + str(len(self.event_unknown)))
 
+    def assert_event(self, txn_hash, event_name, args, timeout=5):
+        """ Assert that `event_name` is emmited with the `args`
```
pirapira: `emmited` --> `emitted`.  Don't ask me why though.

PR #451, opened 2019-01-21
Add tests for deployment script functions: `token` & `register`

loredanacirstea:
- test `deploy_token_contract()` used in the deployment script
- test `register_token_network()` used in the deployment script

The test checks if the above functions work correctly in the happy case,
to make sure no code dependencies have been changed, affecting the deployment script.
The test does not check however that the cli command works correctly.


loredanacirstea:
@pirapira done.

PR #485, opened 2019-01-31
Basic OneToN payment implentation.

karlb:
First implementation of the contract for PFS payments described in https://github.com/raiden-network/raiden-services/issues/38. Another function to reduce gas usage by claiming multiple IOUs at once and freeing old storage slots will follow.


raiden_contracts/constants.py
```diff
@@ -97,10 +98,8 @@ class MessageTypeId(IntEnum):
 
 # Message types used by MonitoringService contract
 class MonitoringServiceEvent(str, Enum):
-    NEW_DEPOSIT = 'NewDeposit'
     NEW_BALANCE_PROOF_RECEIVED = 'NewBalanceProofReceived'
     REWARD_CLAIMED = 'RewardClaimed'
-    WITHDRAWN = 'Withdrawn'
```
pirapira: I missed these last time.

raiden_contracts/contracts/services/OneToN.sol
```diff
@@ -0,0 +1,125 @@
...
+            sender,
+            receiver,
+            amount,
+            expiration_block,
+            signature
+        );
+        require(addressFromSignature == sender);
+
+        // must not be claimed before
+        bytes32 _key = keccak256(abi.encodePacked(receiver, sender, expiration_block));
```
pirapira: Why is `_key` prefixed with an underscore?  (Pure curiosity.)

karlb: Heiko did that when sketching his idea and I thought it might be a keyword and left it that way. But looking it up, it's not one of the reserved keywords: https://solidity.readthedocs.io/en/v0.4.21/miscellaneous.html#reserved-keywords . So the "key" word is not a keyword.

Issue #480, opened 2019-01-31
Rename RaidenServiceBundle contract to ServiceRegistry


karlb:
That name is more descriptive and less confusing.

- [x] rename in solidity code
- [x] update spec

PR #489, opened 2019-02-01
 Rename RaidenServiceBundle to ServiceRegistry

karlb:
That name is more descriptive and less confusing.

See #480


karlb:
> Are these profiling files added on purpose? https://github.com/raiden-network/raiden-contracts/pull/489/files#diff-2a89d1d17eabca6a39aea3da17a42342

No, they are not! I'll remove the right away.

PR #491, opened 2019-02-04
Require Python-3.7

pirapira:
following raiden-network/raiden.

https://github.com/raiden-network/raiden/pull/3350/files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ dist: xenial
 sudo: true
 language: python
 python:
-  - '3.6'
+  - '3.7'
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


pirapira:
>  I don't see any reason to constraint the contract's requirements to python 3.7 at the moment

None.  Except: I don't want two versions around.

LefterisJP:
> None. Except: I don't want two versions around.

What do you mean two versions? For a project to be runnable with more than one version of python it's pretty normal and a good thing.

LefterisJP:
>  So I need to have Python 3.7 around. Now the question is if I need to have Python 3.6 around for raiden-contracts.

Having multiple versions of python around is very very common with python development.

> raiden-contracts runnable with both Python 3.6 and Python 3.7 (because I need to take some care that Travis runs both)

You really don't need to do that. This kind of testing is an overkill, **especially** since python is just a tool to deploy the smart contracts here and serves no other deeper purpose in this repository.

--

Again I am saying there is no need to constraint the version to python 3.7 if running it with python 3.6 also works and you are not using any python 3.7 specific features.

Being able to run with the non-bleeding edge of python is a good thing, not a bad thing.

pirapira:
Then I leave it as it is, or go for both Python 3.6 & Python 3.7 https://github.com/raiden-network/raiden-contracts/issues/494 .

pirapira:
> Having multiple versions of python around is very very common with python development.

I'm trying this now. I just need different `emacs` instances on each virtualenv, but I can tolerate this (probably there is a better way too).

LefterisJP:
I recommend [this](https://github.com/jorgenschaefer/pyvenv) emacs package.

No need to change emacs instance. Just have multiple virtual environments and whenever you want to use a different environment in the editor just do `pyvenv-workon your_other_venv` to switch to `your_other_venv`.

This is generally needed since python version is not the only difference between venvs. The requirements may be different, there may be different flake8 rules e.t.c.

Issue #502, opened 2019-02-04
Refactor repetition in verify_deployed_service_contracts()


pirapira:
A very similar sequence is repeated for every service contract in `verify_deployed_service_contracts()`.  This issue keeps track of refactoring these.

Especially, the lines I makred with `!` are completely repetitive.  This issue can be closed when these complete repetition somehow disappears.

```
!    service_bundle = verify_deployed_contract(
!        web3,
!        contract_manager,
!        deployment_data,
!        CONTRACT_SERVICE_REGISTRY,
!    )

!    # We need to also check the constructor parameters against the chain
!    constructor_arguments = deployment_data['contracts'][
!        CONTRACT_SERVICE_REGISTRY
!    ]['constructor_arguments']
    assert to_checksum_address(
        service_bundle.functions.token().call(),
    ) == token_address
    assert token_address == constructor_arguments[0]

!    print(
!        f'{CONTRACT_SERVICE_REGISTRY} at {service_bundle.address} '
!        f'matches the compiled data from contracts.json',
!    )
```

pirapira:
There seems nothing to do left here.

PR #505, opened 2019-02-04
Move common verify_deployed_contract() usage into the function body

pirapira:
the function body.

This is one proposal for solving https://github.com/raiden-network/raiden-contracts/issues/502.

I'm not super-confident with the direction I took: contract-specific checks as function arguments.

### DIFF
--- a/raiden_contracts/deploy/__main__.py
+++ b/raiden_contracts/deploy/__main__.py
@@ -664,15 +664,11 @@ def verify_deployed_contracts(web3: Web3, contract_manager: ContractManager, dep
     assert contract_manager.contracts_version == deployment_data['contracts_version']
     assert chain_id == deployment_data['chain_id']
 
-    endpoint_registry = verify_deployed_contract(
+    verify_deployed_contract(
         web3,
         contract_manager,
         deployment_data,
         CONTRACT_ENDPOINT_REGISTRY,
-    )
-    print(
-        f'{CONTRACT_ENDPOINT_REGISTRY} at {endpoint_registry.address} '
-        f'matches the compiled data from contracts.json',
     )
 
     secret_registry = verify_deployed_contract(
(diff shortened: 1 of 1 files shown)


raiden_contracts/deploy/__main__.py
```diff
@@ -744,99 +729,70 @@ def verify_deployed_service_contracts(
...
-    assert token_address == constructor_arguments[0]
-
-    print(
-        f'{CONTRACT_SERVICE_REGISTRY} at {service_bundle.address} '
-        f'matches the compiled data from contracts.json',
-    )
-
+    def user_deposit_checker(deploy_record, onchain_instance):
+        # We need to also check the constructor parameters against the chain
+        constructor_arguments = deploy_record['constructor_arguments']
```
karlb: This line is in all checkers, so `verify_deployed_contract` could call each checker directly with `constructor_arguments`.

raiden_contracts/deploy/__main__.py
```diff
@@ -664,55 +664,40 @@ def verify_deployed_contracts(web3: Web3, contract_manager: ContractManager, dep
...
+        ) == secret_registry.address
+        assert secret_registry.address == constructor_arguments[0]
+
+        chain_id = onchain_instance.functions.chain_id().call()
+        assert chain_id == constructor_arguments[1]
+
+        settlement_timeout_min = onchain_instance.functions.settlement_timeout_min().call()
+        settlement_timeout_max = onchain_instance.functions.settlement_timeout_max().call()
+        assert settlement_timeout_min == constructor_arguments[2]
+        assert settlement_timeout_max == constructor_arguments[3]
```
karlb: I would do the assert directly if it fits in a single line, but that's a matter of taste:
```python
assert onchain_instance.functions.settlement_timeout_max().call() == constructor_arguments[3]
```

pirapira:
@karlb 

> but if you want to avoid passing checker functions around, you could do

I like the return-value approach.  So, I will try that and open the PR again.

raiden_contracts/deploy/__main__.py
```diff
@@ -744,99 +729,70 @@ def verify_deployed_service_contracts(
...
+        assert to_checksum_address(
+            onchain_instance.functions.deposit_contract().call(),
+        ) == user_deposit.address
+        assert user_deposit.address == constructor_arguments[0]
+        assert len(constructor_arguments) == 1
+    verify_deployed_contract(
         web3,
         contract_manager,
         deployment_data,
         CONTRACT_ONE_TO_N,
```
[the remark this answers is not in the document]
pirapira: I take this as an indication that I took a too fancy approach.

raiden_contracts/deploy/__main__.py
pirapira: Yes.

I still wonder why a special treatment is necessary for constructor arguments, but sure, that's the way it is.

Issue #490, opened 2019-02-03
Failed to deploy contracts on private network


offerm:
Trying to deploy I have the following error
```
(raiden-contracts) admins-MacBook-Pro-2:raiden-contracts admin$ python -m raiden_contracts.deploy raiden --rpc-provider http://127.0.0.1:8545 --private-key $private_key --gas-price 10 --gas-limit 6000000 --contracts-version 0.4.0
Web3 provider is RPC connection http://127.0.0.1:8545
Enter the private key password: 
INFO:__main__:Skipped checks against the source code because it is not available.
DEBUG:__main__:Deploying EndpointRegistry txHash=0xf35cba813872823d3735528b85161a1cc608ccf7ae15d7f800c04401c7bc00da, contracts version 0.4.0
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.2_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python/3.7.2_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/admin/xud-eth/raiden-contracts/raiden_contracts/deploy/__main__.py", line 831, in <module>
    main()
  File "/Users/admin/.virtualenvs/raiden-contracts/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/admin/.virtualenvs/raiden-contracts/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/admin/.virtualenvs/raiden-contracts/lib/python3.7/site-packages/click/core.py", line 1163, in invoke
    rv.append(sub_ctx.command.invoke(sub_ctx))
  File "/Users/admin/.virtualenvs/raiden-contracts/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/admin/.virtualenvs/raiden-contracts/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/admin/xud-eth/raiden-contracts/raiden_contracts/deploy/__main__.py", line 177, in wrapper
    return func(*args, **kwargs)
[...]

PR #506, opened 2019-02-04
Note Byzantium version is necessary

pirapira:
[no ci]

This closes https://github.com/raiden-network/raiden-contracts/issues/490


pirapira:
@offerm is this the right place to warn against #490?

Issue #507, opened 2019-02-04
make [no ci] flag work


pirapira:
Travis still runs even when I add `[no ci]` in the description of a PR.  This incurs unnecessary waiting time for PRs with only document changes.

pirapira:
`[skip ci]` should work.  https://docs.travis-ci.com/user/customizing-the-build/#skipping-a-build

pirapira:
If I say `[skip ci]`, GitHub waits for Travis forever.
```
continuous-integration/travis-ci Expected — Waiting for status to be reported
```

So it's not a good solution.

LefterisJP:
This is what we have always been doing. You can still merge the PR with a force-merge if you know it's only docs.

As an alternative you can make a special [skip tests] flag that would only do some linting checks and run no other jobs.

pirapira:
Ah, I remember how it was in `raiden` repo.  Good for me.

PR #508, opened 2019-02-04
Document gas measuring for releasing

### LINKED ISSUE #492: Document the gas cost update process
`RELEASE.rst` should make sure `constants.py` is up-to-date about the gas cost for each operation.

pirapira:
This fixes #492.

[no ci] <-- seems not working https://github.com/raiden-network/raiden-contracts/issues/507

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -7,6 +7,7 @@ Package Deliverables
 - smart contracts source code from ``raiden_contracts/contracts``
 - compiled contracts data from ``raiden_contracts/data/contracts.json``
 - deployment information from ``raiden_contracts/data/deployment_*.json``, with Ethereum addresses, transaction data (transaction hash, block number, constructor arguments, gas cost)
+- gas costs information from ``raiden_contracts/constants.py``
 - scripts for deployment & contract verification on Etherscan from ``raiden_contracts/deploy``
 - source code tests from ``raiden_contracts/tests``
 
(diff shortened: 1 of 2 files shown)


RELEASE.rst
```diff
@@ -106,6 +108,25 @@ Etherscan verification is documented here: https://github.com/raiden-network/rai
...
+Measure Gas Costs
+-----------------
+
+``raiden_contracts`` package provides `some constants <https://github.com/raiden-network/raiden-contracts/blob/de13cf9aa7ad7ed230ff204e47103def6a14b0be/raiden_contracts/constants.py#L35>`__ showing the amount of gas that each operation requires. This information is manually updated. The amounts can be measured with a script
+
+::
+
+    pytest -s raiden_contracts/tests/test_print_gas.py
+
+The script prints many numbers like
```
LefterisJP: @pirapira Wouldn't it make more sense to remake the script into `measure_gas.py` and have it auto-generate the `constants.py` for you instead of you manually creating by copying the output of the script?

pirapira: That's https://github.com/raiden-network/raiden-contracts/issues/16.

hackaugusto:
I don't think this should have been merged, the gas cost will change with the version of the smart contract, so this data should go into the json (which has the additional benefit of working with other languages, besides python)

LefterisJP:
@hackaugusto That's also a nice idea -- but perhaps going for other language other than python may be a premature optimization at the moment. But according to @pirapira 's comment [above](https://github.com/raiden-network/raiden-contracts/pull/508#discussion_r253561737) I understand that what this PR documents is temporary and the "real thing" approach is discussed and documented in this [issue](https://github.com/raiden-network/raiden-contracts/issues/16)

pirapira:
@hackaugusto I do document the current ugly approach before changing it.

PR #509, opened 2019-02-04
Document the use of [skip ci]

pirapira:
And this commit should skip the CI.  Related: #507


pirapira:
If I say `[skip ci]`, GitHub waits for Travis forever.
```
continuous-integration/travis-ci Expected — Waiting for status to be reported
```

So it's not a good solution.

Issue #482, opened 2019-01-31
Practice initialization calls on UDC, ServiceRegistry and MS contracts


pirapira:
This issue keeps track of practicing the initialization calls on
* UserDepositContract
* ServiceRegistry contract
* MS contract
on
* kovan
* Ropsten.

This issue also keeps track of adjusting the deployment scripts.

Target date: 2019-02-05.

pirapira:
* ServiceRegistry requires no initialization calls.
* UDC requires `init(msc, one_to_n)`
* MonitoringService requires no initialization calls.
* OneToN requires no initialization calls.

pirapira:
I tested #512 on Kovan and Ropsten.

PR #512, opened 2019-02-05
Call UserDeposit.init() during the service contract deployment

pirapira:
This is a part of #482 .  I tried the script with Kovan.

### DIFF
--- a/raiden_contracts/deploy/__main__.py
+++ b/raiden_contracts/deploy/__main__.py
@@ -129,10 +129,10 @@ class ContractDeployer:
 
     def transact(
         self,
-        to_be_called,
+        contract_method: Contract,
     ):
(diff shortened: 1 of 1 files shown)


raiden_contracts/deploy/__main__.py
```diff
@@ -127,6 +127,16 @@ def deploy(
         )
         return receipt
 
+    def transact(
+        self,
+        to_be_called,
```
karlb: Let's add types to all new production code (we can be less strict with tests)! In this case `Callable`.

pirapira: OK.  Will do.

pirapira: Turns out `mypy` caught my misconception.

PR #518, opened 2019-02-05
Add join-contracts.py script

pirapira:
This PR copies `join-contracts.py` script from `raiden` repository.  (I'll submit a PR there to remove the script from there.)

`join-contracts.py` concatenates Solidity sources that's imported from one source.  This feature is useful for verifying Solidity sources on Etherscan.  The script will provide a much better alternative to [the current manual concatenation and string replacements](https://github.com/raiden-network/raiden-contracts/blob/85f92620671d08c8f9a090b6f1bd2d98a74e3844/raiden_contracts/deploy/etherscan_verify.py#L91).

I think this script should live in `raiden-contracts` rather than in `raiden`.  Both repos are under the MIT license.

---

Also this PR renames `raiden_contracts.utils.type` into `raiden_contracts.utils.type_aliases`.  When I ran `join-contracts.py`, I saw an error (see [the commit message](https://github.com/raiden-network/raiden-contracts/commit/362b1d1e9a55cdee20de746c5a34cc00596d8812)) because [types](https://docs.python.org/3/library/types.html) was hidden.


raiden_contracts/utils/join-contracts.py
```diff
@@ -0,0 +1,71 @@
...
+IMPORT_RE = re.compile(r'^import +["\'](?P<contract>[^"\']+.sol)["\'];$')
+
+"""
+Utility to join solidity contracts into a single output file by recursively
+resolving imports.
+
+example usage:
+
+$ cd raiden/smart_contracts
+$ python ../../tools/join-contracts.py SomeContractWithImports.sol joined.sol
```
czepluch: Can you update this comment to use `utils` instead of `tools`?

pirapira: Yes, will do.

PR #520, opened 2019-02-05
Mention type annotations in the pull-request guide

pirapira:
Echoing @karlb 's [comment](https://github.com/raiden-network/raiden-contracts/pull/512#discussion_r253828354).

[skip ci]


pirapira:
In Solidity, there is no choice anyway. It's hard (I think impossible) to skip types of arguments or variables.

PR #521, opened 2019-02-05
Fix join-contracts.py without --import-map

pirapira:
Before this commit, join-contracts.py without --import-map failed with
an error message:
```
  File "utils/join-contracts.py", line 66, in main
    import_map = json.loads(import_map)
  File "/usr/lib/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

This issue fixes the problem by changing the default value from "" to
"{}".

I know this script is not used anywhere yet, so [skip ci].

### DIFF
--- a/raiden_contracts/utils/join-contracts.py
+++ b/raiden_contracts/utils/join-contracts.py
@@ -16,7 +16,7 @@ resolving imports.
 example usage:
 
 $ cd raiden-contracts/raiden_contracts
-$ python ./utils/join-contracts.py --import-map {} ./contracts/TokenNetwork.sol joined.sol
+$ python ./utils/join-contracts.py ./contracts/TokenNetwork.sol joined.sol
 
 """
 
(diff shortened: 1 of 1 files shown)


pirapira:
That sounds like a cleaner way but forgive my laziness.

My current aim is to remove these concatenations: https://github.com/raiden-network/raiden-contracts/blob/fdede381d1f205d404c67af988e2574df69eb1d9/raiden_contracts/deploy/etherscan_verify.py#L94

Issue #530, opened 2019-02-06
etherscan_verify.py timeouts


pirapira:
When I use `etherescan_verify.py` script on `TokenNetworkRegistry` contract, I get

```
Retrying...                                                      
{'status': '0', 'message': 'NOTOK', 'result': 'Pending in queue'}
Retrying...                                                      
{'status': '0', 'message': 'NOTOK', 'result': 'Pending in queue'}
Retrying... 
```
unltil it timeouts.

However, when I visit Etherscan on a web browser and post the Solidity source, the source verification succeeds within 3 seconds.

pirapira:
I reported the problem to Etherscan and I have a ticket number about this.

pirapira:
I checked out `v0.9.0` of `raiden-contracts` and tried `etherscan-verify.py` there on `SecretRegistry`.  Even that did not work today.

pirapira:
They fixed the issue.

PR #531, opened 2019-02-06
Use join-contracts.py script for Etherscan verification

### LINKED ISSUE #525: Solidity source concatenation using join-contracts.py
This issue keeps track of using `join-contracts.py` from `etherscan_verify.py`.

Currently`etherscan_verify.py` uses string concatenation and string replacements to combine solidity codes into one file.  The code contains the names of all imported files and a specific pragma string:

https://github.com/raiden-network/raiden-contracts/blob/fdede381d1f205d404c67af988e2574df69eb1d9/raiden_contracts/deploy/etherscan_verify.py#L94

Before adding more contracts to `etherscan_verify.py`, I want to clean the script.

pirapira:
This pull-request removes the string concatenation in `etherscan_verify.py` and instead uses `join-contracts.py`.  This closes #525.

Some refactoring will follow (now possible because this PR removes contract-specific code).

Even after this change, Etherscan verification timeouts: https://github.com/raiden-network/raiden-contracts/issues/530 but at least I added an error message explaining how to get the source verified in a web browser.

### DIFF
--- a/raiden_contracts/deploy/etherscan_verify.py
+++ b/raiden_contracts/deploy/etherscan_verify.py
@@ -166,12 +166,13 @@ def etherscan_verify_contract(
     print(content)
     print(f'Status: {content["status"]}; {content["message"]} ; GUID = {content["result"]}')
 
-    if content["status"] == "1":
+    if content["status"] == "1":  # submission succeeded, obtained GUID
+        guid = content["result"]
         status = '0'
         retries = 10
         while status == '0' and retries > 0:
             retries -= 1
-            r = guid_status(etherscan_api, content["result"])
+            r = guid_status(etherscan_api, guid)
             status = r['status']
             if r['result'] == 'Fail - Unable to verify':
                 return
--- a/raiden_contracts/deploy/etherscan_verify.py
+++ b/raiden_contracts/deploy/etherscan_verify.py
@@ -175,8 +175,17 @@ def etherscan_verify_contract(
             status = r['status']
             if r['result'] == 'Fail - Unable to verify':
                 return
+            if r['result'] == 'Pass - Verified':
+                return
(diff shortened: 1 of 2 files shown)


raiden_contracts/deploy/etherscan_verify.py
```diff
@@ -90,24 +92,23 @@ def etherscan_verify(chain_id, apikey, guid, contract_name):
...
+            'contracts/TokenNetworkRegistry.sol',
+            str(joined_file),
+        ]
+        old_working_dir = Path.cwd()
+        chdir(Path(__file__).parent.parent)
+        try:
+            combiner = subprocess.Popen(command)
+            combiner.wait()
+        finally:
+            chdir(old_working_dir)
```
karlb: You can avoid the `wait` call by using `subprocess.run` and the directory handling by using the `pwd` parameter:
```python
    subprocess.run(command, cwd=Path(__file__).parent.parent)
```

raiden_contracts/deploy/etherscan_verify.py
```diff
@@ -174,8 +175,17 @@ def etherscan_verify_contract(
             status = r['status']
```
karlb: Not new in this PR, but a short comment what `content["status"] == "1"` means would be helpful.

raiden_contracts/deploy/etherscan_verify.py
pirapira: I already touch this part in the next PR.

I didn't know `cwd` keyword.  It is really neat.  I'll use it there.

raiden_contracts/deploy/etherscan_verify.py
pirapira: I added a comment.

Issue #477, opened 2019-01-30
Consider using pragmas to specify the exact version of Solidity


pirapira:
This issue keeps track of using Solidity pragmas to choose one specific version of `solc` and remove the check against downgrading `solc`.

Currently this repository has
* Solidity sources saying "compile this with `solc` version x.y.z or above" `pragma solidity ^0.5.2;`
* and a check against downgrading `solc` https://github.com/raiden-network/raiden-contracts/blob/5a117d52fa1df5a28483c4a4c7718a4c26d65b18/raiden_contracts/contract_manager.py#L81

But why do we need to check for `solc` downgrades?
a. the compilation happens on a developer's machine, and
b. the pragma does not specify the exact version of `solc` to be used

I think one version of this codebase should choose a specific version of `solc`.

Thanks @berndbohmeier for pointing this out.

PR #539, opened 2019-02-07
Specify the exact solc version

pirapira:
Before this commit, Solidity sources allowed many solc versions that are
above 0.5.2. After this commit only 0.5.2 works.

As a result, we no longer need to check for downgraded solc versions.

This closes
https://github.com/raiden-network/raiden-contracts/issues/477


karlb:
What would be the problem with using a higher solc version?

pirapira:
> What would be the problem with using a higher solc version?

The resulting bytecode depends on the local environment.  Somebody later uses the original lower version and so on.

This PR came after the following conversation:

1. Bernd asked "for example, what does this [compare_old_solc_version](https://github.com/raiden-network/raiden-contracts/pull/539/files#diff-9ceee07353ca99dee98716658f1f4b7fL118) do?"
2. I answered "because there will be some people with higher `solc` version and some other people with lower `solc` version "
3. Bernd suggested "well, you can specify the Solidity version exactly, and you don't need the function `compare_old_solc_version`".

And I thought that's a good idea.

pirapira:
Don't merge until 0.10.0 is released.

PR #541, opened 2019-02-11
Register hostname in ServiceRegistry

karlb:
The ServiceRegistry is still mostly unspecced and it's not clear how quickly that will change. This is mostly intended as a placeholder for the real ServiceRegistry to avoid the PFS development from being blocked due to the lack of any registry for usable HTTP services.

Let's talk about whether adding this is a good idea at this point.


raiden_contracts/contracts/services/ServiceRegistry.sol
```diff
@@ -27,4 +29,20 @@ contract ServiceRegistry is Utils {
...
         require(token.transferFrom(msg.sender, address(this), amount));
     }
+
+    /// Set the URL used to access a service via HTTP.
+    /// When this is called for the first time, the service's ethereum address
+    /// is also added to `service_addresses`.
+    function setURL(string memory new_url) public {
+        require(bytes(new_url).length != 0);
+        if (bytes(urls[msg.sender]).length == 0) {
+            service_addresses.push(msg.sender);
```
pirapira: Is this a constant-gas operation?

karlb: Not an official source, but according to https://ethereum.stackexchange.com/questions/30843/is-the-cost-of-pushing-proportional-to-the-arrays-size and https://www.reddit.com/r/ethereum/comments/4usq1q/is_the_cost_of_arraypush_a_constant/d5sn2lb/ it is.

raiden_contracts/tests/test_service_registry.py
```diff
@@ -0,0 +1,53 @@
...
+    service_registry.functions.setURL(url1).transact({'from': A})
+    assert service_registry.functions.urls(A).call() == url1
+    assert service_registry.functions.service_addresses(0).call() == A
+    assert service_registry.functions.serviceCount().call() == 1
+    with pytest.raises(TransactionFailed):
+        assert service_registry.functions.service_addresses(1).call()
+
+    # Setting the host for the second time must not add the service address a
+    # second time.
+    service_registry.functions.setURL(url2).transact({'from': A})
```
pirapira: I guess you can add `with pytest.raises(TransactionFailed):` here as well.

PR #557, opened 2019-02-12
Revert "New gas measurement results"

pirapira:
This reverts commit 455198f996d337106b796612dd1a03357f98f51c because the new gas numbers caused test failures in https://github.com/raiden-network/raiden/pull/3470

When this is merged, there will be release 0.10.1.


pirapira:
I can already bump the version number here!

LefterisJP:
Hmmm I am not sure if this is the right solution. Since we have multiple versions of the contracts we should also have multiple versions of the gas measurements.

PR #562, opened 2019-02-14
More checks in CI tests

pirapira:
Each commit adds one more check in the CI tests.  Making sure an event is fired, or checking the post-state and so on.

### DIFF
--- a/raiden_contracts/tests/test_channel_close.py
+++ b/raiden_contracts/tests/test_channel_close.py
@@ -293,12 +293,14 @@ def test_close_first_participant_can_close(
         token_network,
         create_channel,
         get_accounts,
+        get_block,
 ):
     """ Simplest successful closeChannel by the first participant """
     (A, B) = get_accounts(2)
+    settle_timeout = TEST_SETTLE_TIMEOUT_MIN
     channel_identifier = create_channel(A, B)[0]
 
-    token_network.functions.closeChannel(
+    close_tx = token_network.functions.closeChannel(
         channel_identifier,
         B,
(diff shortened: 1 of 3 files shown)


raiden_contracts/tests/test_channel_close.py
```diff
@@ -287,12 +293,14 @@ def test_close_first_participant_can_close(
         token_network,
         create_channel,
         get_accounts,
+        get_block,
 ):
     """ Simplest successful closeChannel by the first participant """
     (A, B) = get_accounts(2)
+    settle_timeout = TEST_SETTLE_TIMEOUT_MIN
```
karlb: No need to give this a local name, IMO.

raiden_contracts/tests/test_channel_close.py
```diff
@@ -301,6 +309,37 @@ def test_close_first_participant_can_close(
...
+
+    (
+        _, _,
+        A_is_the_closer,
+        A_balance_hash,
+        A_nonce,
+        _,
+        _,
+    ) = token_network.functions.getChannelParticipantInfo(channel_identifier, A, B).call()
+    assert A_is_the_closer is True
```
karlb: Some people would say that `... is True` is syntactic noise and you should just `assert A_is_the_closer`. But in this case it's not totally obvious from the context that `A_is_the_closer` is a boolean, so this does make it a tiny bit clearer.

pirapira: I looked around. `... is True` is consistently used in this file, so I follow the local custom for now.

PR #568, opened 2019-02-14
Deployed new instances on test networks with fake service tokens

pirapira:
Before this PR, the service fee on test networks was in WETH.  Form this version on, the testnet fee payments will be made in a specially deployed ServiceToken.


pirapira:
Ah, perhaps I didn't need to bump the contract version.  I didn't think about that.

`raiden-contract` provides access to specific instances on the chain.  This case I wanted to provide the access through packaging.

PR #583, opened 2019-02-19
Enable mypy for functions without type declaration

karlb:
Mypy will just assume `Any` for all types, but that still allows to
catch errors because types will be inferred from signatures in other
places.

### DIFF
--- a/raiden_contracts/tests/fixtures/contracts.py
+++ b/raiden_contracts/tests/fixtures/contracts.py
@@ -1,14 +1,5 @@
 import pytest
 import logging
-from web3.utils.events import get_event_data
-from eth_utils import is_address
-
-from raiden_contracts.utils.transaction import check_succesful_tx
-from raiden_contracts.constants import (
-    CONTRACT_TOKEN_NETWORK,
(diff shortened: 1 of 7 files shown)


raiden_contracts/tests/fixtures/contracts.py
```diff
@@ -99,30 +90,3 @@ def f(contract_name, libs=None, args=None):
...
-        CONTRACT_TOKEN_NETWORK_REGISTRY,
-        EVENT_TOKEN_NETWORK_CREATED,
-    )
-    decoded_event = get_event_data(event_abi, tx_receipt['logs'][0])
-    assert decoded_event is not None
-    assert is_address(decoded_event['args']['token_address'])
-    assert is_address(decoded_event['args']['token_network_address'])
-    token_network_address = decoded_event['args']['token_network_address']
-    token_network_abi = contracts_manager.get_contract_abi(CONTRACT_TOKEN_NETWORK)
-    return web3.eth.contract(abi=token_network_abi, address=token_network_address)
```
pirapira: :+1:

PR #587, opened 2019-02-19
Make dir layout match import statements

karlb:
This is less confusing and allows easier setup of dev tools for new
developers.


raiden_contracts/contract_manager.py
```diff
@@ -84,6 +84,7 @@ def relativise(path):
             '%s=%s' % (k, relativise(v))
             for k, v in self.contracts_source_dirs.items()
         ]
+        import_dir_map.append('.=.')  # allow solc to compile contracts in all subdirs
```
pirapira: `solc` needs `'.=.'` as the first element of `import_dir_map`.

karlb:
@pirapira Any idea what I'm doing wrong, here? It works locally on my machine. Do I need to use `_BASE` instead of `.` or call `relativize` on it for some reason? I can't see what's so different on Travis.

pirapira:
I guess the fix is, instead of appending `'.=.'` at the end, putting it at the beginning of `import_dir_map`.
```
import_dir_map.append('.=.')
```

karlb:
> I guess the fix is, instead of appending `'.=.'` at the end, putting it at the beginning of `import_dir_map`.
> 
> ```
> import_dir_map.append('.=.')
> ```

Unfortunately, that is not the solution:

> E           solc.exceptions.SolcError: An error occurred during execution
E                   > command: `solc --combined-json abi,bin,bin-runtime,metadata,ast .=. raiden=contracts/raiden test=contracts/test services=contracts/services contracts/raiden/TokenNetworkRegistry.sol contracts/raiden/SecretRegistry.sol contracts/raiden/Utils.sol contracts/raiden/TokenNetwork.sol contracts/raiden/Token.sol contracts/raiden/EndpointRegistry.sol`
E                   > return code: `1`
E                   > stderr:
E           
E                   > stdout:
E                   contracts/raiden/TokenNetwork.sol:5:1: Error: Source "contracts/raiden/lib/ECVerify.sol" not found: File not found.
E           import "raiden/lib/ECVerify.sol";
E           ^-------------------------------^

Debugging problems that do not occur locally is always a hassle, so I'm happy about any suggestions.

pirapira:
I see the error locally.  No clue why...

pirapira:
Now I see
```
solc.exceptions.SolcError: An error occurred during execution
> command: `solc --combined-json abi,bin,bin-runtime,metadata,ast .=contracts raiden=contracts/raiden test=contracts/test services=contracts/services`
> return code: `1`
> stderr:

> stdout:
No input files given. If you wish to use the standard input please specify "-" explicitly.
```
on 8e693bd6ec17fe529d5761856da94a51d58a3aff

karlb:
> Please add a Changelog entry.

Done (in the force push).

karlb:
Thanks for taking this over!

I moved `ECVerify.sol` to `raiden/lib`, you added it back to `lib`, so now we have both. We should decide which is the right place and delete the other file.

pirapira:
@karlb ahm 26bcea5bf5a3758910a9b6ea77f124e77aa7c465 seems like purely removing `ECVerify.sol`.  I think the culprit is `.gitignore` that ignores `lib/`.

karlb:
> I think the culprit is .gitignore that ignores lib/.

That's an excellent point, thanks for pointing that out! That explains all the trouble I had with differences between Travis in my machine, too.

I'll prepare a PR that changes the `.gitignore` to have stricter matches (e.g. only root dir for many entries).

pirapira:
Yes, please.  The current situation is rather confusing.

PR #590, opened 2019-02-20
Gas measurements in a JSON file in data directory

### LINKED ISSUE #547: Modify test_print_gas.py to store a json file in data/ directory
This issue keeps track of modifying `test_print_gas.py` to store results in a JSON file in `data` directory, in addition to the current functionality of printing gas consts in `stdout`.

This issue also keeps track of modifying the release process so that this script is run before every package release.

pirapira:
This PR
* adds a new file `raiden_contracts/data/gas.json` that lists the amount of gas required for important operations
* changes `test_print_gas.py` so that an execution updates the new JSON file

This closes #547.

This PR solves these problems:
* the global `constants.py` was not a good place for gas amounts (because gas consumption is different for each version of smart contracts)
* the release process involved manual copy-and-pasting of gas amounts


raiden_contracts/tests/fixtures/base/utils.py
```diff
@@ -124,6 +133,9 @@ def get(txn_hash, message=None, additional_gas=0):
         print('----------------------------------')
         print('GAS USED ' + message, gas_used + additional_gas)
         print('----------------------------------')
+        gas_measurement_results[message] = gas_used + additional_gas
+        with contracts_gas_path().open(mode='w') as target_file:
+            target_file.write(json.dumps(gas_measurement_results))
```
karlb: Adding `sort_keys=True, indent=4` to `json.dumps` will give more readable output with cleaner diffs.

pirapira: Very good idea.  Will do.

pirapira: Done.  Now just waiting for CI.

PR #594, opened 2019-02-20
README only three contract_versions are available

pirapira:
Now I'm documenting what's happening right now.  Maybe is it reasonable to provide deployment information of all contract versions?

[skip ci]

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -53,6 +53,9 @@ If you are using the ``raiden-contracts`` package in your project, you can use::
         EVENT_TOKEN_NETWORK_CREATED,
     )
 
+    # ``raiden-contracts`` provides only three kinds of ``contracts_version``s.
+    # ``None``, ``'0.4.0'`` and ``'0.3._'``.  They are explained below.
+
     # Uses the newest version. Available on Kovan, Rinkeby and Ropsten.
     # Mainnet deployments should not be accessed with contract_version == None.
     contracts_version = None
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -67,6 +70,10 @@ If you are using the ``raiden-contracts`` package in your project, you can use::
     # See https://github.com/raiden-network/raiden-contracts/issues/543
     contracts_version = '0.3._'
 
+    # Other contract versions like ``'0.6.0'`` do not work. Contract versions
```
rakanalh: I don't think there's a need for this line. You could end up using 0.6.0 if you're using the raiden-contract package version `0.10.1` or the latest if you use the latest raiden-contracts package. So the comment above is sufficient.

pirapira: I'll write a unit test to see whether `0.6.0` works or not.

pirapira: Indeed any string works here.
```
    nonexistent_version = 'something_that_does_not_exist'
    manager = ContractManager(contracts_precompiled_path(nonexistent_version))
    manager.verify_precompiled_checksums()
```
causes no exceptions.  I think this is a bug.

pirapira: I filed an issue about this https://github.com/raiden-network/raiden-contracts/issues/601

pirapira: No.  `0.6.0` does not work as I documented here.

As a proof, I added [a test](https://github.com/raiden-network/raiden-contracts/pull/594/files#diff-f230759517889d0dd3c8082a1125e6f5R25).

pirapira: @rakanalh please confirm the test ^ shows `0.6.0` does not work.

rakanalh: As i told you earlier, if you have 0.10.1 installed, then `0.6.0` would work. If you have a later version, it wouldn't. That's why i noted that we don't need to add a line of doc saying that a specific version doesn't work because it depends on what version of the package you've installed.

pirapira: We are not agreeing about facts.  Let me write another test about `0.7.0`.  I believe it doesn't work but you believe it works.

pirapira: I added a test, where "0.6.0" doesn't work.  And I wrote in the README that "0.6.0" doesn't work.  Now what's the problem.

This does not depend on the package version.  It depends on the fact that `data_0.6.0` does not exist in the filesystem.

In a future version, if it changes, and when suddenly "0.6.0" works, CI will detect the change.

pirapira:
I'm sorry but I cannot continue seeing people putting wrong options, due to lack of documentation.

Issue #599, opened 2019-02-21
Provide an unlimited version for testing `raiden`


pirapira:
Many CI tests in `raiden` repo expect versions of smart contracts without limits.  So `raiden-contracts` should provide smart contracts without any limits.

pirapira:
I don't have any source of unlimited versions.  So I need to figure out a few things.

* should the unlimited version have really no limits (not even anything like `0.3._`'s limits)?
* should the unlimited version have the deprecation switch?
* I guess arbitrary numbers of ERC20 tokens can be registered to TokenNetworkRegistry
* I guess there are no other differences between the limited version and the unlimited version.

LefterisJP:
>     * should the unlimited version have really no limits (not even anything like `0.3._`'s limits)?

yes

>     * should the unlimited version have the deprecation switch?

no

>     * I guess arbitrary numbers of ERC20 tokens can be registered to TokenNetworkRegistry

yes

>     * I guess there are no other differences between the limited version and the unlimited version.

exactly. Only differences that should exist is everything that has to do with control of the contracts, like limits and deprecation switch. In all other ways they should be ABI compatible.

pirapira:
* where to store unlimited versions
    * `data_x.y.z-unlimited` should be fine.
* [ ] write a script to generate unlimited versions of Solidity code (and add a `Makefile` goal)
* [ ] modify `make compile_contracts` so it generates `contracts-unlimited.json` in addition to `contracts.json`
* [ ] modify `make verify_contracts` so it checks `contracts-unlimited.json` in addition to `contracts.json`
* [ ] change RELEASE.rst 
* [ ]

pirapira:
^ is the proper way.  The quick way is to introduce `data_0.8._`.

pirapira:
@LefterisJP should the constructor of `TokenNetwork` take `_deprecation_executor` argument?

pirapira:
^ Ah, but it shouldn't matter.  You won't call that constructor yourself.

LefterisJP:
@pirapira you seem to have answered your question. But I am not sure exactly what the question was. So the `_deprecation_executor` sounds like an address that can deprecate the token network, right? This would only be for a single token network since that's where the limits are.

And at the moment the way we deprecate the token network is at the registry contract.

pirapira:
I'm changing `TokenNetwork.sol` like

```
    constructor(
        address _token_address,
        address _secret_registry,
        uint256 _chain_id,
[...]

PR #600, opened 2019-02-21
Provide an unlimited test deployment at `data_0.8._`.  At the same time release v0.12.0.

pirapira:
This PR removes limits from the Solidity sources temporarily and provides a deployment of the unlimited variant on the test networks.  This might heal the CI test of https://github.com/raiden-network/raiden/pull/3511


pirapira:
This is the quick ad-hoc reaction to #599.  The proper solution is yet to come.

pirapira:
The plan is to
* keep the `data_0.8._` directory forever
* recover the limits in the Solidity sources shortly after the release
* properly deliver the unlimited testing deployments (https://github.com/raiden-network/raiden-contracts/issues/599)

Also I want to make sure that the limited version is also used in `raiden` CI tests.

pirapira:
@rakanalh @LefterisJP here is the unlimited version.  After this PR is merged I can push a tag, and a new PyPI package will appear.

raiden_contracts/constants.py
```diff
@@ -1,7 +1,7 @@
 from enum import Enum, IntEnum
 
 # Do not change this, this is handled by bumpversion with .bumpversion_contracts.cfg
-CONTRACTS_VERSION = "0.7.0"
+CONTRACTS_VERSION = "0.8._"
```
rakanalh: Out of curiosity... what does the underscore mean in this case?

pirapira: The message is "This is something similar to `0.3._`".

pirapira:
@rakanalh ah, sorry.  Will try to sort this out before I run and catch the train.

raiden_contracts/constants.py
LefterisJP: Hehe ... the only reason `0.3._` had an underscore for the patch part of the version was to signify that there is no patch version and that each change bumps the minor version. Or something silly like that. No need to repeat that now.

raiden_contracts/tests/test_token_network.py
```diff
@@ -18,6 +18,7 @@ def test_version(token_network):
     assert token_network.functions.contract_version().call() == CONTRACTS_VERSION
 
 
+@pytest.mark.skip(reason="Only for v0.12.0 release")
 def test_constructor_call(
```
LefterisJP: For those you skip like this I would say you should make a fixture that checks the releases and conditionally skips for the non-limits release numbers. Does not need to be part of this PR.

raiden_contracts/data
```diff
@@ -0,0 +1 @@
+data_0.8._
```
LefterisJP: Not required for this PR but eventually you should put all these `data_X.Y.Z` directories under a common `data/` directory.

pirapira: Well, I'm being nice here.  Put `0.8._` in `raiden` codebase and that will keep working for a while.

raiden_contracts/tests/test_token_network.py
pirapira: That kind of sophisticated things will be tracked in https://github.com/raiden-network/raiden-contracts/issues/599#issuecomment-466200340 next week.

This PR is really for getting `raiden` developing going.

raiden_contracts/constants.py
pirapira: I'm intentionally repeating this.  The hidden message is "This is something as ad-hoc as `0.3._`.  Don't take it seriously (or deploy it on the mainnet)".

pirapira:
Etherscan verification is done.

If you merge this give me a call, then I can use the phone internet on the train and kick off the PyPI packaging.

raiden_contracts/constants.py
rakanalh: This convention won't relay the message. Also there is no technical argument for having this version numbering scheme so i would go for using 0.8.0

pirapira: No that's too confusing because `0.8.0` looks normal.  The version in this PR is very different from other versions (not deployable on the mainnet).  It almost deserves a separate PyPI package `raiden-contract-for-testing-only`.

If you want a normal version scheme, wait until [the proper solution](https://github.com/raiden-network/raiden-contracts/issues/599#issuecomment-466200340) lands.  It's up to you.

pirapira: @rakanalh But tell me your choice, if I should push this PR or go implement the proper solution ^.  Currently I'm pushing this because this is the ASAP solution that you wanted.

pirapira: I'm accommodating the abnormal situation, where the bytecode for the mainnet is not tested, but something else is being tested.  And I chose to provide something else here ASAP.  But the unlimited version is something that does not really belong the Ithaca version of `raiden-contracts`.  My opinions aside, I'm taking extra steps because keeping the CI running in `raiden` repo is important on its own.

pirapira: @rakanalh if it's just about names.  Instead of `0.8._` I can also say `0.8-unlimited-only-for-testing`.  Is this fine? I try.

pirapira: No, sorry the train leaves in 15min.  And I cannot change the version string, compile the sources, deploy, and talk to Etherscan.  I give up this PR as it is now, for this week.

pirapira: On my side, `0.8._` does not need to convey the meaning.  It just needs to look weird, and cause questions like "what does it mean"?

LefterisJP: @pirapira In the end it's your choice. Will there ever be a `0.8.0`? Or using the underscore effectively uses up `0.8.0` too?  As described above `0.3._` was just something we did back then. I don't even remember the exact reason so blindly copying it sounds dirty.

If you want the `patch` part of the version being an `_` meaning that the version will be unlimited then that's also your choice. But this should be documented and be offered as part of the interface between raiden contracts and client since we will probably be stuck with `0.8._` for a while.

I like the approach you linked above as the [proper way](https://github.com/raiden-network/raiden-contracts/issues/599#issuecomment-466200340).

How much time would that take? @rakanalh can you wait until that? If not and this satisfies the requirements to get your PR green then no need to analyze it further, just merge it.

pirapira: I think the solution in this PR is dirty, and it deserves a dirty name.

I documented `0.8._` in README.

There will not be `0.8.0`.  The next change will be putting back the limit, and that chages the ABI, so it ill be `0.9.0`.

The proper way, I estimate till Tuesday next week (very optimistic, but achievable if I cut all meetings and any distractions).

rakanalh: I would wait for the "proper" way as well.

pirapira: That's a sane approach!

pirapira:
This PR was an offer I could make for this week (within the travel constraints), but I'm glad I'm given more time to prepare something decent. https://github.com/raiden-network/raiden-contracts/pull/600#discussion_r259372221

PR #607, opened 2019-02-25
Auto limit removal

pirapira:
For #599, this PR implements a simple tool to remove limits from the existing Solidity sources.  The generated sources will be compiled in the next PRs.


README.rst
```diff
@@ -100,6 +100,13 @@ Development
 If you want to test and further develop outside the officially provided source code, compiled files and deployed addresses, you can do it at your own risk.
 
 
+If you want to produce Solidity sources without deposit limits and deprecation switches::
+
+    make remove_limits
+
+then such souces will appear in ``raiden_contracts/contracts-without-limits``.
```
BOR4: "sources"

pirapira: Will fix in this PR.  Thanks.

pirapira:
@hackaugusto @rakanalh and I had a quick chat.

* deployed sources will be visible in `deploy_x.y.z/` directories next to `deployment_***.json`
* in these deployed sources, the TAGS `/* LIMITED_VERSION_START */` and `/* LIMITD_VERSION_END */` will not be visible.

pirapira:
I'll first create a PR that adds the sources next the deployment data.  After that I come back here.

raiden_contracts/contracts/raiden/TokenNetwork.sol
```diff
@@ -31,23 +31,27 @@ contract TokenNetwork is Utils {
         115792089237316195423570985008687907853269984665640564039457584007913129639935
     );
 
+    /*LIMITED-VERSION-START*/
```
LefterisJP: Why not change this to something like `TokenNetwork_template.sol` and then use the tool to generate two different files? This could be a bit cleaner and we could have the code without those ugly "preprocessor" markers.

raiden_contracts/utils/remove_limits.py
```diff
@@ -0,0 +1,62 @@
...
+
+import click
+from pathlib import Path
+import subprocess
+import re
+
+from raiden_contracts.constants import CONTRACTS_VERSION, CONTRACTS_VERSION_LIMITED
+
+"""
+A utility to copy a directory containing Solidity files and
```
LefterisJP: nitpick: /s/A/An

rakanalh: I don't see it as an "an". "A utility" sounds valid for me.

raiden_contracts/contracts/raiden/TokenNetwork.sol
pirapira: That's the way to go!

raiden_contracts/utils/remove_limits.py
LefterisJP: It's not a matter of how you see it or how it sounds to you. It's the language's grammar. Before a vowel `a` become `an`. https://www.englishforums.com/English/AUtilityVsAnUtility/hrmlw/post.htm

rakanalh: True... but doesn't "An utility" sound wrong to you?

pirapira: A vowel is a sound.  It's not a charecter.  `utility` sounds like `[yoo-til-i-tee]` so it takes an `a`.

LefterisJP: > A vowel is a sound

Vowell are letters. The English language has 5 of them. http://www.abcfastphonics.com/vowel-letters.html

But indeed the rule is for vowel letters and there are exception for some words starting with `u` depending on how they sound. So it is indeed a utility, just like a university. as you explained. "yoo-".

pirapira: Now Augusto sent me https://www.englishclub.com/pronunciation/a-an.htm

raiden_contracts/contracts/raiden/TokenNetwork.sol
pirapira: * http://mustache.github.io/
* http://jinja.pocoo.org/ 
* https://docs.python.org/3.4/library/string.html#template-strings doesn't seem to work for this usage.

raiden_contracts/utils/remove_limits.py
LefterisJP: Ha nice link with many examples :)

Issue #596, opened 2019-02-21
Make all deployed versions accessible


pirapira:
The current release procedure does not provide `data_x.y.z` directory for every contract version.  This issue keeps track of modifying the release procedure so that every package release (with a new deployment) adds a new `data_x.y.z` directory.

Of course, no new directories are needed when there have been no changes in the contract sources since the last package release.

https://github.com/raiden-network/raiden-contracts/issues/584#issuecomment-465924523

pirapira:
After #614, all future deployments will be accessible.

Is there any need to recover older versions like `0.7.0`?

pirapira:
@hackaugusto from 0.8.0 onward, all deployment data will be accessible forever.  Do you also want ancient versions recovered?

hackaugusto:
> Do you also want ancient versions recovered?

Only for the versions that we have deployed to the main chain and compatibility will be preserved.

pirapira:
Then I have everything covered I believe.

PR #608, opened 2019-02-25
Make sure that once a contract version works, it works forever

### LINKED ISSUE #606: Make sure contract version 0.6.0 does not work on any versions
@rakanalh says, with package version v0.10.1, contract_version `0.6.0` works. https://github.com/raiden-network/raiden-contracts/pull/594#discussion_r259763329  That's not intended.

This issue keeps track of determining if that's true.  And if yes, investigate why this happens and make sure that usage results in error.

pirapira:
Before this commit, when the user specified the latest version
explicitly, raiden-contracts automatically chose the latest 'data'.

This commit removes that behavior.  If the user wants to use the latest
version (without caring what version it is), the user should say
`precompiled_data_path(None)` with `None` as the contract version.

This closes #606.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Documents changes that result in:
 
 - Add data/gas.json that contains gas measurements on the development version.
 - Move Raiden contracts to "raiden" subdir, so that the imports match the directory layout.
+- Remove a shortcut `contracts_precompiled_path(version)` when `version` is the current development version. The same path is accessible with `contracts_precompiled_path(None)`.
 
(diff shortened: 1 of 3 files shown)


pirapira:
@rakanalh was right about the behavior. Whether a contract version worked or not depended on the package version. This PR removes such dependencies.  Once a contract version works, it should work after that.

pirapira:
I'm kind-of satisfied just by figuring out what was going on.  This part of the document will change anyway with https://github.com/raiden-network/raiden-contracts/issues/596.

PR #615, opened 2019-02-25
Improve MSC tests

### LINKED ISSUE #56: Add tests for MS smart contract
Currently we only have one test that tests the happy case. 

More tests must be added to test edge cases and scenarios where the contract should fail.

karlb:
* Extract basic setup into fixture
* Split single test into multiple tests
* Provide test contract to access private function
* Include checks for error cases in tests

Closes #56


pirapira:
Ran `make compile_contracts` and pushed the result.

pirapira:
Travis build has succeeded but GitHub hasn't noticed that.

PR #614, opened 2019-02-25
v0.12.0: sources, unlimited versions, and all deployments

pirapira:
Sometimes it's hard to find the Solidity source corresponding to a particular deployment. https://github.com/raiden-network/raiden-contracts/issues/556  This PR makes the process easier by copying the sources to `data` directory whenever they are compiled.

With this change, the precompiled bytecode is always accompanied by the source it's compiled from.

PR #617, opened 2019-02-26
Use Mustache to generate limited and unlimited versions

pirapira:
Also `make compile_contracts` and `make verify_contracts` work on both versions.


pirapira:
@LefterisJP tests passed.  Ready for review.

pirapira:
But I noticed a mistake, and am fixing it.

pirapira:
Final push.  Tried the release sequence until `etherscan_verify` locally.

Makefile
```diff
@@ -1,9 +1,12 @@
 all: verify_contracts install
 
-compile_contracts:
+solve_templates:
```
rakanalh: I would suggest something like `render_contracts` or `generate_contracts` from the template

pirapira: `render_contracts` sounds good. I'll change the names in this PR.

raiden_contracts/contract_manager.py
```diff
@@ -93,6 +95,8 @@ def _compile_all_contracts(self) -> None:
         old_working_dir = Path.cwd()
         chdir(_BASE)
 
+        subprocess.run(["make", "-C", "..", "solve_templates"])
```
rakanalh: The pypi package will **not** & IMO should **not** contain the makefile, therefore this would actually break.

I think generating the contracts out of templates shouldn't be done in the contract manager at all... the contracts should be generated manually and then packaged into the pypi package (excluding the templates)

pirapira: Oh I see.  I didn't think about that.

pirapira: Now I've removed these `make` calls.

The sources will be added in the package in the next PR: https://github.com/raiden-network/raiden-contracts/pull/614/files#diff-97c91a104c431d0c365565d3ac03ac13R12 .  I'm not adding templates.

pirapira:
@rakanalh @LefterisJP this one is ready again.

setup.py
```diff
@@ -51,9 +52,57 @@ def run(self):
...
+    if dst.exists() and dst.stat().st_mtime > src.stat().st_mtime:
+        # the destination is newer. No need to do anything.
+        return
+    with src.open(mode='r') as src_file:
+        content = src_file.read()
+    with dst.open(mode='w') as dst_file:
+        dst_file.write(pystache.render(content, mustache_hash))
+
+
+def render_templates(mustache_hash, src, dst):  # TODO: there has to be a template thing
```
rakanalh: >  `# TODO: there has to be a template thing`

Now we use mustache :smile:

pirapira: I'll remove the comment in #614.

Issue #623, opened 2019-02-26
deployment script should make sure limited<-->unlimited confusion is not happening


pirapira:
Currently, the deployment script takes `--flavor limited` or `--flavor unlimited` flags.  When the script interacts with an already deployed smart contract, the script should call `contract_version()` of the contract and see if it's interacting with the correct flavor.

When the script is called with `--flavor limited` but is talking to a `0.8.0_unlimited` version, something is wrong.  The script should detect this situation and fail.

pirapira:
The limited vs. unlimited flavors are going away.

PR #635, opened 2019-02-28
Fix version numbers caused by Mustache typo

pirapira:
I was implementing tests for https://github.com/raiden-network/raiden-contracts/issues/623 and realized I made a typo in the Mustache configuration.

In the templates I had `{{version_suffix}}` with `_`. I was telling Mustache to replace `{{version-suffix}}` (with `-`). 

This PR
* fixes the typo
* starts using Mustache in the stricter mode so similar mistakes will cause build errors
* forces template rendering whenever the rendering script is newer than the target.


pirapira:
Waiting for CI now...

pirapira:
And now the whole codebase has realized that sometimes it's `0.8.0` and sometimes it's `0.8.0_unlimited`.

PR #651, opened 2019-03-01
Remove flavors (Part 2)

pirapira:
This PR removes the flavors (unlimited / limited).  Because we decided to maintain just one Solidity source that fits both uses.

Review #647 first.


pirapira:
I break rules because this portion is now the bottleneck.

Issue #642, opened 2019-03-01
Use pylint to detect more style problems


pirapira:
Travis CI runs `flake8` on `raiden-contracts` but leaves some suspicious code, as pointed out by @LefterisJP. https://github.com/raiden-network/raiden-contracts/pull/641#discussion_r261433593

This issue keeps track of making sure that `raiden-contracts` uses the same flake8 configuration as `raiden`.

pirapira:
I guess running `pylint` too would help.

pirapira:
And don't forget to get `.pylint.rc` from `raiden` repo.

PR #653, opened 2019-03-01
Remove some pylint warnings

pirapira:
Please review #652 first.

This is a first shot at using #642 but removes only some of the warnings (because there are over 2,000 lines of warnings).  After this PR, probably we should use the pylint configuration from `raiden` repo.


pirapira:
Breaking rules for being at the bottleneck.

PR #655, opened 2019-03-02
Make deposit limits configurable at TokenNetowk

pirapira:
deployment time.

After some trial and error, I believe that the best approach to accommodate the different use cases of TokenNetwork is to make it configurable at the deployment time, as opposed to providing multiple flavors or versions of the source code.


pirapira:
I'm breaking rules because this portion is now at the bottleneck.

PR #657, opened 2019-03-04
v0.13.0 release

pirapira:
Followed the release process for v0.13.0.  The main reason for making a release now is to make sure that the current contract interfaces work with other components of raiden.

PR #658, opened 2019-03-04
Remove unnecessary data directories within data

pirapira:
directories.

Thanks @BOR4 for pointing this out: https://github.com/raiden-network/raiden-contracts/pull/657#issuecomment-469171013


pirapira:
The same version will be tested for the release run soon.

PR #662, opened 2019-03-04
Allow fixtures to be reused across multiple tests

karlb:
This only reduces the runtime by ~25%, but further speed increases can be gained by updating the scope of more fixtures.

I've also tried to reduce the number of fixtures to keep the test setup complexity down a bit. See the single commit messages for more details.


pirapira:
This is wonderful.

pirapira:
I rebased this on `master`.

PR #665, opened 2019-03-04
Mark slow tests

karlb:
This allows skipping tests with `pytest -m 'not slow'`. Now, I can get a
run of all non-slow tests down to 90s in my machine (also using `-n 4`).

### DIFF
--- a/raiden_contracts/tests/test_channel_settle_unlock_state.py
+++ b/raiden_contracts/tests/test_channel_settle_unlock_state.py
@@ -259,6 +259,7 @@ def test_settlement_outcome(
     return f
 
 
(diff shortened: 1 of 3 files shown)


karlb:
Well, I only marked those taking longer than 10 seconds. That does not mean that I consider the others to be fast. :snail:

BOR4:
marking tests with slow will require constant updating. May I recommend you put timeout on other tests and once in a run that timeout is hit you mark that test as well with slow.

pirapira:
@BOR4 that's an interesting idea.  I created an issue about that https://github.com/raiden-network/raiden-contracts/issues/666 .

PR #678, opened 2019-03-05
Add a whole balance limit to UserDeposit

### LINKED ISSUE #585: Add deposit limit to UserDeposit
This issue keeps track of adding a deposit limit to the UserDeposit.  Until some concrete number is decided (https://github.com/raiden-network/team/issues/333), we can set up an arbitrary number (something like 12345678 looks temporary.)

pirapira:
which is configurable at the deployment time.

Currently we keep the option of limiting deposits in all smart contracts (https://github.com/raiden-network/team/issues/324#issuecomment-465100567).

I [chose](https://github.com/raiden-network/raiden-contracts/issues/585#issuecomment-469731114) against using the contract's token balance because that poses a DOS attack possibility.

This closes #585.


karlb:
I'm not sure which requirements we have for the choice of our limits. Having a per-account limit would be simpler and more gas efficient than having a global limit. But I assume you know that and did it this way for a good reason.

raiden_contracts/tests/fixtures/token.py
```diff
@@ -4,11 +4,19 @@
     CONTRACT_CUSTOM_TOKEN,
 )
 
-token_args = (10 ** 26, 18, CONTRACT_CUSTOM_TOKEN, 'TKN')
+
+@pytest.fixture(scope='session')
+def custom_token_total_supply():
```
karlb: Why not use plain variables instead of fixtures for constants?

raiden_contracts/tests/fixtures/user_deposit.py
```diff
@@ -3,15 +3,22 @@
...
+    return custom_token_total_supply // 100
+
+
 @pytest.fixture(scope='session')
 def uninitialized_user_deposit_contract(
         deploy_tester_contract,
         custom_token,
+        user_deposit_whole_balance_limit,
 ):
+    print(user_deposit_whole_balance_limit)
```
karlb: Is this `print` here intentionally, or is it left over from debugging?

pirapira: Indeed it's a leftover from debugging. I'll remove this in this PR.

raiden_contracts/tests/fixtures/token.py
pirapira: I'll try making this a constant in this PR.

pirapira:
@karlb the reason for the total limit can be seen here: https://github.com/raiden-network/team/issues/333

In general, one person can create an unlimited number of Ethereum accounts, so usually per-account limitations are not very effective.

PR #682, opened 2019-03-06
Moving functions from deploy script to a separate module

### LINKED ISSUE #681: Moving functions from `deploy/__main__.py` to somewhere else, might increase coverage
Currently Codecov doesn't recognize any line coverage in `deploy/__main__.py`.  This might be due to the special naming `__main__`.  Perhaps, if we move some functions from `__main__.py` to somewhere else, the measured coverage increases.

pirapira:
I guess Codecov works better this way.  This closes #681 .


pirapira:
It does not change the coverage much.

PR #683, opened 2019-03-06
Check final balances in test_settle_no_bp_success

pirapira:
Because in this case, both participants should get the original deposits back.


pirapira:
Thank you.

PR #688, opened 2019-03-06
Run and enable isort

pirapira:
This PR enables `isort` in CI runs.  Many imports had to be reordered.


pirapira:
Thank you, again.

PR #689, opened 2019-03-06
Transfer after all storage modifications

pirapira:
[The common wisdom](https://solidity.readthedocs.io/en/latest/security-considerations.html#use-the-checks-effects-interactions-pattern)
goes like, do all storage updates before talking to other contracts.

I don't see any reason not to follow that here.


pirapira:
Thank you!

PR #691, opened 2019-03-06
Test check_successful_tx

pirapira:
The function `raiden_contracts.utils.check_successful_tx` had no unit tests.  I'm adding them.

### DIFF
--- a/raiden_contracts/tests/test_transaction.py
+++ b/raiden_contracts/tests/test_transaction.py
@@ -7,9 +7,7 @@ from raiden_contracts.utils.transaction import check_successful_tx
 
 def test_check_successful_tx_with_status_zero():
     web3_mock = Mock()
-    web3_mock.eth = Mock()
-    web3_mock.eth.getTransactionReceipt = Mock(return_value={'blockNumber': 300, 'status': 0})
-    web3_mock.eth.getTransaction = Mock()
+    web3_mock.eth.getTransactionReceipt.return_value = {'blockNumber': 300, 'status': 0}
     txid = 'abcdef'
     with pytest.raises(ValueError):
         check_successful_tx(web3=web3_mock, txid=txid)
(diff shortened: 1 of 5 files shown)


raiden_contracts/tests/test_transaction.py
```diff
@@ -0,0 +1,52 @@
...
+
+import pytest
+
+from raiden_contracts.utils.transaction import check_successful_tx
+
+
+def test_check_successful_tx_with_status_zero():
+    web3_mock = Mock()
+    web3_mock.eth = Mock()
+    web3_mock.eth.getTransactionReceipt = Mock(return_value={'blockNumber': 300, 'status': 0})
```
karlb: This can also be written as
```
web3_mock.eth.getTransactionReceipt.return_value = {'blockNumber': 300, 'status': 0}
```
which is a bit nicer IMO.

raiden_contracts/tests/test_transaction.py
```diff
@@ -0,0 +1,52 @@
+from unittest.mock import Mock
+
+import pytest
+
+from raiden_contracts.utils.transaction import check_successful_tx
+
+
+def test_check_successful_tx_with_status_zero():
+    web3_mock = Mock()
+    web3_mock.eth = Mock()
```
karlb: No need for this line. When accessing an attribute of a `Mock`, another mock is created.

raiden_contracts/tests/test_transaction.py
```diff
@@ -0,0 +1,52 @@
...
+    web3_mock.eth.getTransactionReceipt = Mock(return_value={'blockNumber': 300, 'status': 0})
+    web3_mock.eth.getTransaction = Mock()
```
karlb: As above, I don't think this line is necessary.

raiden_contracts/tests/test_transaction.py
pirapira: Will do this in this PR.

raiden_contracts/tests/test_transaction.py
pirapira: That worked.

raiden_contracts/tests/test_transaction.py
pirapira: That worked.

PR #696, opened 2019-03-07
Stop automatically converting version=None

pirapira:
into the current contracts version.

During the release process, the version number in contracts.json will be changed by hand.

### DIFF
--- a/raiden_contracts/tests/test_contracts_compilation.py
+++ b/raiden_contracts/tests/test_contracts_compilation.py
@@ -96,12 +96,6 @@ def test_verification_contracts_checksums():
         manager.verify_precompiled_checksums(contracts_precompiled_path())
 
 
-def test_contracts_version():
-    """ Check the value of contracts_version """
-    manager = ContractManager(contracts_precompiled_path())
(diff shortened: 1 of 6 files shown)


raiden_contracts/tests/test_contracts_compilation.py
```diff
@@ -232,3 +226,16 @@ def test_contract_manager_json(tmpdir):
...
     contract_manager_meta(precompiled_path)
+
+
+def test_contract_manager_constructor_does_not_invent_version():
+    """ ContractManager should not invent a version string """
+    manager = ContractManager(contracts_precompiled_path(version=None))
+    assert manager.contracts_version is None
+
+
+@pytest.mark.parametrize('version', [CONTRACTS_VERSION, '0.3._', '0.4.0'])
```
rakanalh: Could you add `0.9.0` to the list? just to make sure it also works

pirapira: Will do in this PR.

pirapira:
If it breaks I'll hold both pieces.

PR #701, opened 2019-03-07
Split ContractSourceManager from ContractManager

### LINKED ISSUE #695: Split ContractManager into two classes ContractSourceManager and PrecompiledContractManager
Currently, `ContractManager` object can be created with a source directory or a precompiled json document.  This issue keeps track of creating two separate classes for these two kinds of objects.

Perhaps they can share some methods with inheritance and so.

pirapira:
Before this PR, there were two ways to create a ContractManager object, with sources and with precompiled json file.  This PR splits the two uses into two different classes.

This closes #695.


pirapira:
Codecov is not happy.  Writing tests.

PR #702, opened 2019-03-07
An easier way to modify a string

pirapira:
This PR proposes a simpler way to mutate a string.


pirapira:
Thanks for solving the puzzle :)

PR #722, opened 2019-03-11
0.14.0 release

pirapira:
Making a release now to provide the new interface:  [`Removed checks in MonitoringService.canMonitor() until the design is determined.`](https://github.com/raiden-network/raiden-contracts/compare/master...pirapira:0.14.0?expand=1#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR12)


pirapira:
Codecov complains because no CI tests touch `setup.py`.  I think `setup.py` should not be counted  https://github.com/raiden-network/raiden-contracts/issues/724

raiden_contracts/data/gas.json
```diff
@@ -1,16 +1,16 @@
 {
     "EndpointRegistry.registerEndpoint": 48408,
     "MonitoringService.claimReward": 42679,
-    "MonitoringService.monitor": 203858,
-    "OneToN.claim": 90914,
+    "MonitoringService.monitor": 201313,
+    "OneToN.claim": 90897,
```
palango: I remember that there was an issue with raiden when these were lowered, but when switching to the new contracts completely this shouldn't be a problem.

raiden_contracts/data_0.10.0/gas.json
```diff
@@ -0,0 +1,21 @@
+{
+    "EndpointRegistry.registerEndpoint": 48408,
+    "MonitoringService.claimReward": 42679,
+    "MonitoringService.monitor": 201313,
+    "OneToN.claim": 90897,
+    "SecretRegistry.registerSecret": 45757,
+    "TokenNetwork DEPLOYMENT": 4031160,
+    "TokenNetwork.closeChannel": 111226,
```
palango: Ah, my bad above. We have a file per version now.

raiden_contracts/data/gas.json
pirapira: When the last problem happened, the numbers lived in a global `raiden_contracts/constants.py`.  That set of numbers affected all versions. Now the situation should be better. `gas.json` lives in `data_x.y.z` so the numbers are local to a particular version.

But I have to try the new way to see if any problem happens.

raiden_contracts/data_0.10.0/gas.json
pirapira: Ah, I replied before seeing this.

pirapira: One disadvantage of "everything per version" is a big release PR like this one.

Issue #715, opened 2019-03-11
Function handle_log has a Cognitive Complexity of 12 (exceeds 5 allowed). Consider refactoring.


pirapira:
says CodeClimate.

pirapira:
Still, "Function handle_log has a Cognitive Complexity of 12 (exceeds 5 allowed)."

pirapira:
This has been done.

PR #727, opened 2019-03-11
Simplify handle_log

pirapira:
CodeClimate says handle_log is too complicated #715.
This commit tries to make the function simpler.

I saw some numbered index access `[0]` and `[1]` that could be named.  So I introduced a named tuple.

### DIFF
--- a/raiden_contracts/utils/logs.py
+++ b/raiden_contracts/utils/logs.py
@@ -1,5 +1,5 @@
 import functools
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 from inspect import getframeinfo, stack
 from typing import Dict, List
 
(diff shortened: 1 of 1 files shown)


raiden_contracts/utils/logs.py
```diff
@@ -167,3 +171,7 @@ def uninstall(self):
         assert self.filter is not None
         self.web3.eth.uninstallFilter(self.filter.filter_id)
         self.filter = None
+
+
+# A concrete event added in a transaction.
+LogRecorded = namedtuple('LogRecorded', 'message callback count')
```
palango: Maybe move this to the top, this is hard to find otherwise

pirapira: Yes, doing in https://github.com/raiden-network/raiden-contracts/pull/759

PR #728, opened 2019-03-11
Deployment script by default writes to `data` not `data_x.y.z`.

### LINKED ISSUE #723: deployment scripts should work on `data/` not on `data_x.y.z` by default
Currently, when `deploy/__main__.py` script gets no options about the contract versions, it tries to store the deployment information in `raiden_contracts/data_x.y.z` where `x.y.z` is the current contract version.  Due to this behavior, currently, the release procedure cannot be followed literally.

The current deployment procedure is based around the idea:
* do everything in `data` and at the end copy `data` into `data_x.y.z` (and maybe adjust the `contracts_version`)

pirapira:
This closes #723.

I made the first release after #696, and I noticed a glitch.  #696 unintentionally introduced a change.  The deployment script tries to save `deployment*.json` in `data_0.10.0` not `data`.  Every other script currently operates on `data`.  I had to move `data` to `data_0.10.0` and create a symbolic link from `data` to `dta_0.10.0` temporarily during the release process.

This PR moves the default destination back to `data` (with `--contracts-version 0.10.0`, the other behavior is still accessible).


pirapira:
Oh still codecov/patch is 60%.

pirapira:
Thank you!

PR #730, opened 2019-03-11
Simplify LogHander.wait()

pirapira:
CodeClimate says wait() is too complicated (#717). This commit tries to
simplify the method.


pirapira:
Thank you.  I opened an issue for replacing `print()`s.

PR #732, opened 2019-03-13
How to find addresses of deployed contracts

pirapira:
I'm trying to make it easier to figure out the addresses of the deployed contracts.


pirapira:
Thank you.

PR #739, opened 2019-03-13
ServiceRegistry() takes a token what token

### LINKED ISSUE #735: Rename and document _token_address in ServiceRegistry constructor
Maybe `_token_for_registration`.  Also add docstrings.

See https://github.com/raiden-network/raiden-contracts/issues/734#issuecomment-472411627 and the whole issue there.

pirapira:
a token for registration fees.

This closes #735


raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -11,11 +11,12 @@ contract ServiceRegistry is Utils {
...
 
-    constructor(address _token_address) public {
-        require(_token_address != address(0x0));
-        require(contractExists(_token_address));
+    // @param _token_for_registration The address of the ERC20 token contract that services use for registration fees
+    constructor(address _token_for_registration) public {
+        require(_token_for_registration != address(0x0));
+        require(contractExists(_token_for_registration));
 
-        token = Token(_token_address);
```
LefterisJP: Why is the code duplicated? Both here and in ` raiden_contracts/contracts/services/ServiceRegistry.sol` ?

pirapira: Just for historical reasons.  I guess `make compile_contracts` and so should just look at `data/source/...`, and the other can be removed.

pirapira: https://github.com/raiden-network/raiden-contracts/issues/740

LefterisJP: Yes please let's do that, maintaining duplicated contract code sounds like a nightmare.

PR #743, opened 2019-03-14
Quicker test_deployment_script

### LINKED ISSUE #736: Mock deployment data, instead of deploying on a private chain
In https://github.com/raiden-network/raiden-contracts/pull/728/files#diff-3df217e275c3585b92dc3b0ffa51623fR376, just for testing `store_and_verify_deployment_info_raiden` function, an actual deployment happens on a private chain. This is slow. This issue keeps track of setting up a mock of deployment data so the test runs faster.

This will knock 30 seconds off from the CI runs.

pirapira:
Running test_deploy_script took 142 seconds before this commit but now 64 seconds.

This closes #736.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_script:
   - make mypy
 
 script:
-  - pytest --cov=./ raiden_contracts/tests/ -n 2 --cov-config setup.cfg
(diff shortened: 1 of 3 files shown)


pirapira:
CodeCov is not happy because:

In `contract_manager.py:70`, `if version is None` is not covered anymore.

In `deploy/__main.py__:86`, older contract versions are not hit anymore.

I'll add tests to cover these in this PR.

PR #744, opened 2019-03-15
MS should be able to claim reward after settle period ends

LefterisJP:
As per our conversation with @palango and @Dominik1999 in the chat.

It makes no sense to claim a reward after the channel data has been removed. Instead we should only wait until the end of the settlement period.

@pirapira : My system only has solc v0.5.5 and as such I can't run `make compile contracts`. Got no time to compile solidity by hand for v0.5.4 now so I would like to see if someone else can do it. If not I will just have to bite the bullet at some point.

Also is this the file you are supposed to change to introduce changes?

### DIFF
--- a/raiden_contracts/data/source/services/MonitoringService.sol
+++ b/raiden_contracts/data/source/services/MonitoringService.sol
@@ -232,7 +232,7 @@ contract MonitoringService is Utils {
         ));
 
         // Only allowed to claim, if channel is out of the settlement period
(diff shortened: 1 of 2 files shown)


pirapira:
> It makes no sense to claim a reward after the channel data has been removed. Instead we should only wait until the end of the settlement period.

I agree.

> Also is this the file you are supposed to change to introduce changes?

Yes, since last week.

#747 supersedes this PR.

PR #747, opened 2019-03-18
MonitoringService update + contracts recompilation

palango:
Based on #744 

This contains an update of `solc` from 0.5.4 to 0.5.6 (because that's what I have installed locally).

Because the `eth_tester`  dependency doesn't seem fit for the petersburg EVM I force `solc` to output `byzantium` compatible code, but this should be fixed later.


raiden_contracts/contract_source_manager.py
```diff
@@ -55,6 +55,7 @@ def relativise(path):
                     output_values=PRECOMPILED_DATA_FIELDS + ['ast'],
                     import_remappings=import_dir_map,
                     optimize=False,
+                    evm_version='byzantium',
```
pirapira: I think `evm_version` should be recorded in `contracts.json`.

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -231,15 +231,14 @@ contract MonitoringService is Utils {
             token_network_address
         ));
 
-        // Only allowed to claim, if channel is settled
-        // Channel is settled if it's data has been deleted
-        TokenNetwork.ChannelState channel_state;
-        (, channel_state) = token_network.getChannelInfo(
+        // Only allowed to claim, if channel is out of the settlement period
+        uint256 settle_block_number;
+        (settle_block_number,) = token_network.getChannelInfo(
```
pirapira: This looks tricky.  `settle_block_number` is sometimes 0, and sometimes the **length** of the settlement period.  I think both `channel_state` and `settle_block_number` are necessary.

pirapira:
You can also tell me to implement it the way I like.

raiden_contracts/data/source/services/MonitoringService.sol
palango: Makes sense, we should at least check that the channel_state is closed.

pirapira: That should work.

raiden_contracts/contract_source_manager.py
palango: How do I do this?

pirapira: * change `contract_manager.py` to read/write the new field
* set up a default value for `evm_version`.

Maybe I should do it later.   When I modify `contract_manager.py`,  I can also manually add the field to the existing `contracts.json`s with the new solc versions.

pirapira:
#749 took over the change about the MS rewards. #748 will keep track of the Solidity upgrade.

PR #749, opened 2019-03-19
Earlier Monitoring Rewards

pirapira:
After this PR, monitoring services can claim rewards as soon as the settlement period finishes.  Before this PR, they had to wait until all locked amounts are claimed.

See #744 and #747 for how this change came about.


pirapira:
The test output is messy since I introduced codecov (I changed `coverage` into `pytest` and so on, making trial-and-error).

Issue #621, opened 2019-02-26
rename deploy.main register's --registry option into --token-network-registry


pirapira:
`raiden-contracts` codebase now contains many registries.  EndpointRegistry, SecretRegistry and ServiceRegistry.  So just calling an option `--registry` is no longer informative enough.

PR #755, opened 2019-03-20
Rename --registry into --token-network-registry

pirapira:
This closes https://github.com/raiden-network/raiden-contracts/issues/621 .

### DIFF
--- a/raiden_contracts/deploy/__main__.py
+++ b/raiden_contracts/deploy/__main__.py
@@ -553,7 +553,8 @@ def token(
     '--registry-address',
     default=None,
     callback=error_removed_option,
-    help='removed',
+    hidden=True,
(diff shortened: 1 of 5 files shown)


pirapira:
I've given up satisfying `codecov/patch`.

raiden_contracts/deploy/__main__.py
```diff
@@ -548,6 +552,12 @@ def token(
 @click.option(
     '--registry-address',
     default=None,
+    callback=error_removed_option,
+    help='removed',
```
karlb: I think using `hidden=True` instead of providing a help text would be nicer for removed commands.

pirapira: I go for `hidden=True`.  Thanks.

Issue #714, opened 2019-03-11
File __main__.py has 550 lines of code (exceeds 250 allowed). Consider refactoring. 


pirapira:
says CodeClimate.

pirapira:
Still 870 lines.

pirapira:
Still 660 lines or so.

pirapira:
Still 400 lines or so.

pirapira:
Before continuing this, I need to figure out the command chaining of `click`.  The current code seems to implement the command chaining, but I have never used it.  So I won't notice if some refactoring breaks the command chaining.

PR #762, opened 2019-03-21
Split verifyer away from deployer

pirapira:
This PR purely shuffles code around.

The original motivation was to reduce the number of lines in `raiden_contracts/deployer/_main_.py` 
https://github.com/raiden-network/raiden-contracts/issues/714 .

* I moved some code from `_main_.py` into `contract_deployer.py`.
* Then I had to separate `contract_verifyer.py` away because sometimes we don't have enough information to deploy but just want to verify the existing deployement data.


pirapira:
Thank you!

PR #772, opened 2019-03-25
Make a suggestion when failing for obsolete option

### LINKED ISSUE #765: Suggest the new option --token-network-registry-address when refusing --registry-address
When `deploy register` is called with `--registry-address`, the script shows
```
Error: no such option: --registry-address is no longer a valid option
```

The script should also show the new option `--token-network-registry-address`.

pirapira:
When `deploy register` fails for the obsolete option `--registry-option`
the script should make a suggestion to use `--token-network-registry`.

This fixes #765.


pirapira:
Thank you!

PR #773, opened 2019-03-25
Remove contracts_template because stopped Mustache

### LINKED ISSUE #756: Remove raiden_contracts/contracts_template
That's a leftover from the time when Mustache was tried.

pirapira:
a while ago.

This fixes #756.


rakanalh:
Whats the alternative implementation you're going for?

pirapira:
@rakanalh instead of template-filling, I use constructor arguments.

Instead of filling a template like
```
{{#if limit}}
uint256 some_limit;
{{#endif}}
```
I let the constructor choose `some_limit`.  This has been already implemented.  This PR removes a leftover that nobody uses.

Issue #784, opened 2019-03-27
TypeError: __init__() got an unexpected keyword argument 'hidden'


pirapira:
```
python -m raiden_contracts.deploy raiden --rpc-provider http://127.0.0.1:8545  --private-key /home/yoichi/xud-eth/blkchain1/keystore/UTC--2019-03-26T15-46-54.573149705Z--b772b74bc6021aab88ef154d76eb3174258e3865 --gas-price 10 --gas-limit 6000000  --contracts-version 0.3._
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/yoichi/.virtualenvs/offerm-script/lib/python3.7/site-packages/raiden_contracts/deploy/__main__.py", line 406, in <module>
    @click.pass_context
  File "/home/yoichi/.virtualenvs/offerm-script/lib/python3.7/site-packages/click/decorators.py", line 170, in decorator
    _param_memo(f, OptionClass(param_decls, **attrs))
  File "/home/yoichi/.virtualenvs/offerm-script/lib/python3.7/site-packages/click/core.py", line 1460, in __init__
    Parameter.__init__(self, param_decls, type=type, **attrs)
TypeError: __init__() got an unexpected keyword argument 'hidden'
getting tokennetworkregistry address
```

pirapira:
This was on `raiden-contracts 0.16.0`.

pirapira:
And click-6.7.

pirapira:
When I install `click==7.0` this problem disappears.

PR #785, opened 2019-03-27
click must be at least 7.0

pirapira:
because we use `hidden` options.

This commit fixes https://github.com/raiden-network/raiden-contracts/issues/784


pirapira:
Thank you!

PR #798, opened 2019-04-01
Remove get_contracts_deployed()

### LINKED ISSUE #671: get_contracts_deployed() sounds too active; should be fetch_deployment_data()
"get contracts deployed" sounds to me like sending a transaction to deploy contracts.  The function does not do that.  Instead it just finds the deployment data in the filesystem.

pirapira:
because that is no longer used out of the codebase.

Somehow I saw additional mypy errors so I had to fix them.  This closes #671 .

### DIFF
--- a/raiden_contracts/contract_manager.py
+++ b/raiden_contracts/contract_manager.py
@@ -1,7 +1,6 @@
 """ContractManager knows binaries and ABI of contracts."""
 import json
 from copy import deepcopy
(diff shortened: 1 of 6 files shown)


pirapira:
Perhaps I should split this PR into
1. remove all uses of `get_contracts_deployed()`
2. more mypy annotation
3. removing the definition of `get_contracts_deployed()`.

PR #800, opened 2019-04-01
Move DeployedModule enum to constants.py

pirapira:
Without doing this, I was unable to call `get_contracts_deployment_info()` from `contract_verifyer.py`.

I saw an error:
```
        if module not in DeploymentModule:
>           raise ValueError(f'Unknown module {module} given to get_contracts_deployment_info()')
E           ValueError: Unknown module DeploymentModule.RAIDEN given to get_contracts_deployment_info()
```
from `get_contracts_deployment_info()`.

It seemed like the Enum visible locally in contract_manager.py was different from the Enum imported into contract_verifyer.py .

I guess that's why many enums are defined in `constants.py`.`

### DIFF
--- a/raiden_contracts/constants.py
+++ b/raiden_contracts/constants.py
@@ -122,6 +122,13 @@ class OneToNEvent(str, Enum):
     CLAIMED = 'Claimed'
 
 
+class DeploymentModule(Enum):
+    """Groups of contracts that are deployed together"""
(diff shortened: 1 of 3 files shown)


pirapira:
This is the first split part of https://github.com/raiden-network/raiden-contracts/pull/798 .

pirapira:
A potential cargo cult is going on here.

PR #794, opened 2019-03-29
Skip gas estimation to speed up tests

karlb:
This currently causes weird side-effects like transactions that used to
fail with gas cost estimation suddenly succeeding (or at least not
throwing `TransactionFailed`).

@pirapira This a the problem I talked to you about. I have no idea why this happens. Maybe you can give this PR a try.

PR #802, opened 2019-04-01
Introduce transact_for_success()

pirapira:
which is similar to transact() but checks the receipt's status field.

Currently, we rely on the behavior that `transact()` raises `TransactionFailed` exception, but this is true only when we don't specify the amount of gas.  We'll be specifying the amount of gas (so that tests run faster) https://github.com/raiden-network/raiden-contracts/pull/794.


pirapira:
I'm curious how slower the CI test becomes.

pirapira:
The test is taking longer.  The original purpose of these changes was to make them faster.  Perhaps I should just introduce `call_and_transact`.

PR #803, opened 2019-04-01
Type-annotate get_contracts_deployment_info()

pirapira:
This PR introduces more `DeployedContracts` type annotations.

Somehow https://github.com/raiden-network/raiden-contracts/pull/798 woke up `mypy` and it started complaining about this.


pirapira:
Thank you!

Issue #807, opened 2019-04-02
get_contracts_deployment_info(module=ALL) on 0.4.0 should succeed


pirapira:
Currently it fails because the service contracts are missing.  This issue keeps track of fixing this on `master`, making a release and integrate the new version with `raiden`.

See https://github.com/raiden-network/raiden/issues/3723#issuecomment-478893706

pirapira:
This has been done.

PR #809, opened 2019-04-02
Don't try to find service contracts for RedEyes

pirapira:
and older deploymen data.

This is a part of https://github.com/raiden-network/raiden-contracts/issues/807

### DIFF
--- a/raiden_contracts/contract_manager.py
+++ b/raiden_contracts/contract_manager.py
@@ -3,6 +3,7 @@ import json
 from copy import deepcopy
 from json import JSONDecodeError
 from pathlib import Path
+from semver import compare
 from typing import Any, Dict, List, Optional
 
 from deprecated import deprecated
@@ -164,6 +165,16 @@ def merge_deployment_data(dict1: Dict, dict2: Dict) -> Dict:
     return result
 
 
+def version_provides_services(version: Optional[str]) -> bool:
+    if version is None:
+        return True
+    if version == '0.3._':
+        return False
+    if version == '0.8.0_unlimited':
+        return True
+    return compare(version, '0.8.0') > -1
+
+
 def get_contracts_deployment_info(
         chain_id: int,
         version: Optional[str] = None,
@@ -172,7 +183,8 @@ def get_contracts_deployment_info(
     """Reads the deployment data.
 
     Parameter:
-        module The name of the module. ALL means deployed contracts from all modules.
+        module The name of the module. ALL means deployed contracts from all modules that are
+        available for the version.
     """
     if module not in DeploymentModule:
         raise ValueError(f'Unknown module {module} given to get_contracts_deployment_info()')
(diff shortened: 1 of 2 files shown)


raiden_contracts/contract_manager.py
```diff
@@ -164,6 +165,16 @@ def merge_deployment_data(dict1: Dict, dict2: Dict) -> Dict:
...
 
 
+def version_provides_services(version: Optional[str]) -> bool:
+    if version is None:
+        return True
+    if version == '0.3._':
+        return False
+    if version == '0.8.0_unlimited':
+        return True
+    return compare(version, '0.8.0') > -1
```
LefterisJP: Hmm this is confusing. What exactly do you want to achieve with the last check.

Looking at the [docs](https://python-semver.readthedocs.io/en/latest/usage.html) of the semver module this will return true for versions >= 0.8.0.

If this is what you want to achieve then perhaps do `compare(version, '0.8.0') >= 0` since this way it will be a bit more obvious what is happening. The `-1` had me confused and I had to check the docs. And perhaps also add a docstring as not everyone knows how this compare function looks like.

raiden_contracts/contract_manager.py
```diff
@@ -186,7 +198,14 @@ def get_contracts_deployment_info(
...
         ))
 
-    if module == DeploymentModule.SERVICES or module == DeploymentModule.ALL:
+    if module == DeploymentModule.SERVICES and not version_provides_services(version):
+        raise ValueError(
+            f'SERVICES module queried for version {version}, but {version}'
+            'does not provide service contracts.',
+        )
+
+    if (module == DeploymentModule.SERVICES or module == DeploymentModule.ALL) and \
```
LefterisJP: style nitpick:

We never use `\` to prolong a line in raiden. Better use an extra variable for readability

```python
module_match = module == DeploymentModule.SERVICES or module == DeploymentModule.ALL
if module_match and version_provides_services(version):
    file.append(....
```

Also wouldn't the check `if version_provides_services(version)` be enough here?

pirapira: > wouldn't the check if version_provides_services(version) be enough here?

When people ask for `module=RAIDEN`, the service contracts should not be included in the result.

raiden_contracts/contract_manager.py
pirapira: I'll change to `>= 0`.  I agree it's easier.

raiden_contracts/contract_manager.py
```diff
@@ -186,7 +198,14 @@ def get_contracts_deployment_info(
             services=False,
         ))
 
-    if module == DeploymentModule.SERVICES or module == DeploymentModule.ALL:
+    if module == DeploymentModule.SERVICES and not version_provides_services(version):
+        raise ValueError(
+            f'SERVICES module queried for version {version}, but {version}'
```
pirapira: A space is missing.

raiden_contracts/contract_manager.py
LefterisJP: ah yes. My bad!

raiden_contracts/contract_manager.py
LefterisJP: Thanks man

raiden_contracts/contract_manager.py
```diff
@@ -189,9 +189,12 @@ def get_contracts_deployment_info(
     if module not in DeploymentModule:
         raise ValueError(f'Unknown module {module} given to get_contracts_deployment_info()')
 
+    def module_chosen(to_be_added: DeploymentModule):
```
LefterisJP: I will approve the PR anyway, just a small type nitpick. Add `-> bool:` return type.

pirapira: I defer this https://github.com/raiden-network/raiden-contracts/issues/811 and try to make a release now.

PR #812, opened 2019-04-02
0.17.1 release

pirapira:


LefterisJP:
@pirapira just to make sure. Now `get_contract_deployment_info` will return a contracts dict that may or may not contain services contracts, right? I am writing some startup tests in the raiden client at the moment and want to make sure I understand the interface correctly.

pirapira:
@LefterisJP 

>  Now get_contract_deployment_info will return a contracts dict that may or may not contain services contracts, right?

Correct, depending on the `version` argument.

pirapira:
codecov/patch is complaining about `setup.py`.  Maybe I should exclude `setup.py` from the coverage measurement.

PR #827, opened 2019-04-08
fix missing indent

czepluch:


pirapira:
Looks good to me.  I'll update `contracts.json` so Travis is happy.

czepluch:
Ah damn. Forgot to do that. Sorry! And thanks!

Issue #470, opened 2019-01-28
Make sure PyPI package contains the changelog


pirapira:
... and after the next release, confirm that the Changelog is included in the PyPI package in a nice way.

pirapira:
Done in #475.

pirapira:
Ah, no forgot "after the next release, confirm that the Changelog is included in the PyPI package in a nice way.".

pirapira:
No it doesn't.  Changelog has to live in `raiden_contracts` directory.

pirapira:
Isn't this already the case now?

pirapira:
The preferred approach is to have an outward symlink located within `raiden_contracts` that points to `CHANGELOG.md` at the top directory.

pirapira:
The PyPI package now contains a Changelog.

PR #829, opened 2019-04-08
Move CHANGELOG.md into raiden_contracts

pirapira:
so that it's included in the package.

This is a part of #470.


ulope:
The change does what it's supposed to but my question would by why you would want to include the changelog in the distribution package? I'd guess the likelihood of anyone ever seeing it in there is probably close to zero.

pirapira:
Now I don't remember who asked me to do this.  So I can close this as well.

pirapira:
At least @LefterisJP wants to see CHANGELOG.md in the package.

pirapira:
> At least @LefterisJP wants to see CHANGELOG.md in the package.

That was not quite right.  He just wants CHANGELOG somewhere.

pirapira:
GitHub does not follow symlinks automatically.  Maybe `distutils` does.  So the symlink should be placed in `raiden_contracts` package and point outwards.

pirapira:
I close this and follow up in #470.

PR #831, opened 2019-04-09
Deploy on Goerli chain

pirapira:
The aim of this PR is to provide the deployment information on the Goerli chain.  Since the Solidity sources changed a little bit, I made a new deployment on all test nets.


raiden_contracts/data/gas.json
```diff
@@ -1,17 +1,17 @@
 {
     "EndpointRegistry.registerEndpoint": 48408,
     "MonitoringService.claimReward": 42778,
-    "MonitoringService.monitor": 201347,
-    "OneToN.claim": 90897,
+    "MonitoringService.monitor": 201313,
```
LefterisJP: How come these are changing?

pirapira: I don't know.  I can do `git bisect`, but the changes are so small so I don't do it probably.

Anyway, I leave the change in.  Whenever I create a new `raiden_contracts/data_x.y.z` I leave the new measurement.

LefterisJP: Okay fair enough.

PR #840, opened 2019-04-09
Show a better error message before Byzantium

pirapira:
and when a transaction receipt does not contain 'status' field.

It's sometimes hard to connect a missing 'status' and the Byzantium hard fork.


pirapira:
I think I should write a test to hit this diff in this PR.

pirapira:
@nlsdfnbch I think you're right about the missing article and the kind of exception.  I'll fix them in this PR as well.

Issue #839, opened 2019-04-09
Concurrent run of test_print_gas and test_gas_json_has_enough_fields fails


pirapira:
Sometimes in Travis I see a failure of `test_gas_json_has_enough_fields`.  I guess it's because `test_print_gas()` and `test_gas_json_has_enough_fields()` access the same file.  This issue keeps track of making sure that such failures don't happen in Travis runs.

PR #842, opened 2019-04-11
Move register_token_network into ContractDeployer

pirapira:
talking to a TokenNetworkRegistor contract is quite similar to
deploying a contract. ContractDeployer is the natural place for
register_token_network() for now.

### DIFF
--- a/raiden_contracts/deploy/contract_deployer.py
+++ b/raiden_contracts/deploy/contract_deployer.py
@@ -223,15 +223,26 @@ class ContractDeployer(ContractVerifyer):
         """Register token with a TokenNetworkRegistry contract."""
         with_limits = contracts_version_expects_deposit_limits(self.contracts_version)
         if with_limits:
-            assert channel_participant_deposit_limit is not None, \
-                'contracts_version 0.9.0 and afterwards expect channel_participant_deposit_limit'
-            assert token_network_deposit_limit is not None, \
-                'contracts_version 0.9.0 and afterwards expect token_network_deposit_limit'
+            if channel_participant_deposit_limit is None:
+                raise ValueError(
+                    'contracts_version 0.9.0 and afterwards expect '
+                    'channel_participant_deposit_limit',
+                )
+            if token_network_deposit_limit is None:
+                raise ValueError(
+                    'contracts_version 0.9.0 and afterwards expect '
+                    'token_network_deposit_limit',
+                )
         else:
-            assert channel_participant_deposit_limit is None, \
-                'contracts_version below 0.9.0 does not expect channel_participant_deposit_limit'
-            assert token_network_deposit_limit is None, \
-                'contracts_version below 0.9.0 does not expect token_network_deposit_limit'
+            if channel_participant_deposit_limit:
+                raise ValueError(
+                    'contracts_version below 0.9.0 does not expect '
+                    'channel_participant_deposit_limit',
+                )
+            if token_network_deposit_limit:
+                raise ValueError(
+                    'contracts_version below 0.9.0 does not expect token_network_deposit_limit',
+                )
         token_network_registry = self.web3.eth.contract(
             abi=token_registry_abi,
             address=token_registry_address,
--- a/raiden_contracts/deploy/contract_deployer.py
+++ b/raiden_contracts/deploy/contract_deployer.py
@@ -268,7 +268,7 @@ class ContractDeployer(ContractVerifyer):
             token_address,
         ).call()
         token_network_address = to_checksum_address(token_network_address)
-        print(f'TokenNetwork address: {token_network_address}')
+        LOG.debug(f'TokenNetwork address: {token_network_address}')
         return token_network_address
 
     def deploy_service_contracts(
--- a/raiden_contracts/deploy/contract_deployer.py
+++ b/raiden_contracts/deploy/contract_deployer.py
@@ -249,8 +249,11 @@ class ContractDeployer(ContractVerifyer):
         )
 
         version_from_onchain = token_network_registry.functions.contract_version().call()
-        assert version_from_onchain == self.contract_manager.version_string(), \
-            f'got {version_from_onchain}, expected {self.contract_manager.version_string()}'
+        if version_from_onchain != self.contract_manager.version_string():
+            raise RuntimeError(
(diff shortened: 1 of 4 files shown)


pirapira:
I guess https://github.com/raiden-network/raiden-contracts/issues/839 is biting this PR.  I should fix that issue now.

pirapira:
Tests will turn green when #844 is merged into `master` and this PR is rebased.

raiden_contracts/deploy/contract_deployer.py
```diff
@@ -211,6 +212,56 @@ def _deploy_and_remember(
...
+        token_network_registry = self.web3.eth.contract(
+            abi=token_registry_abi,
+            address=token_registry_address,
+        )
+
+        version_from_onchain = token_network_registry.functions.contract_version().call()
+        assert version_from_onchain == self.contract_manager.version_string(), \
+            f'got {version_from_onchain}, expected {self.contract_manager.version_string()}'
+
+        command = token_network_registry.functions.createERC20TokenNetwork(
```
nlsdfnbch: rewriting this as a proper if/else would greatly improve readability.

raiden_contracts/deploy/contract_deployer.py
```diff
@@ -211,6 +212,56 @@ def _deploy_and_remember(
...
+        )
+        self.transact(command)
+
+        token_network_address = token_network_registry.functions.token_to_token_networks(
+            token_address,
+        ).call()
+        token_network_address = to_checksum_address(token_network_address)
+
+        print(
+            'TokenNetwork address: {0}'.format(
```
nlsdfnbch: Do you need to be backwards compatible with older python versions? an f-string would make it leaner, and you likely wouldn't have to break lines as much.

raiden_contracts/deploy/contract_deployer.py
```diff
@@ -211,6 +212,56 @@ def _deploy_and_remember(
...
+            token_registry_abi: Dict,
+            token_registry_address: str,
+            token_address: str,
+            channel_participant_deposit_limit: Optional[int],
+            token_network_deposit_limit: Optional[int],
+    ):
+        """Register token with a TokenNetworkRegistry contract."""
+        with_limits = contracts_version_expects_deposit_limits(self.contracts_version)
+        if with_limits:
+            assert channel_participant_deposit_limit is not None, \
```
nlsdfnbch: Are you aware that `assert` statements are ignored if python is run with optimizations? It's likely not going to be an issue in most scenarios, but I thought i'd better point it out in case these assertions are required in any case, not just when the `__debug__` variable is truthy.

raiden_contracts/deploy/contract_deployer.py
```diff
@@ -211,6 +212,56 @@ def _deploy_and_remember(
...
+            token_address,
+        )
+        self.transact(command)
+
+        token_network_address = token_network_registry.functions.token_to_token_networks(
+            token_address,
+        ).call()
+        token_network_address = to_checksum_address(token_network_address)
+
+        print(
```
nlsdfnbch: Should this be a `log` call?

raiden_contracts/deploy/contract_deployer.py
pirapira: I'll do this in this PR.

raiden_contracts/deploy/contract_deployer.py
pirapira: No need to care about old Python versions.  I'll change this into an f-string in this PR.

raiden_contracts/deploy/contract_deployer.py
pirapira: No, I was not aware of this.  This will turn into `if ...: raise ValueError(...)` in this PR.

raiden_contracts/deploy/contract_deployer.py
pirapira: Yes, I think so.  I'll change this in this PR.

raiden_contracts/deploy/contract_deployer.py
pirapira: I think the whole codebase needs this kind of cleanup, so I created an issue: https://github.com/raiden-network/raiden-contracts/issues/847

pirapira:
@nlsdfnbch  I applied all your suggestions.  A second look, please.

raiden_contracts/deploy/contract_deployer.py
```diff
@@ -211,6 +212,65 @@ def _deploy_and_remember(
...
+                raise ValueError(
+                    'contracts_version below 0.9.0 does not expect token_network_deposit_limit',
+                )
+        token_network_registry = self.web3.eth.contract(
+            abi=token_registry_abi,
+            address=token_registry_address,
+        )
+
+        version_from_onchain = token_network_registry.functions.contract_version().call()
+        assert version_from_onchain == self.contract_manager.version_string(), \
```
nlsdfnbch: leftover `assert`?

raiden_contracts/deploy/contract_deployer.py
```diff
@@ -211,6 +212,65 @@ def _deploy_and_remember(
...
+            f'got {version_from_onchain}, expected {self.contract_manager.version_string()}'
+
+        if with_limits:
+            command = token_network_registry.functions.createERC20TokenNetwork(
+                _token_address=token_address,
+                _channel_participant_deposit_limit=channel_participant_deposit_limit,
+                _token_network_deposit_limit=token_network_deposit_limit,
```
nlsdfnbch: Out of curiosity, what's with the leading underscore in the parameter name?

pirapira: Some Solidity coders use `_arg` for arguments.  For example https://www.ethereum.org/greeter#your-first-citizen-the-greeter .  I'm just keeping the convention.

pirapira:
Now I use `if` and `codecov` understands it!  I'll write tests.  Anyway these nested `if`s need tests.

raiden_contracts/deploy/contract_deployer.py
pirapira: Yes, I think so.

PR #843, opened 2019-04-11
Move runtime_code() function into ContractManager

### LINKED ISSUE #670: runtime_bytecode() function should be a method of ContractManager class
`runtime_hexcode` is a function that takes a `ContractManager` and a name of the contract, and returns the runtime bytecode.  This function should be turned into a method of `ContractManager`.

This would change
```
def runtime_hexcode(contracts_manager, name):
```
into
```
def runtime_hexcode(self, name):
```
and the body of the function will change accordingly.

pirapira:
This closes #670.


pirapira:
Tests will turn green after #844 is merged into `master`, and this PR is rebased.

PR #844, opened 2019-04-11
Do not change gas.json unless explicitly told

pirapira:
in the command line argument 'test_print_gas'.

For historical reasons one test case prints the gas consumption in
gas.json.  There is another test that checks the integrity of gas.json
but it fails when the two tests run in parallel.

This fixes https://github.com/raiden-network/raiden-contracts/issues/839

As a solution, this commit disables the writing unless the
command line argument contains 'test_print_gas'.

### DIFF
--- a/raiden_contracts/tests/fixtures/base/utils.py
+++ b/raiden_contracts/tests/fixtures/base/utils.py
@@ -1,4 +1,5 @@
 import json
+from sys import argv
 from typing import Dict
 
 import pytest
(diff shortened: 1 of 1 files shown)


pirapira:
I feel I chose a dirty solution, but I don't come up with any other simple solutions.

pirapira:
@karlb thanks for the link.  That looks cleaner.  I merge this right now because I want tests to pass, but will try to pick up the idea of a special command line option.

Issue #849, opened 2019-04-12
OneToN: include chain_id and one_to_n_contract_address in session id


karlb:
If we don't include this in the session_id and signature, the same IOU can be redeemed multiple times on different chains or different versions of the OneToN contract.

PR #853, opened 2019-04-12
Add one_to_n_address and chain_id as arguments to claim()

pirapira:
OneToN.claim().  Otherwise, the same IOU would be usable on different
instances of OneToN contract.

This commit fixes https://github.com/raiden-network/raiden-contracts/issues/849


pirapira:
By the way, `chain_id` value is never checked in any contracts https://github.com/raiden-network/raiden-contracts/issues/855. That should be fixed.

raiden_contracts/utils/proofs.py
```diff
@@ -231,12 +231,16 @@ def sign_one_to_n_iou(
         receiver,
         amount,
         expiration,
```
karlb: Since we're changing the signature of this function anyway, could we use this opportunity to rename `expiration` to `expiration_block`? Then all arguments would be consistent with the JSON representation of the IOU. And it's also more descriptive.

pirapira: Good idea.  I'll do this in this PR.

pirapira:
True.  And why does the new test succeed.

PR #857, opened 2019-04-15
Pull out a while loop in an if-statement

pirapira:
There was a while loop in the body of an if statement, and
probably this increased the code complexity.

This commit pulls the while loop out of the if statement.
Probably this makes CodeClimate happier.

### DIFF
--- a/raiden_contracts/tests/test_etherscan_verify.py
+++ b/raiden_contracts/tests/test_etherscan_verify.py
@@ -160,9 +160,13 @@ def test_etherscan_verify_already_verified():
         etherscan_api = api_of_chain_id[chain_id]
         m.post(
             etherscan_api,
-            text='{ "status": "0",'
-            '"result" : "Contract source code already verified",'
-            '"message" : "" }',
+            text="""
+            {
+                "status": "0",
+                "result" : "Contract source code already verified",
+                "message" : ""
+            }
+            """,
         )
         runner = CliRunner()
         result = runner.invoke(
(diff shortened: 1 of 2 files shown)


pirapira:
Codecov is complaining.

I'll implement a test to touch
```
        if r['result'] == 'Fail - Unable to verify':
```

pirapira:
The diff coverage should increase from 53.33%

raiden_contracts/tests/test_etherscan_verify.py
```diff
@@ -204,6 +204,37 @@ def test_etherscan_verify_unknown_error():
...
+        etherscan_api = api_of_chain_id[chain_id]
+        m.post(
+            etherscan_api,
+            text='{ "status": "1",'
+            '"result" : "dummy_guid", '
+            '"message" : "" }',
+        )
+        m.get(
+            etherscan_api,
+            text='{ "status": "0",'
```
palango: Would it be easier/nicer to use 
```
text="""
{...}
"""
```
here?

pirapira: Yes, that's much nicer.  I'll change these in this PR.

raiden_contracts/tests/test_etherscan_verify.py
```diff
@@ -160,9 +160,13 @@ def test_etherscan_verify_already_verified():
         etherscan_api = api_of_chain_id[chain_id]
         m.post(
             etherscan_api,
-            text='{ "status": "0",'
-            '"result" : "Contract source code already verified",'
-            '"message" : "" }',
+            text="""
```
palango: From a quick ipython session it seems like you don't even need to escape the `"`:
```
In [1]: a = """
   ...: {"abc": "def"}
   ...: """

In [2]: a
Out[2]: '\n{"abc": "def"}\n'

In [3]:             text="""
   ...:             {
   ...:                 \"status\": \"0\",
   ...:                 \"result\" : \"Contract source code already verified\",
   ...:                 \"message\" : \"\"
   ...:             }
   ...:             """,

In [4]: text
Out[4]: ('\n{\n    "status": "0",\n    "result" : "Contract source code already verified",\n    "message" : ""\n}\n',)
```

pirapira: My bad, not even tried this.

PR #861, opened 2019-04-15
Remove render_contracts that no longer exists

pirapira:
README contained a line `make render_contracts`. This does not
work anymore.  The line is a leftover from the time we tried
Mustache.


pirapira:
Thank you!

PR #863, opened 2019-04-15
Deployed 0.4.0 on Goerli

pirapira:
In https://github.com/raiden-network/raiden/pull/3808 we found that the Raiden client expects the 0.4.0 deployment on the Goerli chain.  This commit adds that deployment.


pirapira:
@czepluch ah, I forgot etherscan.  Doing now.  Thanks.

pirapira:
Now I see etherscan_verify.py doesn't work with older versions. https://github.com/raiden-network/raiden-contracts/issues/864

pirapira:
This was a first step for https://github.com/raiden-network/raiden/issues/3819.

PR #869, opened 2019-04-16
Split two cases with_limits and without_limits

pirapira:
Split `register_token_network()` into two functions for deploying a TokenNetwork with deposit limits, and another function for deploying a TokenNetwork without deposit limits.

The original function contained lots of
```
if with_limits:
    ...
else:
    ...
```

I guess that caused CodeClimate to complain about the complexity of `register_token_network()` function.

### DIFF
--- a/raiden_contracts/deploy/contract_deployer.py
+++ b/raiden_contracts/deploy/contract_deployer.py
@@ -223,26 +223,44 @@ class ContractDeployer(ContractVerifyer):
         """Register token with a TokenNetworkRegistry contract."""
         with_limits = contracts_version_expects_deposit_limits(self.contracts_version)
         if with_limits:
-            if channel_participant_deposit_limit is None:
-                raise ValueError(
-                    'contracts_version 0.9.0 and afterwards expect '
-                    'channel_participant_deposit_limit',
-                )
-            if token_network_deposit_limit is None:
-                raise ValueError(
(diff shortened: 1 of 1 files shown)


raiden_contracts/deploy/contract_deployer.py
```diff
@@ -223,26 +223,44 @@ def register_token_network(
...
+            self,
+            token_registry_abi: Dict,
+            token_registry_address: str,
+            token_address: str,
+            channel_participant_deposit_limit: Optional[int],
+            token_network_deposit_limit: Optional[int],
+    ):
+        """Register token with a TokenNetworkRegistry contract
+
+        where a contract-version that doesn't requires deposit limits in the TokenNetwork
```
LefterisJP: where -> with and requires  -> require

raiden_contracts/deploy/contract_deployer.py
```diff
@@ -255,16 +273,58 @@ def register_token_network(
...
+            self,
+            token_registry_abi: Dict,
+            token_registry_address: str,
+            token_address: str,
+            channel_participant_deposit_limit: Optional[int],
+            token_network_deposit_limit: Optional[int],
+    ):
+        """Register token with a TokenNetworkRegistry contract
+
+        where a contract-version that requires deposit limits in the TokenNetwork
```
LefterisJP: same as above

PR #870, opened 2019-04-16
Put license in each source that's sent around

### LINKED ISSUE #707: Put license in Solidity source themselvess
Just for clarifying the situation when we send around the Solidity sources.

pirapira:
When the source is uploaded to Etherscan, the source should be better
sent with the license. This commit accomplishes that by putting the
license text directly in the Solidity sources.

This fixes #707.


pirapira:
@agatsoh  This one is blocking the next mainnet deployment.

Issue #643, opened 2019-03-01
Use kwargs all over


pirapira:
When arguments are only positionally identified, `f(flavor,version, x, y, z)`, bugs can be introduced when `f` has more arguments or less arguments.  This issue keeps track of using keyword arguments as much as possible.

pirapira:
Is there a tool to detect too many positional arguments?

pirapira:
This is more or less done and that's enough.

PR #873, opened 2019-04-16
Use more keyword arguments

pirapira:
This PR introduces many keyword arguments instead of positional arguments.

This is a part of #643.


raiden_contracts/contract_manager.py
```diff
@@ -216,7 +219,7 @@ def _load_json_from_path(f: Path):
     try:
         with f.open() as deployment_file:
             return json.load(deployment_file)
-    except (FileNotFoundError) as ex:
+    except FileNotFoundError as ex:
```
nlsdfnbch: if you do not need to acces sor use the exception, there's no need for `as ex`.

pirapira: I'll fix this in this PR.

PR #874, opened 2019-04-16
A new method get_constructor_argument_types()

### LINKED ISSUE #871: A part of get_constructor_args() should become a method of ContractManager
Currently in `etherescan_verify.py` there is a function called `get_constructor_args()`.   This function fetches some data from ContractManager object.  This function should become a method of ContractManager class.

pirapira:
These three lines were in etherscan_verify.py but they belong to
ContractManager in terms of functionality.

This commit closes #871.

### DIFF
--- a/raiden_contracts/contract_manager.py
+++ b/raiden_contracts/contract_manager.py
@@ -85,6 +85,14 @@ class ContractManager:
             event_name=event_name,
         )
 
+    def get_constructor_argument_types(self, contract_name: str) -> List:
+        abi = self.get_contract_abi(contract_name=contract_name)
+        return [
+            arg['type'] for arg in list(
+                filter(lambda x: x['type'] == 'constructor', abi),
+            )[0]['inputs']
+        ]
+
     @property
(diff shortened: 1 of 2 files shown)


raiden_contracts/contract_manager.py
```diff
@@ -82,6 +82,14 @@ def get_event_abi(self, contract_name: str, event_name: str) -> Dict:
         contract_abi = self.get_contract_abi(contract_name)
         return find_matching_event_abi(contract_abi, event_name)
 
+    def get_constructor_argument_types(self, contract_name: str) -> List:
+        abi = self.get_contract_abi(contract_name=contract_name)
+        return [
+            arg['type'] for arg in list(
```
ulope: Nitpick: I know you only moved this, but if you want this could be cleaned up into a single list comprehension without the nested `list` and `filter` calls.

pirapira: Good idea.  I'll do this in this PR.

Issue #847, opened 2019-04-12
Remove Python 'assert' uses that might sometimes fail


pirapira:
As @nlsdfnbch [points out](https://github.com/raiden-network/raiden-contracts/pull/842#discussion_r274887666), Python `assert` is no-op when the interpreter is not in debug-mode.  This issue keeps track of removing `assert(...)` that's used for detecting non-internal errors that can happen (wrong user inputs, wrong file content, wrong smart contract responses etc).

pirapira:
Currently, I'm pushed for the main net release, and I'm not doing this anymore.  Removing my assignment.

PR #878, opened 2019-04-16
Replace some asserts with if-statements

pirapira:
This is a part of #847.

### DIFF
--- a/raiden_contracts/deploy/contract_verifyer.py
+++ b/raiden_contracts/deploy/contract_verifyer.py
@@ -145,7 +145,7 @@ class ContractVerifyer:
         chain_id = int(self.web3.version.network)
         assert deployment_data is not None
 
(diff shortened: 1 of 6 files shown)


raiden_contracts/deploy/contract_verifyer.py
```diff
@@ -143,8 +145,10 @@ def _verify_deployment_data(
         chain_id = int(self.web3.version.network)
         assert deployment_data is not None
 
-        assert self.contract_manager.version_string() == deployment_data['contracts_version']
-        assert chain_id == deployment_data['chain_id']
+        if self.contract_manager.version_string() != deployment_data['contracts_version']:
```
nlsdfnbch: `version_string`, if you need it dynamically generated, should be a property instead of a method.

pirapira: That's a good idea. I'll fix this in this PR.

PR #879, opened 2019-04-16
Remove a wrong 'is' usage

pirapira:
Comparing a string against a string literal with 'is' looks wrong,
said pylint.


pirapira:
That's a good link! Thank you.

PR #884, opened 2019-04-18
Use pipdeptree to detect inconsistent dependencies

### LINKED ISSUE #882: CI should fail when pip finds inconsistent versions
When `pip install ...` finds an inconsistency like `A has requirement B<2.0.0,>=1.1.0, but you'll have B 2.1.0 which is incompatible`, the Travis build should fail.

pirapira:
So that when pip installs incompatible packages, the CI builds fail.

Closes #882.


pirapira:
I saw a failure for a dependency conflict.  That's a good sign.

pirapira:
Thank you!

Issue #868, opened 2019-04-16
Update to eth-tester for Istanbul compatibility


palango:
According to the [changelog](https://github.com/ethereum/eth-tester/blob/master/CHANGELOG) this sets the default VM rules to Constantinople

pirapira:
eth-tester 0.1.0-beta.39 causes some version inconsistency https://github.com/raiden-network/raiden-contracts/pull/876#issuecomment-484396973 .  The next step: somehow figure out the versions, or wait for the 0.1.0 release.

pirapira:
`web3` package up to `4.9.2` has version conflicts with the new `eth-tester`.  `web3` from `5.0.0a1` has a circular dependency `web3 => ethpm => web3`.

pirapira:
But I don't care about the circular dependency.  So filing a PR: https://github.com/raiden-network/raiden-contracts/pull/891

pirapira:
I'll wait till `web3>=5.0.0` is available (maybe beta is good enough).

PR #891, opened 2019-04-25
Upgrade eth-tester

pirapira:
This PR upgrades eth-tester.  Then the tests run the Constantinople rules.  That enables upgrading the Solidity compiler to 0.5.7.

This closes https://github.com/raiden-network/raiden-contracts/issues/868

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
   - pip install pytest-travis-fold
   - pip install -r requirements-dev.txt
   - pip install pytest-xdist pytest-sugar codecov
-  - pipdeptree --warn fail
+  - pip check
   - make verify_contracts
 
 before_script:
--- a/raiden_contracts/deploy/contract_deployer.py
+++ b/raiden_contracts/deploy/contract_deployer.py
@@ -47,7 +47,7 @@ class ContractDeployer(ContractVerifyer):
         if gas_price != 0:
             self.transaction['gasPrice'] = gas_price * int(units['gwei'])
 
-        self.web3.middleware_stack.add(
+        self.web3.middleware_onion.add(
             construct_sign_and_send_raw_middleware(private_key),
         )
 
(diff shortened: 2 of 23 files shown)


.travis.yml
```diff
@@ -26,7 +26,7 @@ install:
   - pip install pytest-travis-fold
   - pip install -r requirements-dev.txt
   - pip install pytest-xdist pytest-sugar codecov
-  - pipdeptree --warn fail
+  - pip check
```
palango: Nice, didn't know this existed.

pirapira:
Codecov is complaining about diff coverage, but I defer the action to https://github.com/raiden-network/raiden-contracts/issues/897 .

raiden_contracts/deploy/contract_deployer.py
```diff
@@ -170,7 +170,7 @@ def deploy_raiden_contracts(
 
         deployed_contracts: DeployedContracts = {
             'contracts_version': self.contract_version_string(),
-            'chain_id': int(self.web3.version.network),
+            'chain_id': int(self.web3.eth.protocolVersion),
```
palango: I think these replacements are wrong: ` web3.eth.protocolVersion` is the protocol version (which is something in the 60s right now) and `web3.version.network` is the chain or network id. You can access it with `web3.net.version`.

From a tester chain:
```
(Pdb++) web3.net.version
'1'
(Pdb++) web3.eth.protocolVersion
'99'
(Pdb++) web3.version.network
'1'
```

requirements.txt
```diff
@@ -4,6 +4,6 @@ coincurve>=8.0.0
 py-solc>=3.0.0
 semver
 mypy-extensions
-web3
+web3>=5.0.0a1
```
palango: I'm not sure if we should do this at this point, as this will require the client to update as well.

raiden_contracts/deploy/contract_deployer.py
pirapira: Thanks for checking this.

pirapira:
Then I give this up for now.

PR #898, opened 2019-04-29
Replace asserts that might fail

pirapira:
with an explicit exception raising.

I made changes in `deploy/__main__.py` and tried to satisfy Codecov.  As a side-effect the code coverage increased by 2%.


pirapira:
Codecov is complaining.  I'll try `with patch.object` to touch the diff.

pirapira:
Codecov is complaining.  I should be able to touch the line.  https://stackoverflow.com/questions/39183741/how-to-mock-components-within-unit-tests-for-click-based-cli-applications

pirapira:
The coverage improved a bit.  Maybe `private_key_to_address` is always failing?

pirapira:
The coverage increased further.

pirapira:
I'll do that in this PR.

PR #908, opened 2019-04-29
Enable pylint against too many instance attributes

pirapira:
This hit ChannelValues class so I introduced LockedAmounts that
combined three attributes of ChannelValues.


raiden_contracts/tests/fixtures/channel.py
```diff
@@ -732,34 +737,41 @@ def get(
...
-                vals_B.transferred,
-                vals_B.locked,
-                vals_B.locksroot,
-                A,
-                vals_A.transferred,
-                vals_A.locked,
-                vals_A.locksroot,
+                channel_identifier=channel_identifier,
+                participant1=B,
+                participant1_transferred_amount=0,
```
hackaugusto: why are you not using `transferred` anymore?

pirapira: oops.  That's a bug.

PR #909, opened 2019-04-30
Force ordering of monitor calls

karlb:
Closes raiden-network/raiden-services#287.


raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -211,6 +216,70 @@ contract MonitoringService is Utils {
...
+        address participant2,
+        address monitoring_service_address
+    )
+        public pure
+        returns (uint256)
+    {
+        // avoid overflows when multiplying with percentages
+        require(settle_timeout < uint256(2**256 - 1) / 100);
+
+        // First allowed block as percentage of settle_timeout. We're using iteger's
```
pirapira: nitpick: `iteger`-->`integer`.

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -211,6 +216,70 @@ contract MonitoringService is Utils {
...
+        uint256 settle_timeout,
+        address participant1,
+        address participant2,
+        address monitoring_service_address
+    )
+        public pure
+        returns (uint256)
+    {
+        // avoid overflows when multiplying with percentages
+        require(settle_timeout < uint256(2**256 - 1) / 100);
```
pirapira: To prevent an overflow in `best_case_block`, I'd check
```
require(closed_at_block < uint256(2**256 - 1) / 100);
require(settle_timeout < uint256(2**256 - 1) / 100 - closed_at_block);
```

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -211,6 +216,70 @@ contract MonitoringService is Utils {
...
+        // Length of the range into which the first allowed block will fall
+        uint256 range_length = (WORST_CASE - BEST_CASE) * settle_timeout / 100;
+
+        // Offset for this specific MS within the range
+        uint256 ms_offset = (
+            uint256(participant1)
+            + uint256(participant2)
+            + uint256(monitoring_service_address)
+        ) % range_length;
+        assert(ms_offset <= best_case_block + range_length);
```
pirapira: Shouldn't this be `assert(ms_offset < range_length);`?

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -175,6 +175,15 @@ contract MonitoringService is Utils {
         canMonitor(msg.sender)
         public
     {
+        TokenNetwork token_network = TokenNetwork(token_network_address);
+        uint256 channel_identifier = token_network.getChannelIdentifier(
+            closing_participant, non_closing_participant
+        );
```
pirapira: I have to think through. What can happen when `getChannelIdentifier()` calls back into the MS contract?  (The address `token_network` is given from outside so cannot be trusted.)

raiden_contracts/data/source/services/MonitoringService.sol
karlb: > require(closed_at_block < uint256(2**256 - 1) / 100);

I was assuming that block numbers don't grow that large. I'll add it to make that explicit.

> require(settle_timeout < uint256(2**256 - 1) / 100 - closed_at_block);

I don't see a need for adding `closed_at_block`:
```
best_case_block
= closed_at_block + BEST_CASE * settle_timeout / 100
<= closed_at_block + 100 * settle_timeout / 100
= closed_at_block + settle_timeout
< uint256(2**256 - 1) / 100 + uint256(2**256 - 1) / 100
= uint256(2**256 - 1) / 50
< uint256(2**256 - 1)
```

raiden_contracts/data/source/services/MonitoringService.sol
karlb: Yes, you're right. I'll just remove that assert, since that property is obviously true after the modulo operation.

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -211,6 +216,70 @@ contract MonitoringService is Utils {
...
+        uint256 settle_block_number;
+        (settle_block_number, channel_state) = token_network.getChannelInfo(
+            channel_identifier, closing_participant, non_closing_participant
+        );
+        require(channel_state == TokenNetwork.ChannelState.Closed);
+
+        // We don't actually know when the channel has been closed. So we'll
+        // make a guess so that assumed_close_block >= real_close_block.
+        uint256 assumed_settle_timeout = token_network.settlement_timeout_min();
+        uint256 assumed_close_block = settle_block_number - assumed_settle_timeout;
```
pirapira: Before the subtraction, let's add `require(settle_block_number >= assumed_settle_timeout);`.

`token_network` is an externally given address and can be arbitrarily nasty.

raiden_contracts/data/source/services/MonitoringService.sol
pirapira: Right. We know `uint256(2**256 - 1) / 100` is a small number!

raiden_contracts/data/source/services/MonitoringService.sol
pirapira: OK!

raiden_contracts/data/source/services/MonitoringService.sol
karlb: My current thoughts on this are

1. The client can't be cheated by providing a bad token network, since the client's signature is checked and includes the token network address.

2. The MS could already be cheated before, because it already has a call to `token_network.updateNonClosingBalanceProof`. The MS will have to verify that the token network is registered in the token network registry.

Is it safe to assume that a registered token network behaves as intended in the long run?

pirapira: When `token_network` is malicious, it can call back `monitor()` and register a balance proof of the same nonce on the same channel. For that,
* the channel participant must have given a signed balance proof to the attacker
* the victim MS must have overlooked the weird feature of the TokenNetwork.

What else...?

pirapira: Instead of thinking about all possibilities, I suggest moving everything below `updateReward()`.

* That doesn't change the functionality in the normal executions.  When any of the `require` fails, the storage-change in `updateReward()` is reverted.
* That causes the reentrant call to fail, when it's about the same channel and the same nonse.

karlb: That helps against reentrancy, but the attacker could still make the transaction fail so that the MS spent gas on the transaction but does not get any reward.

pirapira: > the attacker could still make the transaction fail 

I agree. It's just a matter of making sure nothing worse happens.

Even with the current code, that might be the case, but with reentrancy, the property is harder to establish.  All users need to think about every possible re-entry into any external/public functions.

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -211,6 +218,71 @@ contract MonitoringService is Utils {
...
+        address participant1,
+        address participant2,
+        address monitoring_service_address
+    )
+        public pure
+        returns (uint256)
+    {
+        // avoid overflows when multiplying with percentages
+        require(settle_timeout < uint256(2**256 - 1) / 100);
+        require(closed_at_block < uint256(2**256 - 1) / 100);
```
pirapira: Just a thought & no need to react: I wondered how Solidity compiler understands `2**256`.  It's a complicated way to write zero, but it works and conveys the intuition.

Maybe I'll move `MAX_SAFE_UINT256` from `TokenNetwork.sol` into `Utils.sol`, and use the constant here.

PR #896, opened 2019-04-26
Split store_and_verify_deployment_info_services()

pirapira:
because this function took a boolean and did two completely different things depending on the boolean.

PR #917, opened 2019-05-02
Fix how to call verify_contracts_deployment_data

pirapira:
This is a leftover of #896.  Some function signatures changed there
but not all calling sites were updated accordingly.

I expect a Codecov error right now.  I need to write tests to cover those cases.


pirapira:
Thank you!

PR #920, opened 2019-05-03
Replace asserts in _verify_deployed_contract()

pirapira:
with if statements.

This is a part of https://github.com/raiden-network/raiden-contracts/issues/847.

I'm expecting a Codecov failure, and I'll add tests.


pirapira:
```
if receipt['contractAddress'] != contracts[contract_name]['address']:
```
This one is hard to break.  When I set a wrong value in `contracts[contract_name]['address']`, an earlier check about the onchain code fails.

pirapira:
But, I should be able to split away a function that takes a receipt and `contracts`, so I can touch the case.

pirapira:
Thank you!

PR #921, opened 2019-05-03
Replace asserts with if statements in verify_service_contracts_deployment_data()

pirapira:
verify_service_contracts_deployment_data()

I'm first expecting a CodeCov failure.

### DIFF
--- a/raiden_contracts/deploy/contract_verifier.py
+++ b/raiden_contracts/deploy/contract_verifier.py
@@ -273,19 +273,25 @@ def _verify_user_deposit_deployment(
     monitoring_service_address: Address,
 ):
     """ Check an onchain deployment of UserDeposit and constructor arguments at deployment time """
-    assert len(constructor_arguments) == 2
-    assert to_checksum_address(user_deposit.functions.token().call()) == token_address
-    assert token_address == constructor_arguments[0]
-    assert user_deposit.functions.whole_balance_limit().call() == user_deposit_whole_balance_limit
+    if len(constructor_arguments) != 2:
+        raise RuntimeError("UserDeposit has a wrong number of constructor arguments.")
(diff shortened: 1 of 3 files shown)


pirapira:
I think I need to split the tested function so that each function's failure case can be reached easily.  This will be done in a separate PR.

pirapira:
Codecov is complaining.  Perhaps we should separate the condition checking into a different function so that it can be unit-tested more easily.

pirapira:
That's why #924 exists.

raiden_contracts/deploy/contract_verifier.py
```diff
@@ -273,19 +273,25 @@ def _verify_user_deposit_deployment(
...
+        raise RuntimeError("UserDeposit has a wrong token address onchain.")
+    if user_deposit.functions.whole_balance_limit().call() != user_deposit_whole_balance_limit:
+        raise RuntimeError("UserDeposit has a wrong whole_balance_limit onchain")
+    if user_deposit_whole_balance_limit != constructor_arguments[1]:
+        raise RuntimeError("UserDeposit received a wrong whole_balance_limit during construction.")
     assert user_deposit_whole_balance_limit == constructor_arguments[1]
-    assert (
-        to_checksum_address(user_deposit.functions.one_to_n_address().call()) == one_to_n_address
-    )
+    if to_checksum_address(user_deposit.functions.one_to_n_address().call()) != one_to_n_address:
```
palango: You could also use `eth_utils.is_same_address` instead of converting them. But the advantage of your way is that you make sure it's actually a checksum address.

pirapira: I didn't think about that.  Maybe `is_same_address` says more precisely what I wanted to check here.

PR #924, opened 2019-05-03
Split subroutines away from verify_service_contracts_deployment()

pirapira:
Previously it was very difficult to test those failure cases where
a contract's address stored in another contract's storage is wrong.
For that I would need to mock the contract's function calls, and that's
somewhere deep inside web3.py.

Instead, I chose to split away those checks in smaller functions
so I can write tests on these checks separately, with more precise
control over which assertion should fail.

This is a prerequisite for #921. There, since tests are hard to write,
I couldn't calm down CodeCov.


pirapira:
Thank you!

PR #928, opened 2019-05-06
Introduce the black formatter

### LINKED ISSUE #893: Enable black formatter
The first step is to figure out what `raiden-services` is doing with `black`.

pirapira:
This closes #893.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+.PHONY: all compile_contracts verify_contracts install lint isort black format mypy clean release
+
 all: verify_contracts install
 
 compile_contracts:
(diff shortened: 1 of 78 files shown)


Makefile
```diff
@@ -12,14 +12,20 @@ install:
...
 lint:
+	black --check $(BLACK_PARAMS)
 	flake8 raiden_contracts/
 	pylint --rcfile .pylint.rc --load-plugins pylint_quotes raiden_contracts/
 	isort $(ISORT_PARAMS) --check-only
 
 isort:
 	isort $(ISORT_PARAMS)
 
+black:
```
palango: In the services we have an additional `format` target which runs `isort` and `black`, which is quite convenient.

pirapira: Good idea.  Added in this PR.

PR #954, opened 2019-05-09
Deploy contracts_version 0.12.0

pirapira:
* MonitoringService contract now sets the priority on services.  Which service can call `monitor()` first.
* OneToN contract now checks the chain ID.


karlb:
I don't see an update for the ChangeLog.

pirapira:
I was going to do that in the next PR, for bumping the PyPI package version.  Would it work?

karlb:
Yes, that's ok, too. As long as no change happens in between, of course. I was just not what I was expecting.

PR #956, opened 2019-05-09
Fix black --check

pirapira:
Somehow `black --check` failed on `master`. This PR fixes it.


pirapira:
Surprisingly, Travis was not configured to enforce a successful `make lint`.  This PR also fixes that.

pirapira:
Codecov is complaining but @agatsoh was writing a test for mint_token script...

PR #965, opened 2019-05-13
Replace asserts in sign()

pirapira:
This PR replaces asserts in `sign()` with `if not ...: raise ...`.  Sometimes `assert` is omitted in an optimised execution, but input checks should be always performed.

This PR is a part of #847.

I expect a CodeCov failure, and then I'll add tests to cover the new paths.

### DIFF
--- a/raiden_contracts/utils/signature.py
+++ b/raiden_contracts/utils/signature.py
@@ -7,22 +7,22 @@ from eth_utils.typing import ChecksumAddress
 sha3 = keccak
 
 
-def sign(privkey: str, msg: bytes, v: int = 0) -> bytes:
-    if not isinstance(msg, bytes):
-        raise TypeError("sign(): msg is not an instance of bytes")
-    if len(msg) != 32:
-        raise ValueError("sign(): msg has to be exactly 32 bytes")
+def sign(privkey: str, msg_hash: bytes, v: int = 0) -> bytes:
+    if not isinstance(msg_hash, bytes):
+        raise TypeError("sign(): msg_hash is not an instance of bytes")
+    if len(msg_hash) != 32:
+        raise ValueError("sign(): msg_hash has to be exactly 32 bytes")
     if not isinstance(privkey, str):
         raise TypeError("sign(): privkey is not an instance of str")
     if v not in {0, 27}:
         raise ValueError(f"sign(): got v = {v} expected 0 or 27.")
 
     pk = PrivateKey.from_hex(remove_0x_prefix(privkey))
-    sig: bytes = pk.sign_recoverable(msg, hasher=None)
+    sig: bytes = pk.sign_recoverable(msg_hash, hasher=None)
     assert len(sig) == 65
 
     pub = pk.public_key
-    recovered = PublicKey.from_signature_and_message(sig, msg, hasher=None)
+    recovered = PublicKey.from_signature_and_message(sig, msg_hash, hasher=None)
     assert pub == recovered
 
     sig = sig[:-1] + bytes([sig[-1] + v])
--- a/raiden_contracts/utils/signature.py
+++ b/raiden_contracts/utils/signature.py
@@ -9,11 +9,11 @@ sha3 = keccak
 
 def sign(privkey: str, msg: bytes, v: int = 0) -> bytes:
     if not isinstance(msg, bytes):
-        raise TypeError('sign(): msg is not an instance of bytes')
+        raise TypeError("sign(): msg is not an instance of bytes")
     if len(msg) != 32:
-        raise ValueError('sign(): msg has to be exactly 32 bytes')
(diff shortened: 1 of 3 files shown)


raiden_contracts/utils/signature.py
```diff
@@ -8,12 +8,14 @@
...
 
 def sign(privkey: str, msg: bytes, v=0) -> bytes:
-    assert isinstance(msg, bytes)
-    assert isinstance(privkey, str)
+    if not isinstance(msg, bytes):
+        raise TypeError("sign(): msg is not an instance of bytes")
+    if len(msg) != 32:
+        raise ValueError("sign(): msg has to be exactly 32 bytes")
+    if not isinstance(privkey, str):
+        raise TypeError("sign(): privkey is not an instance of str")
```
err508: I think here's an `assert v in (0,27)` missing, otherwise looks good

err508: Also an assertion, that the signature is valid and recoverable would be nice

pirapira: I'll do these in this PR.

pirapira:
Now I'll write tests where `sign()` is called with
* a message that's not a byte sequence
* a message that's not exactly 32 bytes long
* a private key that's not a string
* v not being 0 or 27.

raiden_contracts/utils/signature.py
```diff
@@ -8,15 +8,23 @@
 
 
 def sign(privkey: str, msg: bytes, v: int = 0) -> bytes:
-    assert isinstance(msg, bytes)
```
err508: Just wondering, should we use privkey: str or privkey: bytes to get rid of potential encoding errors here?

pirapira: Good point.  I added an issue to track this https://github.com/raiden-network/raiden-contracts/issues/1059 .

raiden_contracts/utils/signature.py
```diff
@@ -8,15 +8,23 @@
...
+        raise ValueError("sign(): msg has to be exactly 32 bytes")
+    if not isinstance(privkey, str):
+        raise TypeError("sign(): privkey is not an instance of str")
+    if v not in {0, 27}:
+        raise ValueError(f"sign(): got v = {v} expected 0 or 27.")
 
     pk = PrivateKey.from_hex(remove_0x_prefix(privkey))
-    assert len(msg) == 32
-
     sig: bytes = pk.sign_recoverable(msg, hasher=None)
```
err508: Don't want to start to bikeshed here, but should we fix ` hasher = keccak256` as in the contracts? in case coincurve switches their default for some reason and we wouldn't notice in time, nodes would produce signatures which would not be accepted onchain.

err508: Okay I just saw you have a test for ECVerify and coincurve compatibility, so we would at least notice.

err508: Third thoughts: So you do the hashing on your side, which is why `hasher=None`. If you want to keep it like this, I'd change `message` to `message_hash` in the function signature. 
Also, hashing as part of signing are strongly coupled operations, so maybe moving the hashing into the sign function makes sense? Sign would than accept size wise unrestricted messages and it would be almost impossible to break compatibility between hashing and signing.

raiden_contracts/utils/signature.py
```diff
@@ -8,15 +8,23 @@
...
     assert len(sig) == 65
 
+    pub = pk.public_key
+    recovered = PublicKey.from_signature_and_message(sig, msg, hasher=None)
+    assert pub == recovered
```
err508: I know you have a test for this already, but an additional `assert signature_test_contract.functions.verify(msg, sig).call() == pub_to_address` (maybe you can make this pretty) would give me some serious peace of mind here.

raiden_contracts/utils/signature.py
pirapira: Let me have a look in `raiden` how they do it.  The idea currently is to share the implementation first and improve the shared implementation.

raiden_contracts/utils/signature.py
pirapira: I'll add it in this PR.

raiden_contracts/utils/signature.py
pirapira: For the moment, I just change the function signature here.

pirapira: The rest will be kept track in https://github.com/raiden-network/raiden-contracts/issues/1041

raiden_contracts/utils/signature.py
pirapira: Or not.  The problem is that we don't have access to any smart contract here.

pirapira: https://github.com/raiden-network/raiden-contracts/issues/1067 will keep track of this.

Issue #770, opened 2019-03-25
Add a test to deploy 0.4.0 contracts and register tokens


pirapira:
because this seems to be still a common usage.

pirapira:
Also a test for 0.3._ is necessary to prevent bugs like #787 .

pirapira:
Just calling `deploy_raiden_contracts` is not enough.  We need to check the option parsing as well.  So we need to use click to run the command.

pirapira:
We've done "calling the script with CliRunner".  Are we already testing contract_deployer with 0.4.0?

pirapira:
test_deploy_script_register_without_limit already does that.

PR #966, opened 2019-05-13
Add a test to call the deploy script with 0.4.0

pirapira:
contracts_version.

This is a part of https://github.com/raiden-network/raiden-contracts/issues/770

### DIFF
--- a/raiden_contracts/tests/test_deploy_script.py
+++ b/raiden_contracts/tests/test_deploy_script.py
@@ -625,9 +625,7 @@ def test_deploy_token_with_balance(get_accounts, get_private_key):
 
 
 def deploy_raiden_arguments(
-        privkey: str,
-        save_info: Optional[bool],
-        contracts_version: Optional[str],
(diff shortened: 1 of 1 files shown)


raiden_contracts/tests/test_deploy_script.py
```diff
@@ -624,27 +624,34 @@ def test_deploy_token_with_balance(get_accounts, get_private_key):
...
+    else:
+        max_arguments = []
+
+    if contracts_version is None:
+        contracts_version_arguments: List = []
+    else:
+        contracts_version_arguments = ["--contracts-version", contracts_version]
+
+    common_arguments: List = ["--private-key", privkey, "--rpc-provider", "rpc_provider"]
+    return common_arguments + save_info_arguments + contracts_version_arguments + max_arguments
```
karlb: Wouldn't it be easier to have only one `arguments` list append to conditionally? You could skip all the `..._arguments = []` branches and the concatenation at the end.

pirapira: I agree.  Will do in this PR.

PR #970, opened 2019-05-13
Prevent tests from writing to contracts.json

### LINKED ISSUE #912: Tests should not modify contracts.json in place
Sometimes when I run `pytest -n auto raiden_contracts/tests`, I see failures for empty `contracts.json`.  I think this is because some tests are modifying `contracts.json`.  This issue keeps track of making sure that no tests modify `contracts.json`.

pirapira:
This fixes #912.

### DIFF
--- a/raiden_contracts/tests/fixtures/base/contract_manager.py
+++ b/raiden_contracts/tests/fixtures/base/contract_manager.py
@@ -13,5 +13,5 @@ def contract_source_manager():
 
 @pytest.fixture(scope="session")
 def contracts_manager(contract_source_manager):
-    target_path = NamedTemporaryFile()
-    return contract_source_manager.compile_contracts(Path(target_path.name))
+    with NamedTemporaryFile() as target_path:
+        yield contract_source_manager.compile_contracts(Path(target_path.name))
--- a/raiden_contracts/tests/fixtures/base/contract_manager.py
+++ b/raiden_contracts/tests/fixtures/base/contract_manager.py
@@ -1,6 +1,8 @@
+from pathlib import Path
+from tempfile import NamedTemporaryFile
+
 import pytest
 
-from raiden_contracts.contract_manager import contracts_precompiled_path
 from raiden_contracts.contract_source_manager import ContractSourceManager, contracts_source_path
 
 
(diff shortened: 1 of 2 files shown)


raiden_contracts/tests/fixtures/base/contract_manager.py
```diff
@@ -11,4 +13,5 @@ def contract_source_manager():
 
 @pytest.fixture(scope="session")
 def contracts_manager(contract_source_manager):
-    return contract_source_manager.compile_contracts(contracts_precompiled_path())
+    target_path = NamedTemporaryFile()
+    return contract_source_manager.compile_contracts(Path(target_path.name))
```
nlsdfnbch: If you `yield` instead of `return`ing the value (i.e. turn your fixtures into generators), you can add clean up code after the `yield` statement, which is called after the fixture's specified scope ends:

```python
def contracts_manager(contract_source_manager):
    target_path = NamedTemporaryFile()
    yield contract_source_manager.compile_contracts(Path(target_path.name))
    target_path.close()
```

Of course, you can also use the context manager, making it a bit prettier:

```python
def contracts_manager(contract_source_manager):
    with NamedTemporaryFile() as target_path:
        yield contract_source_manager.compile_contracts(Path(target_path.name))
```

In addition, I'd like to point out that there's a fixture for this already, supplied by pytest: `tmp_path`. By default, its contents are removed after every test run. Luckily, there's a factory which creates fixtures with a session scope as well:

```python
def session_tmp_file(tmp_path_factory):
    test_session_fpath = tmp_path_factory.mktemp("data").join("test_file")
    # Touch to create
    test_session_fpath.touch()
    return test_session_fpath

def contracts_manager(contract_source_manager, session_tmp_file):
    # Note: we no longer need to clean up anything - this is done by the fixture automatically; hence, no yield.
    return contract_source_manager.compile_contracts(session_tmp_file.name)
```

Perhaps @karlb has a better approach than mine - I believe him to be much better versed in all things `pytest`.

pirapira: I'll `yield` in the context manager.  Thanks!

pirapira:
Thank you!

PR #979, opened 2019-05-16
Use SHA256 for hashlocks

pirapira:
because it's used more commonly.


pirapira:
The change is straightforward.

PR #984, opened 2019-05-20
Check whether the contracts field is None

pirapira:
The new check makes sure that
`assert self.contracts` in `get_contract()` never fails.

This is a part of #847.


pirapira:
I'll write a test that touches the new codepath, in this PR.

PR #988, opened 2019-05-20
Rename unlock related arguments

### LINKED ISSUE #964: Rename arguments of `unlock` related functions
The party who locked the transfer and the party who whould unlock the transfer should be called `sender` and `receiver` in the functions about `unlock`.

https://github.com/raiden-network/raiden/issues/4055

pirapira:
Before this PR, unlock related functions in TokenNetwork received arguments `participant` and `partner`.  Sometimes `participant` was the sender of the locked transfer, but sometimes `partner` was the sender.

This PR renames these two arguments into `sender` and `receiver`.  The sender of the locked transfer is always named `sender` and the receiver always `receiver`.

This changes the ABI, but now is a good time for breaking changes.

This fixes #964.


pirapira:
Thank you!

Issue #983, opened 2019-05-17
bumpversion --config-file ./.bumpversion_contracts.cfg breaks version numbers


pirapira:
After the command, I saw
```
CONTRACTS_VERSION = "CONTRACTS_VERSION = '0.13.0'"
```

This issue keeps track of preventing this kind of problem on the master branch.

PR #990, opened 2019-05-20
Use double quote in bumpversion template

pirapira:
Since we introduced black, we are using double quotes.

This fixes https://github.com/raiden-network/raiden-contracts/issues/983


pirapira:
yay!

Issue #918, opened 2019-05-03
enable mypy strict mode


pirapira:

pirapira:
Try enabling checks one by one:

https://github.com/raiden-network/raiden-services/blob/master/setup.cfg#L29

pirapira:
In `--strict`, all generic types require type parameters.

pirapira:
Also `mypy --strict` requires annotating many classes and functions in Web3.py.

karlb:
The `--strict` flag is extremely strict (at the moment it's 1150 errors). It's probably not worth making the code adhere to it fully, but we can enable a few more checks in the `setup.cfg`. We're currently missing
```
# disallow_any_generics = True
# disallow_subclassing_any = True
# disallow_untyped_calls = True
# disallow_incomplete_defs = True
# disallow_untyped_decorators = True
# no_implicit_optional = True
# warn_redundant_casts = True
# warn_return_any = True
# no_implicit_reexport = True
```

PR #993, opened 2019-05-20
Random type annotations

pirapira:
This PR adds type annotations to functions.  This is a part of #918.


pirapira:
Maybe you used your luck somewhere else :)

PR #994, opened 2019-05-21
Turn a failable assert into an if statement

pirapira:
This is a part of https://github.com/raiden-network/raiden-contracts/issues/847

I expect CodeCov to complain.  Then I'll add a test that touches the new code path.


pirapira:
First merge #991, and Codecov will be happy.

pirapira:
yay!

PR #995, opened 2019-05-21
Rename Address into AddressHex

pirapira:
to match the names in the Raiden client codebase.

### DIFF
--- a/raiden_contracts/contract_manager.py
+++ b/raiden_contracts/contract_manager.py
@@ -5,11 +5,11 @@ from json import JSONDecodeError
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
+from eth_typing.evm import HexAddress
 from mypy_extensions import TypedDict
 from semver import compare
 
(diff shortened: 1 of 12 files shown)


pirapira:
@palango pointed out that I could use the type defined in `eth_typing` https://github.com/ethereum/eth-typing/blob/d06fcac1b6e396a5ecee70b527469f64ba703943/eth_typing/evm.py#L12

hackaugusto:
why is the code base using `hex` encoded values instead of `binary`?

pirapira:
> why is the code base using `hex` encoded values instead of `binary`?

Maybe because it's hex on every interface of `raiden-contracts`.  In JSON files it's hex, and web3 is also happy with hex (checksummed).

But I believe we should follow the sandwich principle, and the interface between `raiden` and `raiden-contracts` should use the binary format.  I created an issue. https://github.com/raiden-network/raiden-contracts/issues/997

PR #1003, opened 2019-05-23
Why mypy forces so many changes

pirapira:
because we have been so lazy to annotate types.

This PR introduces a lot of type annotations together with a few interesting changes.

* A new TypedDict `CompiledContract` is introduced.  Previously, a single type `DeployedContract` was used for two completely different kinds of dictionaries.
* A single function with a boolean parameter was duplicated into two.  Depending on the boolean, the function's type signature changed. `contract_manager_meta(_, True)` is now `contract_source_manager_meta(_)`; `contract_manager_meta(_, False)` is now `contract_manager_meta(_)`.


pirapira:
Thank you!

PR #1006, opened 2019-05-24
More type annotations and a small fix

pirapira:
This commit is mostly type annotations.  There is one change that affects the runtime behavior.  Somehow `bytes` were converted into `str` when there was no need to.

This is a part of #918.

### DIFF
--- a/raiden_contracts/tests/fixtures/token_network_registry.py
+++ b/raiden_contracts/tests/fixtures/token_network_registry.py
@@ -1,7 +1,8 @@
-from typing import Callable
+from typing import Any, Callable, Collection, List
 
 import pytest
 from eth_utils import is_address
+from web3 import Web3
 from web3.contract import Contract, get_event_data
 
 from raiden_contracts.constants import (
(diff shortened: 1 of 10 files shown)


raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -759,7 +759,7 @@ def test_channel_unlock_unregistered_locks(
         ChannelValues(deposit=40, withdrawn=10, transferred=20),
     )
 
-    vals_A.locksroot = "0x" + get_merkle_root(pending_transfers_tree.merkle_tree).hex()
```
pirapira: There was no need of conversion here.

raiden_contracts/tests/fixtures/token_network_registry.py
```diff
@@ -1,7 +1,8 @@
-from typing import Callable
+from typing import Any, Callable, Collection, List
 
 import pytest
 from eth_utils import is_address
+from web3 import Web3
```
jomuel: I expected this and the other `Web3` and `Contract` imports to be in an `if TYPE_CHECKING:` block. Is the pattern not used in `raiden_contracts`?

pirapira: No, I've never seen the pattern here.  Probably because in most cases we need the library at runtime.

PR #1015, opened 2019-05-24
Separate get_pending_transfers_tree() into two

### LINKED ISSUE #1002: Simplify the interface of def get_pending_transfers_tree()
The function currently takes both unlockable_amount and unlockable_amounts, and if unlockable_amount is not None, unlockable_amounts is ignored.

The function should only take unlockable_amounts.

pirapira:
because the function had many optional arguments and had
essentially two separate usage patterns.

This fixes #1002.


pirapira:
Thank you!

PR #1023, opened 2019-05-27
Remove contract_version from each contract

### LINKED ISSUE #967: Remove "contract_version" from the Solidity sources
The "contracts_version" is supposed to be common to all contracts, but that doesn't work because SecretRegistry shouldn't be updated as frequently as the other contracts.  This issue keeps track of removing "contracts_version" from the Solidity sources.

For checking the version onchain, the runtime bytecode in `deployment.json` is useful.

pirapira:
In order to prevent deploying multiple SecretRegistry contract on the
same chain, we need to use the same contract deployment in multiple
`contracts_version`s.

So far, every contract had a public state variable contract_version
that changes in sync.  However, since we are going to use the same
SecretRegistry with multiple versions of other contracts, we can no
longer expect `contract_version`s to move in sync.

We can try to accommodate the situation by adding special code to deal with
plural `contract_version`s but the simpler approach is to remove the
`contract_version`s from the Solidity files.

If some new identifier is needed, that will be introduced.

This closes #967.


pirapira:
> Why are the versions being removed?

If we keep the version numbers, we need to keep track of many `contract_version`s within one `contracts_version`.  If we drop the contract_versions, we don't have to keep track of many `contract_versions`.

> some sort of bytecode checks.

Yes, `contracts.json` contains `bin-runtime`, that must be identical to what's seen on Web3 `getCode()`.

pirapira:
The alternative to this PR is to invent a way to manage seven independent version numbers in an efficient way.

Or, perhaps, it's fine to replace the `contract_version` with some random-looking hashes.

pirapira:
ContractVerifyer.verify_deployment_data() checks the bytecode on the chain.

PR #1025, opened 2019-05-27
MonitoringService.monitor() takes a MessageTypeId

### LINKED ISSUE #1021: MonitoringService.monitor() should add a MessageID when it recovers an address from the signature
Currently, when MonitoringService.monitor() recovers an address from a signature, it uses a function `recoverAddressFromRewardProof()`.  The function does not use `MessageTypeID.RewardProof`.  So there can be another kind of signature on another kind of message that accidentally coincides with the signature used here.

This issue keeps track of adding a message ID so the signature used in MonitoringService.monitor() won't be reusable in another context.

See how `TokenNetwork.recoverAddressFromBalanceProofUpdateMessage()` uses `MessageTypeID.BalanceProofUpdate`.

pirapira:
into account when validating the signature.

This closes #1021 .


raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -342,11 +344,13 @@ contract MonitoringService is Utils {
         returns (address signature_address)
     {
         bytes32 message_hash = keccak256(abi.encodePacked(
-            "\x19Ethereum Signed Message:\n148",
+            "\x19Ethereum Signed Message:\n200",
+            monitoring_service_address,
+            chain_id,
+            uint256(MessageTypeId.MSReward),
```
karlb: If we add a message id to the signature, shouldn't we put it in the same place for all messages (e.g. first in the signature)? Otherwise there could still be conflicts (although highly unlikely). Example:

Message1 (of MessageTypeId 5):
* 10: TokenAmount
* 5: MessageTypeId
* signature

Message2 (of MessageTypeId 10):
* 10: MessageTypeId
* 5: ChannelID
* signature

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -342,11 +344,13 @@ contract MonitoringService is Utils {
         returns (address signature_address)
     {
         bytes32 message_hash = keccak256(abi.encodePacked(
-            "\x19Ethereum Signed Message:\n148",
+            "\x19Ethereum Signed Message:\n200",
+            monitoring_service_address,
```
karlb: The reward is not bound to a specific monitoring service. Did you want to use monitoring_service_contract_address here?

raiden_contracts/data/source/services/MonitoringService.sol
pirapira: Yes, we should put the ID in the same place always.

In TokenNetwork, the message ID is always the third element after an address and a chain_id.
```
        bytes32 message_hash = keccak256(abi.encodePacked(
            signature_prefix,
            message_length,
            address(this),
            chain_id,
            uint256(MessageTypeId.Withdraw),
            channel_identifier,
            participant,
            total_withdraw
        ));
```

I can change them as well, but that causes some additional work (adjusting the client).

raiden_contracts/data/source/services/MonitoringService.sol
pirapira: Ah, suddenly I remember that there should be a competition between MSs.

And yes, monitoring_service_contract_address is the right one.

raiden_contracts/data/source/services/MonitoringService.sol
karlb: I didn't expect the third place to be the fixed location. This also required the same length of the first two arguments in all cases to be safe. But I guess this is the case for us. No need to change if you watch out for this.

pirapira: I'll add the convention in the contribution guide.  I think the current convention is
```
address message_destination,
uint256 chain_id,
uin256 message_type,
...
```

I agree the third position looks like a random choice.

pirapira:
@karlb please have another look.

raiden_contracts/utils/proofs.py
```diff
@@ -220,6 +224,7 @@ def sign_reward_proof(
     message_hash = hash_reward_proof(
         channel_identifier=channel_identifier,
         reward_amount=reward_amount,
+        monitoring_service_address=monitoring_service_address,
```
karlb: This should be monitoring_service_contract_address, too.

raiden_contracts/utils/proofs.py
```diff
@@ -106,14 +106,17 @@ def hash_reward_proof(
     channel_identifier: int,
     reward_amount: int,
     token_network_address: HexAddress,
+    monitoring_service_address: HexAddress,
```
karlb: This should be monitoring_service_contract_address, too.

pirapira:
@karlb ready for a merge?

PR #1028, opened 2019-05-28
0.21.0 Release

pirapira:
Mainly for delivering the new gas measurement.


raiden_contracts/data_0.14.0/source/lib/SHA256.sol~
```diff
@@ -0,0 +1,3 @@
+pragma solidity 0.5.4;
```
palango: What is this for?

pirapira: That was my git mistake.  Added a whole directory without looking what's in.

PR #1029, opened 2019-05-28
Replace an assert that might fail

pirapira:
merge_deployment_data() might get two dictionaries that have the
same keys.

This is a part of #847.

### DIFF
--- a/raiden_contracts/contract_manager.py
+++ b/raiden_contracts/contract_manager.py
@@ -174,7 +174,11 @@ def merge_deployment_data(dict1: DeployedContracts, dict2: DeployedContracts) ->
     if not dict2:
         return dict1
     common_contracts: Dict[str, DeployedContract] = deepcopy(dict1["contracts"])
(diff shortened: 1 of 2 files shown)


pirapira:
I need to add a new test case to touch the new path.

raiden_contracts/contract_manager.py
```diff
@@ -174,7 +174,8 @@ def merge_deployment_data(dict1: DeployedContracts, dict2: DeployedContracts) ->
     if not dict2:
         return dict1
     common_contracts: Dict[str, DeployedContract] = deepcopy(dict1["contracts"])
-    assert not common_contracts.keys() & dict2["contracts"].keys()
+    if common_contracts.keys() & dict2["contracts"].keys():
+        raise ValueError("dict1 and dict2 share some keys.")
```
andrevmatos: Maybe some explanation as why they sharing some key is bad would be useful

pirapira: I made a more detailed explanation here.

PR #1031, opened 2019-05-31
An assert might fail, so replacing with if: raise

pirapira:
This is a part of #847.

I'm expecting a Codecov failure first.  Then I'll add a test to fix that.


pirapira:
Thank you!

PR #1040, opened 2019-06-04
Type-annotate all functions

pirapira:
This is a part of #918 .


pirapira:
I think it's better to merge this rather than rebasing it again and again.

Issue #1041, opened 2019-06-04
Exporting packing and signing functions to the Raiden client


pirapira:
Currently both `raiden-contract` and `raiden` implement the same functionality of packing and signing onchain-submitted messages.

At least
* `pack_balance_proof`
* `pack_balance_proof_update`
* `pack_reward_proof`
should be moved to `raiden-contracts`.

Perhaps
* `Signer`
* `LocalSigner`
can be moved to `raiden-contracts` too.

pirapira:
Also the `sign()` function should receive a message, not a message hash.  And then use `keccak` as the hasher when calling functions from `coincurve`.

pirapira:
https://github.com/raiden-network/raiden/pull/4284/files introduces `_data_to_sign` method.  See if it's merged and then think.

karlb:
Since packing has to be done in all cases before a signature is generated, packing can be introduced as a shared lowest common denominator. Here's a specific suggestion using the reward proof signature as an example:
* have a pack_reward_proof in the contracts repo
* import that from raiden and wrap it inside a raiden-specific pack_reward_proof function to accomodate the different conventions (CanonicalIdentifier, more detailed types)

This approach is neither the most elegant, not does it provide a large amount of code reuse. But it is a suitable stepping stone for further unifications and when we change the data that goes into the proof, we will get a mypy error instead of failing signature checks. Since debugging signatures can get hard quickly, this advantage justifies the effort of implementing this change, as far as I am concerned.

pirapira:
@karlb let's do that.  Assign yourself this issue before starting (so as to avoid duplicate work).

karlb:
I will do this for the reward proof first, since I have to touch that due to https://github.com/raiden-network/raiden-contracts/issues/1117. If that works out fine I'll expand that approach to other signatures.

karlb:
Remaining candidates are
- [ ] pack_balance_proof
- [ ] pack_balance_proof_update
- [ ] pack_withdraw

The `cooperative_settle_message` is not used by the client and `sign_one_to_n_iou` is already in use. If I'm missing anything, let me know.

karlb:
https://github.com/raiden-network/raiden/pull/4423 will be ready for review after the updating the contracts used in raiden. @pirapira Any recommendations on next steps?

pirapira:
@karlb no, I haven't looked at this closely enough.  I said something about `Signer` and `LocalSigner` above, but take them as random ideas.

[...]

PR #1042, opened 2019-06-04
Rename MAX_UINT256 into UINT256_MAX

pirapira:
because it's called that way in `raiden` repo.


pirapira:
This is the first step for #1041.

For #1041, the definition of `CanonicalIdentifier` needs to move to `raiden-contract`, and the definition uses `UINT256_MAX`.

pirapira:
Thank you!

PR #1043, opened 2019-06-04
Add a symlink to Changelog in the package

pirapira:
so CHANGELOG.md will be included in the package.

Thsi is a part of #470.

### DIFF
--- a/raiden_contracts/CHANGELOG.md
+++ b/raiden_contracts/CHANGELOG.md
@@ -1,0 +1,1 @@
+../CHANGELOG.md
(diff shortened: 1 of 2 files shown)


raiden_contracts/CHANGELOG.md
```diff
@@ -0,0 +1 @@
+../CHANGELOG.md
```
karlb: I can't see it in the diff, but I assume that this is indeed a symlink, not a file with the content `../CHANGELOG.md`.

PR #1047, opened 2019-06-04
Add an FAQ entry about block gas exceeded

### LINKED ISSUE #914: Add an FAQ entry about 'block gas exceeded'
Maybe the reason is `--gas-limit` with a too big argument.

pirapira:
This closes #914.


pirapira:
Thank you!

PR #1049, opened 2019-06-04
Assert that the expiration block number is not zero

### LINKED ISSUE #945: OneToN.claim() should make sure to change the storage
Currently, when `OneToN.claim()` is called with `expiration_block=0`, the storage of the contract might not change even when the call is successful.  This issue keeps track of preventing this possibility.

Note: since no transaction is executed in the genesis block, this condition actually holds.  Just adding an assertion should be enough.

pirapira:
because otherwise double-spend is possible.  This closes #945.


raiden_contracts/data/source/services/OneToN.sol
```diff
@@ -94,6 +94,7 @@ contract OneToN is Utils {
         if (transferable > 0) {
             // register to avoid double claiming
             settled_sessions[_key] = expiration_block;
+            assert(expiration_block != 0);
```
palango: Can we somehow access the current block number? Otherwise I'm not sure if this check improves things. Maybe make it at least `expiration_block > 0`?

pirapira: I'll say `expiration_block > 0` here.  It reads much better.

PR #1050, opened 2019-06-04
claimReward() returns a boolean

### LINKED ISSUE #980: Function signature of `claimReward()` returns bool, but the function body has no return statement
https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/data/source/services/MonitoringService.sol#L291

`MonitoringService.claimReward()`'s signature says it returns a boolean, but the function body does not contain a return statement.  The function body doesn't contain an assignment into the return variable either.

pirapira:
as the function signature suggests.

This closes #980.


palango:
Does it make sense to always return true? Maybe it would be better to have no return type?

pirapira:
@palango It's true it's redundant. I'm just not confident enough to cope with some boolean-returning functions and nothing-returning functions.

Currently we can wrap all external calls in `require(...)` like
```
require(MS.claimReward(...));
require(token.transfer(...));
...
```
An external call without `require()` stands out, and people are likelier to notice mistakes.

---

Let me try if Solidity compiler detects unused return values.

palango:
I have no good overview over the contracts, so if we return booleans everywhere that's totally fine for me.

pirapira:
I procrastinate again: https://github.com/raiden-network/raiden-contracts/issues/1075

pirapira:
@palango I think currently it's safer to always return a boolean (even when it's always true), and always check the return value.

Solidity compiler doesn't warn when we forget to check the return value of a function call, so sticking with "always check return values" is a safe convention to follow.

PR #1051, opened 2019-06-04
Make MonitoringService.reward internal

### LINKED ISSUE #981: MonitoringService.reward's visibility should be specified explicitly.
In MonitoringService, a state variable `rewards` is defined as
```
 mapping(bytes32 => Reward) rewards;
```

the visibility of this variable should be explicitly specified.

pirapira:
Probably public is a reasonable choice.

This closes #981.


raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -19,7 +19,7 @@ contract MonitoringService is Utils {
 
     // keccak256(channel_identifier, token_network_address) => Struct
     // Keep track of the rewards per channel
-    mapping(bytes32 => Reward) rewards;
+    mapping(bytes32 => Reward) public rewards;
```
palango: Is there a use case you have in mind for accessing this? I'd rather have it private otherwise.

pirapira: No, I had nothing in particular.  I'll say `private` here.

Issue #646, opened 2019-03-01
Define a new role that can raise deposit limits


pirapira:
This issue keeps track of adding a new role in TokenNetwork that can raise deposit limits.  This role can be assigned to an Ethereum address when TokenNetwork is deployed.

pirapira:
The new address should be given similarly to how the deprecator is set.

pirapira:
When I implement it naively, I hit
```
E           eth_tester.exceptions.TransactionFailed: deployment failed
```

pirapira:
The runtime codesize hasn't hit the max code size.  I guess this is a gas issue.

pirapira:
Yes, raising the block gas limit to 10M solves this particular issue.

karlb:
There are currently no plans to increase the limits on an already deployed contract.

PR #1055, opened 2019-06-04
Limit raiser

pirapira:
For #646.  Work in progress.


pirapira:
I leave it here.  #1061 is more urgent.

pirapira:
I leave this here, to focus on changes that don't change the ABI.

PR #1064, opened 2019-06-07
0.23.0 release

pirapira:
This is a usual release PR, but I had to remove some leftover of EndpointRegistry here and there.

This version brings
* cheaper "unlock" without Merkle tree construction
* more collision resistant signature check in OneToN.claim() signature (with a new MessageID).


pirapira:
@czepluch yes, that's the case. The smart contracts are only copied over.

PR #1080, opened 2019-06-20
TokenNetwork.unlock() now works with no locks

### LINKED ISSUE #1063: unlock() should go through even when `locks` contains no locks
If we remove
```
        require(locks.length > 0);
```
even when a party submits a balance proof with inconsistent `lockedamount` and `locksroot`, at least the tokens can leave the contract.  See https://github.com/raiden-network/raiden-contracts/pull/1062#discussion_r291326872

pirapira:
The new execution path allows a receiver to give back the
non-unlockable transfers back to the sender, when the receiver
has accidentially submitted an empty locksroot.

That's not supposed to happen with perfectly working clients,
but it's nice to remove one possibility where tokens are held
locked forever.

The new execution path doesn't allow any participants to
get more tokens than they deserve.

This fixes #1063.


raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -802,8 +802,6 @@ contract TokenNetwork is Utils {
         // for the external APIs
         require(channels[channel_identifier].state == ChannelState.NonExistent);
 
-        require(locks.length > 0);
```
hackaugusto: IMO this edge case should be unit tested.

pirapira: OK.  No problem adding a test.

raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -1097,6 +1115,42 @@ def test_unlock_twice_fails(
...
 
+def test_unlock_nothing(
+    web3: Web3, token_network: Contract, get_accounts: Callable, create_settled_channel: Callable
+) -> None:
+    """ unlock() should work on no pending locks """
+    (A, B) = get_accounts(2)
+    settle_timeout = 8
+
+    # Mock pending transfers data
+    pending_transfers_tree_1 = get_pending_transfers_tree(
```
pirapira: Why is it called `...1`?

Issue #401, opened 2019-01-07
Minimal code splitting for TokenNetwork - using a library for pure functions


loredanacirstea:
Parent issue: https://github.com/raiden-network/raiden-contracts/issues/215

This was done in https://github.com/raiden-network/raiden-contracts/pull/264. However, it was a big PR and it needs to be split into multiple PRs. This was already started with https://github.com/raiden-network/raiden-contracts/pull/379.

This prepares for https://github.com/raiden-network/raiden-contracts/issues/277, https://github.com/raiden-network/raiden-contracts/issues/399, https://github.com/raiden-network/raiden-contracts/issues/81

However, it needs to be assessed again wether it is really needed if only https://github.com/raiden-network/raiden-contracts/issues/277 is done within Ithaca. As, the others as scoped out or low priority.

PR #1084, opened 2019-06-21
Add error messages in require() in TokenNetwork

pirapira:
And turn on the optimizer of the Solidity compiler.  Otherwise, the runtime bytecode becomes too big for deployment.

This is a part of https://github.com/raiden-network/raiden-contracts/issues/81


hackaugusto:
What about the other solidity files?

pirapira:
> There are places missing messages, checked with

I'll see.  By the way `assert()`s don't fail.

> What about the other solidity files?

Will be kept track in #81.

pirapira:
I'll try to keep the optimizer off, by splitting away the messages into a different contract and saving the codesize.

pirapira:
@hackaugusto `assert()` is implemented using `INVALID` opcode, which cannot return a message anyway.

pirapira:
When I move the strings to a library, the runtime code becomes bigger.  So a serious split up is necessary https://github.com/raiden-network/raiden-contracts/issues/401 .

PR #1089, opened 2019-06-27
Turn an assert into if not ...: raise ...

pirapira:
because the condition might fail depending on the input to the function.

I'm expecting CodeCov to complain about this.  After seeing CodeCov complain, I'll add a test.

This is a part of #847.


pirapira:
yay!

Issue #1087, opened 2019-06-27
Implement a simple service registry


pirapira:
where the registration and removal of services are done by a single owner.

First I'll list the use cases.

pirapira:
* the owner adds a service
* the owner removes a service
* a registered service changes it's associated url or something
* another contract checks if an address is a registered service

pirapira:
1. Add an owner in ServiceRegistry
2. Add an interface for the owner to register a service
3. Add an interface for the owner to remove a service
4. Figure out what kind of metadata belongs to a service
4. Add an interface for a registered service to change some metadata

pirapira:
The specification for this is now in a private repo https://github.com/raiden-network/team/issues/475#issuecomment-508736262

pirapira:
This has been done.

PR #1090, opened 2019-06-27
Add an owner to SecretRegistry

pirapira:
At the momment the owner cannot do anything special,
but eventually it can do something special.

This is a part of #1087.


pirapira:
Exactly.

PR #1097, opened 2019-07-01
Turn an assertion into if ...: raise ...

pirapira:
The onchain contract instance might say something weird, so
the check should not be omitted when the script runs in some
optimized mode.  That's why the check should not be just an assertion.

This is a part of #847.

### DIFF
--- a/raiden_contracts/tests/test_deploy_script.py
+++ b/raiden_contracts/tests/test_deploy_script.py
@@ -87,6 +87,12 @@ def deployed_raiden_info(deployer: ContractDeployer) -> DeployedContracts:
 
 @pytest.mark.slow
 @pytest.fixture(scope="session")
+def deployed_raiden_info2(deployer: ContractDeployer) -> DeployedContracts:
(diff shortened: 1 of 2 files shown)


pirapira:
I'm expecting a CodeCov failure.  After I see it, I'll add a test.

pirapira:
How do I hit this case?  Maybe I make two sets of deployments, and create a mixed-up deployment data.

raiden_contracts/tests/test_deploy_script.py
```diff
@@ -85,6 +85,12 @@ def deployed_raiden_info(deployer: ContractDeployer) -> DeployedContracts:
     return deployer.deploy_raiden_contracts(max_num_of_token_networks=1)
 
 
+@pytest.mark.slow
+@pytest.fixture(scope="session")
+def deployed_raiden_info2(deployer: ContractDeployer) -> DeployedContracts:
```
ulope: Not a very informative name ;)

pirapira: True.  Though I just needed another one.

PR #1104, opened 2019-07-02
The owner of ServiceRegistry can add services

pirapira:
and the services can no longer register themselves.

This is a part of #1087.


pirapira:
A more interesting implementation is coming.

PR #1109, opened 2019-07-05
Checking constructor args should not be assert

pirapira:
because assertions can be omitted by optimizations.


pirapira:
This PR is a part of #847.

pirapira:
Yay!

PR #1114, opened 2019-07-08
Turn actual check from an assert into

pirapira:
an if not ...: raise ...

This is a part of #847.

I think CodeCov will complain.  Then I'll add the missing test.


pirapira:
Somehow CodeCov is not working.

PR #1116, opened 2019-07-08
ServiceRegistry (without parameter reconfiguration)

pirapira:
This is a part of #1087 .

----

### Getting and Extending the Membership

There is always a deposit amount for six-month of membership.
* a new member can join the pool for making the deposit.
* an existing member can extend their membership for six months, for the same amount of deposit

### The Change of Deposit Amounts

* whenever somebody joins or extends the membership, the price increases by 20%
* while nobody joins, the price decreases 0.5% per day.

### Terms of the Deposits

The deposit is held until the six months period.
* A new member's deposit is held for six months.
* When an existing member extends the membership,
    * the new deposit is held until the end of the extended period
    * the old deposits' conditions stay unchanged


raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,143 @@ pragma solidity 0.5.4;
 import "raiden/Token.sol";
 import "raiden/Utils.sol";
 
+contract Deposit {
+    Token public token;
+    address public owner;
```
pirapira: Perhaps rename to `withdrawer` (more consistent with the constructor argument).

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,143 @@ pragma solidity 0.5.4;
...
+        release_at = _release_at;
+        owner = _withdrawer;
+    }
+
+    function deposit(uint256 _amount) external returns (bool success) {
+        require(token.transferFrom(msg.sender, address(this), _amount));
+        return true;
+    }
+
+    function withdraw(address _to) external {
```
pirapira: Return a bool as everywhere else.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,143 @@ pragma solidity 0.5.4;
...
+    }
+}
+
+
 contract ServiceRegistry is Utils {
     Token public token;
     address public owner;
 
-    mapping(address => uint256) public deposits;  // token amount staked by the service provider
+    uint256 public last_price;
```
pirapira: Add a comment.  It's easy to miss `last_price` is a price before the bump.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,143 @@ pragma solidity 0.5.4;
...
+    uint256 public last_price_at;
+
+    mapping(address => uint256) public service_valid_till;
     mapping(address => string) public urls;  // URLs of services for HTTP access
-    address[] public service_addresses;  // list of available services (ethereum addresses)
+
+    event RegisteredService(address indexed service, uint256 valid_till, uint256 deposit_amount, Deposit deposit_contract);
```
pirapira: Add a test that sees this event.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,143 @@ pragma solidity 0.5.4;
...
 
     // @param _token_for_registration The address of the ERC20 token contract that services use for registration fees
-    constructor(address _token_for_registration) public {
+    constructor(address _token_for_registration, uint256 _initial_price) public {
         require(_token_for_registration != address(0x0));
         require(contractExists(_token_for_registration));
+        require(_initial_price >= 100);
```
pirapira: This and `min_price` below are not consistent.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,143 @@ pragma solidity 0.5.4;
...
+        // Check against overflow.
+        require(valid_till < valid_till + 180 days);
+        valid_till = valid_till + 180 days;
+        assert(valid_till > service_valid_till[msg.sender]);
+        service_valid_till[msg.sender] = valid_till;
 
-        // This also allows for MSs to deposit and use other MSs
-        deposits[msg.sender] += amount;
+        // Record the price
+        last_price = amount;
```
pirapira: I think it's better to record the already bumped price here.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,143 @@ pragma solidity 0.5.4;
...
+        if (price < min_price) {  // Maybe make this configurable too?
+            price = min_price;
+        }
+        return price;
+
+    }
+
+    function current_price() public view returns (uint256) {
+        // The price increased for the last event.
+        uint256 price = last_price * 6 / 5; // Maybe make this configurable too?
```
pirapira: No need to bump the price if `last_price` already has the bumped price.

raiden_contracts/tests/test_deploy_script.py
```diff
@@ -130,7 +131,9 @@ def deployed_service_info(
     deployer: ContractDeployer, token_address: HexAddress
 ) -> DeployedContracts:
     return deployer.deploy_service_contracts(
-        token_address=token_address, user_deposit_whole_balance_limit=DEPOSIT_LIMIT
+        token_address=token_address,
+        user_deposit_whole_balance_limit=DEPOSIT_LIMIT,
+        initial_service_deposit=SERVICE_DEPOSIT // 2,
```
pirapira: `data_x.y.z` that contains older service contracts should be removed.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,146 @@ pragma solidity 0.5.4;
 import "raiden/Token.sol";
 import "raiden/Utils.sol";
 
+contract Deposit {
```
pirapira: A few sentences will make the contract more understandable.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,146 @@ pragma solidity 0.5.4;
...
+        withdrawer = _withdrawer;
+    }
+
+    function deposit(uint256 _amount) external returns (bool success) {
+        require(token.transferFrom(msg.sender, address(this), _amount));
+        return true;
+    }
+
+    function withdraw(address _to) external returns (bool success) {
+        uint256 sent_amount = token.balanceOf(address(this));
```
pirapira: `balance` is a better name than `sent_amount`.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,146 @@ pragma solidity 0.5.4;
...
         owner = msg.sender;
+
+        // Set up the price and the set price timestamp
+        set_price = _initial_price;
+        set_price_at = now;
     }
 
-    function deposit(uint amount) public {
-        require(amount > 0);
+    function deposit(uint _limit_amount) public {
```
pirapira: Add docstring.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,146 @@ pragma solidity 0.5.4;
...
+
+    mapping(address => uint256) public service_valid_till;
     mapping(address => string) public urls;  // URLs of services for HTTP access
-    address[] public service_addresses;  // list of available services (ethereum addresses)
+
+    event RegisteredService(address indexed service, uint256 valid_till, uint256 deposit_amount, Deposit deposit_contract);
 
     // @param _token_for_registration The address of the ERC20 token contract that services use for registration fees
-    constructor(address _token_for_registration) public {
+    constructor(address _token_for_registration, uint256 _initial_price) public {
```
pirapira: Add docstring.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,157 @@ pragma solidity 0.5.4;
...
+    }
+
+    // In order to make a deposit, transfer the ERC20 token into this contract.
+    // If you transfer a wrong kind of ERC20 token or ETH into this contract,
+    // these tokens will be lost forever.
+
+    /// @notice Withdraws the tokens that have been deposited.
+    /// Only `withdrawer` can call this.
+    /// @param _to The address where the withdrawn tokens should go.
+    function withdraw(address _to) external returns (bool success) {
```
pirapira: Add a test that calls `withdraw`.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,171 @@ pragma solidity 0.5.4;
...
+
+        uint256 price = _set_price * P / Q;
+
+        // Not allowing a price smaller than 1000.
+        // Once it's too low it's too low forever.
+        if (price < min_price) {  // Maybe make this configurable too?
+            price = min_price;
+        }
+        return price;
+
```
pirapira: This empty line is weird.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,170 @@ pragma solidity 0.5.4;
...
     Token public token;
     address public owner;
 
-    mapping(address => uint256) public deposits;  // token amount staked by the service provider
+    // After a price is set to set_price at timestamp set_price_at,
+    // the price decays according to decayed_price().
+    uint256 public set_price;
+    uint256 public set_price_at;
+
+    // Once the price is too low, 20% increase cannot not move the price upwards.
```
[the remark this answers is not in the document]
pirapira: That describes the code.  My comment was doing something else.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,170 @@ pragma solidity 0.5.4;
...
+        view returns (uint256) {
+        // We are here trying to approximate some exponential decay.
+        // exp(- X / A) where
+        //   X is the number of seconds since the last price change
+        //   A is the decay constant (A = 200 days correspnods to 0.5% decrease per day)
+
+        // exp(- X / A) ~~ P / Q where
+        //   P = 24 A^4
+        //   Q = 24 A^4 + 24 A^3X + 12 A^2X^2 + 4 AX^3 + X^4
+        // Note: swap P and Q, and then think about the Taylor expansion.
```
karlb: Nice approach, works great!

raiden_contracts/deploy/__main__.py
```diff
@@ -208,6 +208,12 @@ def raiden(
     type=int,
     help="Maximum amount of tokens deposited in UserDeposit",
 )
+@click.option(
+    "--initial_service_deposit",
+    required=True,
+    type=int,
+    help="Maximum amount of tokens deposited in UserDeposit",
```
karlb: This help text is copy/pasted from above.

pirapira: :man_facepalming: fixed.

raiden_contracts/tests/test_service_registry.py
```diff
@@ -2,49 +2,80 @@
...
+    custom_token.functions.mint(SERVICE_DEPOSIT).call_and_transact({"from": A})
+    custom_token.functions.approve(service_registry.address, SERVICE_DEPOSIT).call_and_transact(
+        {"from": A}
+    )
 
     # happy path
-    service_registry.functions.deposit(10).call_and_transact({"from": A})
-    assert service_registry.functions.deposits(A).call() == 10
-    assert custom_token.functions.balanceOf(A).call() == 0
+    old_balance = custom_token.functions.balanceOf(A).call()
```
karlb: There's a second happy path when extending the validity of an existing slot. At least a rough test for that case would make sense.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
...
+
+        uint256 X = _seconds_passed;
+
+        if (X >= 2 ** 64) { // The computation below overflows.
+            return min_price;
```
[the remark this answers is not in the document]
pirapira: Thanks for finding this.  I'll go for `2 ** 60` (which is still 36 billion years).

```
>>> A = 200
>>> X = 2 ** 60 - 1
>>> P = 24 * (A ** 4)
>>> Q = P + 24*(A**3)*X + 12*(A**2)*(X**2) + 4*A*(X**3) + X**4
>>> Q < (2 ** 256 - 1)
True
```

raiden_contracts/tests/test_service_registry.py
```diff
@@ -2,49 +2,80 @@
...
+def test_deposit_contract(
+    get_deposit_contract: Callable, custom_token: Contract, get_accounts: Callable
+) -> None:
+    (A,) = get_accounts(1)
+    custom_token.functions.mint(100).call_and_transact({"from": A})
+    depo = get_deposit_contract([custom_token.address, 0, A])
+    custom_token.functions.transfer(depo.address, 100).call_and_transact({"from": A})
+    assert custom_token.functions.balanceOf(A).call() == 0
+    assert custom_token.functions.balanceOf(depo.address).call() == 100
+    depo.functions.withdraw(A).call_and_transact({"from": A})
```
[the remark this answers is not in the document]
pirapira: I'll add that test.

raiden_contracts/tests/test_service_registry.py
pirapira: I agree.  I'll do that as a continuation of this case.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
...
+    /// @notice Withdraws the tokens that have been deposited.
+    /// Only `withdrawer` can call this.
+    /// @param _to The address where the withdrawn tokens should go.
+    function withdraw(address _to) external returns (bool success) {
+        uint256 balance = token.balanceOf(address(this));
+        require(msg.sender == withdrawer);
+        require(now >= release_at);
+        require(balance > 0);
+        require(token.transfer(_to, balance));
+        return true;
```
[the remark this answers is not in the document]
pirapira: I think currently it's safer to always return a boolean (even when it's always true), and always check the return value.

Solidity compiler doesn't warn when we forget to check the return value of a function call, so sticking with "always check return values"  is a safe convention to follow.

----

I'll document this in the contribution guide.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
 import "raiden/Token.sol";
 import "raiden/Utils.sol";
 
+contract Deposit {
```
palango: Why is this it's own contract? Couldn't you also model this as a struct inside the `ServiceRegistry` contract?

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
...
+
+    mapping(address => uint256) public service_valid_till;
     mapping(address => string) public urls;  // URLs of services for HTTP access
-    address[] public service_addresses;  // list of available services (ethereum addresses)
+
+    event RegisteredService(address indexed service, uint256 valid_till, uint256 deposit_amount, Deposit deposit_contract);
 
     // @param _token_for_registration The address of the ERC20 token contract that services use for registration fees
-    constructor(address _token_for_registration) public {
+    constructor(address _token_for_registration, uint256 _initial_price) public {
```
palango: Should `min_price` be settable as well?

pirapira: This follows in the next PR.  `min_price`, `6/5` bumping ratio and `200 days` will be configurable.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
...
+        uint256 amount = current_price();
+        require(_limit_amount >= amount, "not enough limit");
+
+        // Extend the service position.
+        uint256 valid_till = service_valid_till[msg.sender];
+        if (valid_till < now) { // a first time joiner or an expired service.
+            valid_till = now;
+        }
+        // Check against overflow.
+        require(valid_till < valid_till + 180 days);
```
palango: maybe add an error message here

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
...
+    uint256 public set_price_at;
+
+    // Once the price is at min_price, it can't decay further.
+    uint256 constant min_price = 1000;
+
+    mapping(address => uint256) public service_valid_till;
     mapping(address => string) public urls;  // URLs of services for HTTP access
-    address[] public service_addresses;  // list of available services (ethereum addresses)
+
+    event RegisteredService(address indexed service, uint256 valid_till, uint256 deposit_amount, Deposit deposit_contract);
```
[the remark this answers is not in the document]
pirapira: Yes.  And I'll add a comment explaining the log arguments.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
...
+
+        // Extend the service position.
+        uint256 valid_till = service_valid_till[msg.sender];
+        if (valid_till < now) { // a first time joiner or an expired service.
+            valid_till = now;
+        }
+        // Check against overflow.
+        require(valid_till < valid_till + 180 days);
+        valid_till = valid_till + 180 days;
+        assert(valid_till > service_valid_till[msg.sender]);
```
palango: I'm not sure about low-level differences, but why do you use `assert`  here?

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
...
+    /// @param _release_at The timestap after which the withdrawer can withdraw the deposit.
+    /// @param _withdrawer The address that can withdraw the deposit after the release time.
+    constructor(address _token, uint256 _release_at, address _withdrawer) public {
+        token = Token(_token);
+        // Don't care even if it's in the past.
+        release_at = _release_at;
+        withdrawer = _withdrawer;
+    }
+
+    // In order to make a deposit, transfer the ERC20 token into this contract.
```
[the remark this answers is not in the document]
pirapira: Unfortunately, contracts cannot refuse incoming ERC20 tokens.

In the beginning, I had a `deposit()` function, but it did nothing useful.  So I removed it.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
...
+        set_price = amount * 6 / 5;
+        set_price_at = now;
+
+        // Move the deposit in a new Deposit contract.
+        require(token.transferFrom(msg.sender, address(this), amount));
+        Deposit depo = new Deposit(address(token), valid_till, msg.sender);
+        require(token.transfer(address(depo), amount));
+
+        // Fire event
+        emit RegisteredService(msg.sender, valid_till, amount, depo);
```
palango: Above you write that all functions should return `true`, so this one as well?

pirapira: I believe so.  I'll add `return true`.

raiden_contracts/data/source/services/ServiceRegistry.sol
pirapira: Because I know the assert is always true.  So I use `assert` and then the SMT solver in `solc` tries to find a counterexample.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
...
+    /// Only a currently registered service can call this successfully.
+    /// @param new_url The new URL string to be stored.
     function setURL(string memory new_url) public {
+        require(now < service_valid_till[msg.sender]);
         require(bytes(new_url).length != 0, "new url is empty string");
-        if (bytes(urls[msg.sender]).length == 0) {
-            service_addresses.push(msg.sender);
-        }
         urls[msg.sender] = new_url;
     }
```
palango: Same here, should this return `true`?

raiden_contracts/data/source/services/ServiceRegistry.sol
pirapira: Yes, I will.

raiden_contracts/data/source/services/ServiceRegistry.sol
pirapira: That's also true.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
...
+    /// @notice Calculates the decreased price after a number of seconds.
+    /// @param _set_price The initial price.
+    /// @param _seconds_passed The number of seconds passed since the initial
+    /// price was set.
+    function decayed_price(uint256 _set_price, uint256 _seconds_passed) public
+        view returns (uint256) {
+        // We are here trying to approximate some exponential decay.
+        // exp(- X / A) where
+        //   X is the number of seconds since the last price change
+        //   A is the decay constant (A = 200 days correspnods to 0.5% decrease per day)
```
palango: s/correspnods/corresponds

raiden_contracts/data/source/services/ServiceRegistry.sol
karlb: To say it in a different way: `assert` should only fail for bugs in the contract, `require` is for all other cases.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
...
+        if (X >= 2 ** 64) { // The computation below overflows.
+            return min_price;
+        }
+
+        uint256 A = decay_constant;
+
+        uint256 P = 24 * (A ** 4);
+        uint256 Q = P + 24*(A**3)*X + 12*(A**2)*(X**2) + 4*A*(X**3) + X**4;
+
+        uint256 price = _set_price * P / Q;
```
palango: Once you make `A` configurable, you might have to check for an overflow here as well.

raiden_contracts/deploy/__main__.py
```diff
@@ -208,6 +208,12 @@ def raiden(
     type=int,
     help="Maximum amount of tokens deposited in UserDeposit",
 )
+@click.option(
+    "--initial_service_deposit",
```
palango: I think `initial_service_price` would fit better

raiden_contracts/deploy/contract_verifier.py
```diff
@@ -349,13 +353,13 @@ def _verify_service_registry_deployment(
     """ Check an onchain deployment of ServiceRegistry and constructor arguments """
     if to_checksum_address(service_registry.functions.token().call()) != token_address:
         raise RuntimeError("service_registry has a wrong token address")
+    if len(constructor_arguments) != 2:
+        raise RuntimeError(
+            "service_registry was deployed with a wrong number of constructor arguments"
```
palango: Above you use CamelCase for the contract name

raiden_contracts/data/source/services/ServiceRegistry.sol
pirapira: Yes, everything could be in `ServiceRegistry` and that would be gas-efficient.

The current choice follows security-in-depth.  Even if `ServiceRegistry` is compromized, the deposits made earlier would not be affected.

Is this conceptual security worth more than the additional gas?  I don't know for sure, but the deposits are rarer and bigger than usual Raiden transfers.

raiden_contracts/data/source/services/ServiceRegistry.sol
pirapira: Yes, the next PR will contain lots of those checks.  I hope nobody needs the price decaying over billions of years.

raiden_contracts/deploy/__main__.py
pirapira: Now I hear it, I agree.   Launching `raiden deploy ... --initial_service_deposit 30000` sounds like you're already making a deposit, but actually you're only setting a price.

raiden_contracts/deploy/contract_verifier.py
pirapira: I'll fix this together with similar occurrences above and below.

raiden_contracts/data/source/services/ServiceRegistry.sol
palango: That argument makes sense to me, but I don't have much experience. The question was aimed at understanding the trade-off better!

raiden_contracts/tests/test_service_registry.py
```diff
@@ -2,49 +2,80 @@
...
+from raiden_contracts.tests.utils.constants import EMPTY_ADDRESS, SERVICE_DEPOSIT
 
 
-def test_owner_of_service_registry(service_registry: Contract) -> None:
-    """ The owner of ServiceRegistry should be the deployer """
-    assert service_registry.functions.owner().call() == CONTRACT_DEPLOYER_ADDRESS
+def test_deposit_contract(
+    get_deposit_contract: Callable, custom_token: Contract, get_accounts: Callable
+) -> None:
+    (A,) = get_accounts(1)
```
palango: nit: You don't need braces here

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -3,49 +3,167 @@ pragma solidity 0.5.4;
...
+    }
+
+    // In order to make a deposit, transfer the ERC20 token into this contract.
+    // If you transfer a wrong kind of ERC20 token or ETH into this contract,
+    // these tokens will be lost forever.
+
+    /// @notice Withdraws the tokens that have been deposited.
+    /// Only `withdrawer` can call this.
+    /// @param _to The address where the withdrawn tokens should go.
+    function withdraw(address _to) external returns (bool success) {
```
palango: The contract could self-destruct after a successful withdraw.

raiden_contracts/tests/test_service_registry.py
pirapira: I like it without the braces, but it's with the braces everywhere in the codebase.  For now I just follow the local convention.

raiden_contracts/data/source/services/ServiceRegistry.sol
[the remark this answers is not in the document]
pirapira: I've written this in `CONTRIBUTING.md`.

Issue #1117, opened 2019-07-09
Use `non_closing_signature` in reward proof


karlb:
This prevents the reward proof to be swapped out independently of the other data in monitoring requests. This will prevent attacks such as https://github.com/raiden-network/raiden-services/issues/418.

- [x] update contract (#1119)
- [x] update client (https://github.com/raiden-network/raiden/pull/4364)
- [x] check/update MS (https://github.com/raiden-network/raiden-services/pull/446)
- [x] update spec (https://github.com/raiden-network/spec/pull/239)

PR #1119, opened 2019-07-09
Include `non_closing_signature` in reward proof

karlb:
This prevents attacks which are based on the fact that two signatures in
a monitor request are independent. It also makes it easier to reason
about the security properties of MRs.

See https://github.com/raiden-network/raiden-contracts/issues/1117.


raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -98,6 +98,7 @@ contract MonitoringService is Utils {
         uint256 reward_amount,
         uint256 nonce,
         address monitoring_service_address,
+        bytes memory non_closing_signature,
```
pirapira: Please add
```
/// @param non_closing_signature <description>
```
above.

karlb: Done. Didn't we have a linter at some point that checked for these docstrings?

pirapira: No, I don't remember.  But if you find one, just bring it in.

karlb: Ahh, it's in the `compile_contracts`, not in `make lint`. I probably managed to compile the contracts by running the tests instead of calling `compile_contracts`.

pirapira:
@karlb I'm seeing
```
E       fixture 'create_reward_proof' not found
```

PR #1127, opened 2019-07-16
Make mypy read all config from `setup.cfg`

karlb:
This allows running mypy without any arguments, which makes more tooling
work out of the box.

`--check-untyped-defs` is not needed anymore, since untyped defs are disallowed.


pirapira:
Which change directs `mypy` to read `setup.cfg`?

karlb:
> Which change directs `mypy` to read `setup.cfg`?

No change is necessary for that. Mypy does it by default.

PR #1132, opened 2019-07-22
Admin can control parameters of SecretRegistry

pirapira:
This PR allows ServiceRegistry's controller to change
* the speed of price decrease
* the ratio of price bump when a deposit is made
* and the duration of service registrations

The deployer can choose to have a controller or no controllers.


raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -43,16 +43,149 @@ contract Deposit {
...
+        set_price = current_price();
+        set_price_at = now;
+    }
+
+    function set_price_bump_parameters(
+            uint256 _price_bump_numerator,
+            uint256 _price_bump_denominator
+    ) private {
+        require(_price_bump_denominator > 0, "divide by zero");
+        price_bump_numerator = _price_bump_numerator;
```
karlb: A `require(_price_bump_numerator > _price_bump_denominator)` could prevent accidental reversal of those two numbers.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -43,16 +43,149 @@ contract Deposit {
...
+        decay_constant = _decay_constant;
+    }
+
+    function set_registration_duration(uint256 _registration_duration) private {
+        // No checks.  Even allowing zero (when no new registrations are possible).
+        registration_duration = _registration_duration;
+    }
+
+
+    /// @notice The amount of deposits for registration or extension.
```
karlb: "amount of deposits" sounds like you need multiple deposits, not a deposit of a specific size. Maybe "amount to deposit"?

raiden_contracts/tests/fixtures/service_registry_fixtures.py
```diff
@@ -3,11 +3,48 @@
...
 
 @pytest.fixture(scope="session")
 def service_registry(deploy_tester_contract: Callable, custom_token: Contract) -> Contract:
     return deploy_tester_contract(
-        CONTRACT_SERVICE_REGISTRY, [custom_token.address, 3000 * (10 ** 18)]
+        CONTRACT_SERVICE_REGISTRY,
+        [
+            custom_token.address,
+            CONTRACT_DEPLOYER_ADDRESS,
+            3000 * (10 ** 18),
```
karlb: I usually prefer to write these as `int(3000e18)`, but that's a matter of taste.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -43,16 +43,149 @@ contract Deposit {
...
+        if (X >= 2 ** 60) { // The computation below overflows.
+            return min_price;
+        }
+
+        uint256 A = decay_constant;
+
+        uint256 P = 24 * (A ** 4);
+        uint256 Q = P + 24*(A**3)*X + 12*(A**2)*(X**2) + 4*A*(X**3) + X**4;
+
+        uint256 price = _set_price * P / Q;
```
karlb: If both `decay_constant` and `_set_price` are very large, this will overflow. Practically, this can't be triggered by a user of the contract and won't happen unless the controller sets nonsensical values.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -43,16 +43,149 @@ contract Deposit {
...
+    // The duration of service registration/extension in seconds
+    uint256 public registration_duration = 180 days;
+
+    function change_parameters(
+            uint256 _price_bump_numerator,
+            uint256 _price_bump_denominator,
+            uint256 _decay_constant,
+            uint256 _min_price,
+            uint256 _registration_duration
+    ) public onlyController {
```
palango: This is an external function call and thus should return a `bool`.

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -43,16 +43,149 @@ contract Deposit {
...
+        }
+
+        uint256 A = decay_constant;
+
+        uint256 P = 24 * (A ** 4);
+        uint256 Q = P + 24*(A**3)*X + 12*(A**2)*(X**2) + 4*A*(X**3) + X**4;
+
+        uint256 price = _set_price * P / Q;
+
+        // Not allowing a price smaller than 1000.
```
palango: It's now variable, so s/1000/min_price

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -43,16 +43,149 @@ contract Deposit {
...
+        // Once it's too low it's too low forever.
+        if (price < min_price) {  // Maybe make this configurable too?
```
palango: This comment is obsolete

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -43,16 +43,149 @@ contract Deposit {
...
     // the price decays according to decayed_price().
     uint256 public set_price;
     uint256 public set_price_at;
 
+    /// The amount of time (in seconds) till the price decreases to roughly 1/e.
+    uint256 public decay_constant = 200 days;
+
     // Once the price is at min_price, it can't decay further.
-    uint256 constant min_price = 1000;
+    uint256 public min_price;
```
palango: why not have a default here, bot for all other values?

raiden_contracts/data/source/services/ServiceRegistry.sol
pirapira: That's a good idea.  I think that's a likely mistake.

raiden_contracts/data/source/services/ServiceRegistry.sol
pirapira: OK.  I also think "amount to deposit" is less ambiguous.

raiden_contracts/tests/fixtures/service_registry_fixtures.py
pirapira: I'm not a seasoned Python user so I follow your taste here.

raiden_contracts/data/source/services/ServiceRegistry.sol
pirapira: I'll limit the `decay_constant` at `2 ** 60` or something.

I'm thinking whether or not to code a hard-limit on the price.  This complicates code for an unlikely scenario, where things have already gone crazy.

raiden_contracts/data/source/services/ServiceRegistry.sol
pirapira: Yes, it's hard for me to follow what I say.

raiden_contracts/data/source/services/ServiceRegistry.sol
pirapira: Ah, the `decay_constant` is already limited at `2 ** 60`.

But `price_bump_numerator` was not capped.

The price limit of 2 ** 60 would be way too low, but 2 ** 128 looks ok.

pirapira: The overflow is closer than I thought.

I think the price cap at `2 ** 90` and the decay_constant cap at `2 ** 40` are reasonable.

pirapira: I added those limits.  I still want to add tests that try to cross the limit.

pirapira:
I'm adding more tests, which set out-of-limit parameters.

pirapira:
@karlb I added the test with big numbers.  Thanks.

PR #1136, opened 2019-07-23
Add a deprecation switch to ServiceRegistry

pirapira:
Only the controller can turn on the deprecation switch. After it's turned on, the ServiceRegistry contract does not accept new deposits.  The already made deposits are not affected by the switch.


raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -236,6 +244,8 @@ contract ServiceRegistry is Utils, ServiceRegistryConfigurableParameters {
     // The call fails if the current price is higher (this is always possible
     // when other parties have just called `deposit()`).
     function deposit(uint _limit_amount) public returns (bool _success) {
+        require(! deprecation_switch, "this contract was deprecated");
```
ulope: This is nitpicky and a taste question, but I always prefer to avoid double negation and inverted logic.

e.g. in this case naming the variable `registry_active` (or similar) with a default of `true` would make the require clearer: `require(registry_active, "registry is deprecated");`

pirapira: For now, I keep this aligned with the switch of TokenNetwork.

PR #1140, opened 2019-07-25
List addresses that have ever made deposits

### LINKED ISSUE #1139: Add a list of addresses that have ever made a deposit in ServiceRegistry
because sometimes the client needs to choose a random service.

Given the list of addresses that have ever made a deposit, the client can if each address is still a registered service.

Perhaps there can be a better way to manage a finite set onchain in a more precise way, but as the first approximation, an append-only list will be sufficient.

pirapira:
When I was integrating `raiden-contrtacts 0.26.0` with the Raiden client, I realized that there must be a way for the client to look up all registered services in ServiceRegistry contract.  This PR provides a poor man's solution to this problem, namely, a list of all addresses that have ever made a deposit.  The client can then query about any address in the list.

This closes #1139.


raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -249,8 +253,13 @@ contract ServiceRegistry is Utils, ServiceRegistryConfigurableParameters {
         uint256 amount = current_price();
         require(_limit_amount >= amount, "not enough limit");
 
+        bool new_joiner = false;
+
         // Extend the service position.
         uint256 valid_till = service_valid_till[msg.sender];
+        if (valid_till == 0) { // a first time joiner
+            new_joiner = true;
```
karlb: Why not just do a `ever_made_deposits.push(msg.sender);` here and skip the `new_joiner` variable completely?

raiden_contracts/data/source/services/ServiceRegistry.sol
```diff
@@ -280,12 +294,17 @@ contract ServiceRegistry is Utils, ServiceRegistryConfigurableParameters {
...
-    function setURL(string memory new_url) public returns (bool _success)  {
+    function setURL(string memory new_url) public returns (bool _success) {
         require(now < service_valid_till[msg.sender], "registration expired");
         require(bytes(new_url).length != 0, "new url is empty string");
         urls[msg.sender] = new_url;
         return true;
     }
+
+    /// A getter function for seeing the length of ever_made_deposits array.
+    function ever_made_deposits_len() public view returns (uint256 _len) {
```
karlb: It's really weird that solidity does not provide this for public arrays itself.

raiden_contracts/data/source/services/ServiceRegistry.sol
pirapira: That sounds better.  I'll do this.

raiden_contracts/data/source/services/ServiceRegistry.sol
pirapira: I agree.  It's weird.

PR #1146, opened 2019-07-29
Rename balance_proof_update into balance_proof_countersign

pirapira:
The function was only for updateBalanceProof() method of TokenNetwork
but it's now also used for closeChannel() method.

The new name "countersign" suggests it's a second signature on a balance
proof, and it's what the function is about.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Documents changes that result in:
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


pirapira:
I came up with a better name `signed_balance_proof`.

pirapira:
This is not a good idea. We need to differentiate closing and updating.

Issue #1086, opened 2019-06-27
Make ContractVerifyer useful in the proxies


pirapira:
Currently `ContractVerifier` exposes interfaces for 1) checking all Raiden contract deployments and 2) checking all service contract deployments.  However, in `raiden` client code, each proxy only needs to check one onchain existence.  In the constructor of proxies, we currently need to check the onchain bytecode in an ad-hoc way.  It's better to reuse the code in `ContractVerifier`.  This issue keeps track of extending ContractVerifier so it can check the onchain deployment of a single contract.

pirapira:
This must be implemented in order to maintain the Python code coverage in https://github.com/raiden-network/raiden-contracts/pull/1148 .

pirapira:
This turned out not necessary for now because CodeCov was down for a while.

PR #1148, opened 2019-07-29
MonitoringService now works only with a particular TokenNetworkRegistry

### LINKED ISSUE #1141: Only allow certain TokenNetwork addresses in MonitoringService contract
MonitoringService contract should work only with TokenNetworks that are registered in a particular TokenNetworkRegistry contract.  This issue keeps track of implementing this feature.

pirapira:
so it can see which token contract is registered there.

This closes #1141.


pirapira:
I still need to add a test where a service tries to use MonitoringService contract for an unrelated TokenNetwork.

karlb:
Why do we need this? Wouldn't it be sufficient if MSs just ignore MR for non registered token networks?

raiden_contracts/tests/test_monitoring_service.py
```diff
@@ -268,6 +268,34 @@ def test_monitor(
...
+    A, B = monitor_data["participants"]
+
+    # wait until MS is allowed to monitor
+    token_network_in_another_token_network_registry.web3.testing.mine(
+        monitor_data["first_allowed"] - web3.eth.blockNumber
+    )
+
+    # monitor() call fails because the TokenNetwork is not registered on the
+    # supposed TokenNetworkRegistry
+    with pytest.raises(TransactionFailed):
```
karlb: Now that we have messages on our requires, we could start checking that the tests fail for the right reason. Since we don't measure the coverage of our solidity code, we could have a wrong require and not notice it in any way if our transaction can also fail for a different reason.

pirapira:
> Why do we need this? Wouldn't it be sufficient if MSs just ignore MR for non registered token networks?

Here we don't trust MSs.  They might try to use the MS smart contract for other token networks.  This PR is against this (this is a relatively new requirement: see https://github.com/raiden-network/team/issues/516#issue-471166518).

raiden_contracts/tests/test_monitoring_service.py
pirapira: I agree.  I'll figure out how to check the error messages generated by `require()` messages.

pirapira:
The not-yet-covered code is hard to reach.  https://codecov.io/gh/raiden-network/raiden-contracts/compare/d97b8f66132e0da58076637719f2d035e7fe6b99...1393e1fde114d5cf4659a5a9db82f124ba99b890/diff#D1-x308

We need to expose the verification of each service contract (currently there is a single function that checks the validity of all service contracts).

pirapira:
This can be done in https://github.com/raiden-network/raiden-contracts/issues/1086 .

raiden_contracts/tests/test_monitoring_service.py
pirapira: This has been done.  https://github.com/raiden-network/raiden-contracts/issues/1158 keeps track of doing the same for all other available `require()` messages.

PR #1150, opened 2019-07-30
Differentiate close and update signature checks

### LINKED ISSUE #1149: The signatures for closeChannel() and updateBalanceProof() should use different message IDs
Currently, these two calls have exactly the same signature validation with the same message IDs, but since the messages have different message IDs, the signature validations should use different message IDs.

pirapira:
with different MessageTypeId's.  This closes #1149.


raiden_contracts/utils/proofs.py
```diff
@@ -44,6 +44,29 @@ def pack_balance_proof(
     )
 
 
+def pack_balance_proof_close_message(
```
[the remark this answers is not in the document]
pirapira: I'll add `msg_type` as an argument.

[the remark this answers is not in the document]
pirapira: I did it here, but I'm hearing a want for more thorough refactoring.

pirapira:
> $ pipdeptree --warn fail

is now complaining.

raiden_contracts/tests/fixtures/channel.py
```diff
@@ -674,11 +676,46 @@ def get(
...
+    return get
+
+
+@pytest.fixture(scope="session")
+def create_balance_proof_updating_countersignature(
+    token_network: Contract, get_private_key: Callable
+) -> Callable:
+    def get(
+        participant: HexAddress,
+        channel_identifier: int,
```
pirapira: This function can still get `msg_type` as a new argument, and one of the similar two functions can be removed.

raiden_contracts/tests/unit/test_unit_signatures.py
```diff
@@ -130,38 +131,121 @@ def test_recover_address_from_balance_proof_update(
         channel_identifier, A, other_token_network=other_token_network
     )
     balance_proof_update_signature = create_balance_proof_countersignature(
-        B, channel_identifier, *balance_proof, other_token_network=other_token_network
+        B,
+        channel_identifier,
+        MessageTypeId.BALANCE_PROOF_UPDATE,
+        *balance_proof,
+        other_token_network=other_token_network,
```
palango: why just use one kwarg?

pirapira: It was like this before the PR.

I can go all kwarg.  Then I have to unpack `balance_proof` (but I did so somewhere else).

I can go all positional, but that's a bit worse than the current situation.

pirapira: I'm going all kwarg here.  Eventually `balance_proof` here must become a dictionary, so `**balance_proof` can be used.

pirapira: To be tracked in https://github.com/raiden-network/raiden-contracts/issues/1167

Issue #1147, opened 2019-07-29
MonitoringService and OneToN should check if the service is registered


pirapira:
This issue keeps track of making sure that the MonitoringService and OneToN contracts work only for service providers that are registered in the ServiceRegistry contract.

pirapira:
MonitoringService.monitor() should work only for a registered service provider.  MonitoringService.claimReward() can be called by anybody (maybe by a service provider whose registration has expired).

--------

The other way around doesn't make sense.  A random service provider can call `monitor()` many times, and then they can register later when they want to claim the reward.

PR #1151, opened 2019-07-30
OneToN contract no longer works with unregistered services

pirapira:
This is a part of #1147.


raiden_contracts/tests/test_print_gas.py
```diff
@@ -260,6 +260,15 @@ def print_gas_monitoring_service(
...
 
+    # B registers itself as a service provider
+    deposit = service_registry.functions.currentPrice().call()
+    custom_token.functions.mint(deposit).call_and_transact({"from": B})
+    custom_token.functions.approve(service_registry.address, deposit).call_and_transact(
+        {"from": B}
+    )
+    service_registry.functions.deposit(deposit).call_and_transact({"from": B})
+    assert service_registry.functions.hasValidRegistration(B).call()
+
```
pirapira: This blob is copied and pasted four times.  I should turn this into a function.

raiden_contracts/tests/test_one_to_n.py
```diff
@@ -119,17 +121,57 @@ def test_claim(
...
+    deposit_to_udc: Callable,
+    get_accounts: Callable,
+    get_private_key: Callable,
+    web3: Web3,
+) -> None:
+    """OneToN contract should not work for an unregistered service provider."""
+    (A, B) = get_accounts(2)
+    deposit_to_udc(A, 30)
+
+    # happy case
```
palango: why is this the happy case?

raiden_contracts/tests/test_one_to_n.py
```diff
@@ -119,17 +121,57 @@ def test_claim(
...
+    amount = 10
+    expiration = web3.eth.blockNumber + 2
+    chain_id = int(web3.version.network)
```
palango: this should be `web3.eth.chainId` after https://github.com/raiden-network/raiden-contracts/pull/1153

raiden_contracts/tests/test_print_gas.py
```diff
@@ -260,6 +262,15 @@ def print_gas_monitoring_service(
     )
     service_registry.functions.deposit(SERVICE_DEPOSIT).call_and_transact({"from": MS})
 
+    # B registers itself as a service provider
+    deposit = service_registry.functions.currentPrice().call()
```
palango: Maybe you can refactor this out and use it in `create_service_account` as well to reduce duplication.

raiden_contracts/tests/test_one_to_n.py
pirapira: No, I copied and pasted the happiness but left the unhappy case only.

raiden_contracts/tests/test_print_gas.py
pirapira: Turns out these lines can simply be removed.  It's all done in `create_service_account` above.

raiden_contracts/tests/test_print_gas.py
```diff
@@ -316,11 +318,22 @@ def print_gas_one_to_n(
...
     print_gas: Callable,
+    service_registry: Contract,
+    custom_token: Contract,
 ) -> None:
     """ Abusing pytest to print gas cost of OneToN functions """
     (A, B) = get_accounts(2)
     deposit_to_udc(A, 30)
 
+    # B registers itself as a service provider
+    deposit = service_registry.functions.currentPrice().call()
```
palango: This still seems duplicated:
- https://github.com/raiden-network/raiden-contracts/pull/1151/files#diff-a0ec52e4288fe8c7b94e7d1fc9704dc6R397
- https://github.com/raiden-network/raiden-contracts/pull/1151/files#diff-7ecca76ac1d6a1de743191515e65a0e0R62

Not sure what your general approach is in this repo, as it is only 3 lines...

pirapira: The general approach is to reuse the fixture.  A fix is coming in this PR.

pirapira: I've removed this part.

There is something similar in the same file, but there the duplication is harder to remove.  I need the gas consumption of the deposit transaction.

PR #876, opened 2019-04-16
Upgrade eth-tester to 0.1.0-beta.39

pirapira:
This closes https://github.com/raiden-network/raiden-contracts/issues/868

PR #1153, opened 2019-08-07
Update web3.py, eth-tester

palango:
Next try at what #876 was doing.

There is still a test failure, but I'll report that upstream.


palango:
Question posted here: https://github.com/ethereum/web3.py/issues/1412

PR #1160, opened 2019-08-13
Revert web3 py

pirapira:
This reverts #1153.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ dist: xenial
 sudo: true
 language: python
 python:
-  - '3.6'
+  - '3.7'
 
 env:
   global:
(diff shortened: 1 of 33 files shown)


.travis.yml
```diff
@@ -2,7 +2,7 @@ dist: xenial
 sudo: true
 language: python
 python:
-  - '3.7'
+  - '3.6'
```
palango: Should we leave this?

requirements-dev.txt
```diff
@@ -13,9 +13,11 @@ flake8-tuple>=0.2.13
 bump2version
 hypothesis==3.44.14
 requests
-mypy==0.720
+mypy==0.670
```
palango: might also be nice to keep this up to date

LefterisJP: +1 ... we also use this mypy version in the client.

.travis.yml
pirapira: Yes, changed to 3.7.  I also changed another 3.6 to 3.7.

requirements-dev.txt
pirapira: Kept the newer `mypy`.

PR #1181, opened 2019-08-19
Make sure deposit fails for deprecation switch

pirapira:
not because the amount is not enough.

Before this PR, it was unclear why the deposit transaction fails.

### DIFF
--- a/raiden_contracts/tests/test_service_registry.py
+++ b/raiden_contracts/tests/test_service_registry.py
@@ -295,12 +295,13 @@ def test_deprecation_switch(
     assert service_registry.functions.deprecation_switch().call()
     # A user tries to make a deposit
     (A,) = get_accounts(1)
-    custom_token.functions.mint(SERVICE_DEPOSIT).call_and_transact({"from": A})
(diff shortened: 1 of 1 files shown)


raiden_contracts/tests/test_service_registry.py
```diff
@@ -295,12 +295,11 @@ def test_deprecation_switch(
     assert service_registry.functions.deprecation_switch().call()
     # A user tries to make a deposit
     (A,) = get_accounts(1)
-    custom_token.functions.mint(SERVICE_DEPOSIT).call_and_transact({"from": A})
-    custom_token.functions.approve(service_registry.address, SERVICE_DEPOSIT).call_and_transact(
-        {"from": A}
-    )
-    with pytest.raises(TransactionFailed):
-        service_registry.functions.deposit(SERVICE_DEPOSIT).call_and_transact({"from": A})
+    mint = service_registry.functions.currentPrice().call()
```
karlb: The variable `mint` could have a better name like `current_price` or `minted_amount`.

PR #1195, opened 2019-08-23
Test changing the bump denominator to zero

pirapira:
This PR adds a test that changes the bump denominator of ServiceRegistry to zero.  Such a change fails with a specific error message.


pirapira:
I don't have a good decisive answer.  We can test `changeParametersInternal()` and in effect test both cases.  However, what we care ultimately is what happens on external calls.

PR #1197, opened 2019-08-23
Deprecation switch can be turned on twice

pirapira:
but the second call has no effects.


pirapira:
There are no good reasons for these behaviors.

PR #1198, opened 2019-08-23
Test setting the min price above the current price

pirapira:
Then the current price becomes the new min price.


raiden_contracts/tests/test_service_registry.py
```diff
@@ -238,6 +238,19 @@ def test_changing_min_price(service_registry: Contract) -> None:
     assert service_registry.functions.min_price().call() == DEFAULT_MIN_PRICE * 2
 
 
+def test_changing_min_price_above_current(service_registry: Contract) -> None:
+    """Changin min_price above the current price."""
```
karlb:
```suggestion
    """Changing min_price above the current price."""
```

PR #1199, opened 2019-08-23
Test too big duration so overflow happens

pirapira:
when a deposit is made.

### DIFF
--- a/raiden_contracts/tests/test_service_registry.py
+++ b/raiden_contracts/tests/test_service_registry.py
@@ -142,8 +142,8 @@ def test_changing_duration(
         _min_price=DEFAULT_MIN_PRICE,
         _registration_duration=new_duration,
     ).call_and_transact({"from": CONTRACT_DEPLOYER_ADDRESS})
-    assert service_registry.functions.registration_duration().call() == new_duration
     # make sure that the duration has changed.
+    assert service_registry.functions.registration_duration().call() == new_duration
     (A,) = get_accounts(1)
     custom_token.functions.mint(2 * SERVICE_DEPOSIT).call_and_transact({"from": A})
     custom_token.functions.approve(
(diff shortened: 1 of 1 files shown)


raiden_contracts/tests/test_service_registry.py
```diff
@@ -156,6 +156,29 @@ def test_changing_duration(
...
+    new_duration = 2 ** 256 - 1
+    service_registry.functions.changeParameters(
+        _price_bump_numerator=DEFAULT_BUMP_NUMERATOR,
+        _price_bump_denominator=DEFAULT_BUMP_DENOMINATOR,
+        _decay_constant=DEFAULT_DECAY_CONSTANT,
+        _min_price=DEFAULT_MIN_PRICE,
+        _registration_duration=new_duration,
+    ).call_and_transact({"from": CONTRACT_DEPLOYER_ADDRESS})
+    assert service_registry.functions.registration_duration().call() == new_duration
+    # make sure that the duration has changed.
```
karlb: Shouldn't this comment be one line above?

raiden_contracts/tests/test_service_registry.py
```diff
@@ -156,6 +156,29 @@ def test_changing_duration(
...
+    (A,) = get_accounts(1)
+    custom_token.functions.mint(2 * SERVICE_DEPOSIT).call_and_transact({"from": A})
+    custom_token.functions.approve(
+        service_registry.address, 2 * SERVICE_DEPOSIT
+    ).call_and_transact({"from": A})
+    with pytest.raises(TransactionFailed, match="overflow during extending the registration"):
```
karlb: Great to see this `match` parameter! I didn't know it exists and I'm glad to see that it works fine for the `require` case!

Issue #581, opened 2019-02-18
Test unhappy cases of Service Registry contract


pirapira:
* [x] a case where `onlyController()` fires. (already done)
* [x] try to overflow, underflow the `set_price` variable (looks impossible)
* [x] try to use `set_price_at` in an uninitialized situation (constructor-->token-->reentrancy); that's impossible because the code is not deployed yet
* [x] what if the constructor calls something that calls back the ServiceRegistry?  That fails because the code is not deployed yet.
* [x] try to use absurd numbers for `decay_constant` https://github.com/raiden-network/raiden-contracts/pull/1190 https://github.com/raiden-network/raiden-contracts/pull/1194
* [x]  try `price_bump_denominator == 0` https://github.com/raiden-network/raiden-contracts/pull/1195
* [x] try to cause overflow in multiplying by `price_bump_numerator` (seems impossible)
* [x] try `price_bump_numerator == 0` https://github.com/raiden-network/raiden-contracts/pull/1196
* [x] try calling `setDeprecationSwitch()` twice with the controller's private key https://github.com/raiden-network/raiden-contracts/pull/1197
* [x] try calling `changeParameters` after deprecation switch is set https://github.com/raiden-network/raiden-contracts/pull/1197
* [x] isn't the return value of `changeParametersInternal()` ignored? (it doesn't return a value)
* [x] try setting `min_price` above `currentPrice()`. https://github.com/raiden-network/raiden-contracts/pull/1198
* [x] try setting a huge `registration_duration` so new registrations cause overflows in time. https://github.com/raiden-network/raiden-contracts/pull/1199
* [x] try to cause an overflow in `decayedPrice()` calculation (seems impossible)
* [x] try calling `setPriceBumpParameters()` directly from out of the contract https://github.com/raiden-network/raiden-contracts/pull/1202
* [x] try calling `setMinPrice()` directly from out of the contract https://github.com/raiden-network/raiden-contracts/pull/1203
* [x] try calling `setDecayConstant()` directly from out of the contract #1204
* [x] try creating `Deposit` with a release timestamp in the past (cannot happen https://github.com/raiden-network/raiden-contracts/pull/1205)
* [x] try `withdraw` from a wrong caller (to be done in #1184)
* [x] try `withdraw` where `service_registry` has no code (to be done in #1184)
* [x] try `withdraw` with some ETH balance (to be done in #1184)
[...]

PR #1204, opened 2019-08-30
Test calling the internal setDecayConstant()

pirapira:
function from out of the contract.

This commit is a part of solving #581.


pirapira:
No, we usually don't test this kind of things.  This time I saw similar-looking functions `changeParameters` and `changeParametersInternal`, and thought they might be confusing.

PR #1207, opened 2019-08-30
Add a test creating ServiceRegistry with zero

pirapira:
as bump denominator.

This commit is a part of solving #581.


pirapira:
kwargs and argument documentation will be kept track in separate issues. ^

PR #1237, opened 2019-09-24
Comment on subtleties

pirapira:
### What this PR does

This PR edits comments on some tricky parts.  They explain possible reentrancy and why some failure is crucially important.

### Why I'm making this PR

We rely on the fact that TokenNetwork.updateNonClosingBalanceProof() fails
when the same balance proof has been already submitted.  Otherwise Monitoring
Services can gain rewards again and again using the same Reward Proof.

We call an untrusted external contract in MonitoringService.monitor() contract
and it's tricky.  So I document the tricky parts here.

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep


pirapira:
The compiled bytecode contains a hash of metadata (containing the source).

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -177,8 +177,13 @@ contract MonitoringService is Utils {
...
     {
-        // Call updateReward first to avoid reentrancy problems when the
-        // token_network_address belongs to a malicious contract.
+        // Here we're trying to do bookkeeping first, but updateReward() first calls
+        // token_network_address.  So reentrancy is possible.
+        // In that case, the outer frame fails and reverts the state
+        // because token_network_address is not registered in the token_network_registry.
+        //
+        // Maybe it's simpler and safer to take the token address as an argument instead,
+        // and ask the TokenNetworkRegistry for the token_network_address.
```
karlb: I don't see why it would be easier, but if we rely on registering networks for security, we definitely should check the registration.

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -177,8 +177,13 @@ contract MonitoringService is Utils {
     )
         public
     {
-        // Call updateReward first to avoid reentrancy problems when the
-        // token_network_address belongs to a malicious contract.
+        // Here we're trying to do bookkeeping first, but updateReward() first calls
+        // token_network_address.  So reentrancy is possible.
```
karlb: I see that it calls `TokenNetwork` on `token_network_address`.

pirapira: TokenNetwork(...) is like a mypy `BlockNumber(...)`.  It doesn't check anything at runtime.

raiden_contracts/data/source/services/MonitoringService.sol
pirapira: If we deploy this contract with care, we can make sure `token_network_registry.token_to_token_networks(token_network_token)`, causes no reentrancies.

In the current code, when we call token_network_address, we tolerate the possibility of reentrancy.

PR #1255, opened 2019-09-27
Error on unknown contract names

pirapira:
### What this PR does

This PR modifies etherscan_verify so it raises an error on an unknown contract name.

### Why I'm making this PR

Before this PR the script silently succeeded while not doing anything.  This can be confusing when a user makes a typo.

Nothing.

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -21,3 +21,4 @@ isort
 pipdeptree
 eth-typing<2.0.0
 eth-abi<=1.2.2
+six>=1.12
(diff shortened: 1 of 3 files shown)


requirements-dev.txt
```diff
@@ -21,3 +21,4 @@ isort
 pipdeptree
 eth-typing<2.0.0
 eth-abi<=1.2.2
+six>=1.12
```
karlb: What's that for?

pirapira: CI fails without it.  Somehow `master` is also broken.  Maybe merging this alone https://github.com/raiden-network/raiden-contracts/pull/1264 is the better way ahead.

pirapira: #1261 applied the same change.

PR #1260, opened 2019-10-02
Constructor computes checksums (so users don't need to initialize the object)

pirapira:
### What this PR does

This PR makes ContractSourceManager's constructor compute the checksums.  Before this PR, users had to initialize the object by calling `compute_checksums()` before calling some other methods.

### Why I'm making this PR

It's nicer when users don't need to care about the internal states of the class.

I made sure that once initialized `contracts_checksums` and `overall_checksum` do not change.

----

Any reviewer can check these:

* [x] If the PR is fixing a bug or adding a feature, add an entry to CHANGELOG.md.
* [x] If the PR changed a Solidity source, run `make compile_contracts` and add the resulting `raiden_contracts/data/contracts.json` in the PR.
* [x] If the PR is changing documentation only, add `[skip ci]` in the commit message so Travis does not waste time.
    * [x] But, if the PR changes comments in a Solidity source, do not add `[skip ci]` and let Travis check the hash of the source.
* [x] In Python, use keyword arguments
* [x] Squash unnecessary commits
* [x] Comment commits
* [x] Follow naming conventions
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* [x] Follow the Signature Convention in [CONTRIBUTING.md](./CONTRIBUTING.md)
* For each new contract
    * [x] The deployment script deploys the new contract.
    * [x] `etherscan_verify.py` runs on the new contract.
* Bookkeep
    * [x] The gas cost of new functions are stored in `gas.json`.
    * [x] Document arguments of functions in natspec
    * [x] Care reentrancy problems
* [x] When you catch a require() failure in Solidity, look for a specific error message like `pytest.raises(TransactionFailed, match="error message"):`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Documents changes that result in:
 
 ## Unreleased
 
+- [#1260](https://github.com/raiden-network/raiden-contracts/pull/1260) Simplify the usage of ContractSourceManager. Users don't need to call compute_checksums() to initialize the object.
+
 ## [0.32.0](https://github.com/raiden-network/raiden-contracts/releases/tag/v0.32.0) - 2019-09-25
 
 - The minimum settlement window is now shorter (20 blocks) on test networks.
--- a/raiden_contracts/contract_source_manager.py
+++ b/raiden_contracts/contract_source_manager.py
@@ -3,7 +3,7 @@ import hashlib
 import json
 from os import chdir
 from pathlib import Path
-from typing import Dict
+from typing import Dict, Tuple
 
 from solc import compile_files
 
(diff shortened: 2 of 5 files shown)


requirements-dev.txt
```diff
@@ -21,3 +21,4 @@ isort
 pipdeptree
 eth-typing<2.0.0
 eth-abi<=1.2.2
+six>=1.12
```
konradkonrad: This looks unrelated to the rationale of the PR. Is 2/3 compatibility really a concern?

pirapira: This caused CI failures unrelated to this PR. But, I've added this line on `master` already. This diff will disappear when I rebase.

pirapira: This diff has disappeared.

PR #1285, opened 2019-10-14
Simplify gas price logic

### LINKED ISSUE #1284: What's special about gas_price == 0?
The constructor of `ContractDeployer` deals with `gas_price == 0` case in a special way.  What is the purpose of this special case?

https://github.com/raiden-network/raiden-contracts/blob/f6d5879cbe0d014ff74622b5dcea64fa0201ff86/raiden_contracts/deploy/contract_deployer.py#L51
```
        if gas_price != 0:
            self.transaction["gasPrice"] = gas_price * int(units["gwei"])
```

pirapira:
Before this commit, if the user specifies `--gas-price 0`, that
caused the Ethereum node to choose the gas price.  That was
counterintuitive, and this feature was not documented.

This commit removes this possibility.

This fixes #1284.

### What this PR does

This commit removes
* the undocumented feature of `ContractDeployer` that lets the Ethereum node choose the gas price, when 0 was specified as the gas price.
* the default gas price value in `ContractDeployer`.

### Why I'm making this PR

These features were not documented and were never used. The scripts should just fail in these weird cases rather than try to find something reasonable on its own.

Nothing.

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep

### DIFF
--- a/raiden_contracts/deploy/contract_deployer.py
+++ b/raiden_contracts/deploy/contract_deployer.py
@@ -39,7 +39,7 @@ class ContractDeployer(ContractVerifier):
         web3: Web3,
         private_key: str,
         gas_limit: int,
-        gas_price: int = 1,
+        gas_price: int,
         wait: int = 10,
         contracts_version: Optional[str] = None,
     ):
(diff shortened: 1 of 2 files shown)


raiden_contracts/deploy/contract_deployer.py
```diff
@@ -39,7 +39,7 @@ def __init__(
         web3: Web3,
         private_key: str,
         gas_limit: int,
-        gas_price: int = 1,
+        gas_price: int,
```
palango: It might make sense to adapt the Click parameter as well to only allow bigger than zero values:

```py
@click.option("--gas-price", default=5, type=int, help="Gas price to use in gwei")
```

pirapira: That's a good idea.

PR #1306, opened 2019-10-23
Move data_0.25.2 into data_0.33.0

### LINKED ISSUE #1305: Move data_0.25.2 into a new version number
Since we [changed](https://github.com/raiden-network/raiden-contracts/pull/1303) the version number scheme, it's reasonable to move `data_0.25.2` into `data_0.33.0`.  Since `data_0.25.2` never worked for anybody, the move does not harm anybody.

pirapira:
The next PyPI package version will be 0.33.3 so moving
the contracts_version to 0.33.0.

This follows a new policy that's introduced in
https://github.com/raiden-network/raiden-contracts/pull/1303

I chose the patch version 0 because the patch version in
contracts_version will always be zero (until they hit 1.0.0).

This closes #1305.

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep


raiden_contracts/constants.py
```diff
@@ -8,7 +8,7 @@
 from raiden_contracts.utils.type_aliases import ChainID
 
 # Do not change this, this is handled by bumpversion with .bumpversion_contracts.cfg
```
karlb: Is this really still handled by bumpversion?

raiden_contracts/constants.py
```diff
@@ -8,7 +8,7 @@
...
-CONTRACTS_VERSION = "0.25.2"
+CONTRACTS_VERSION = "0.33.0"
```
karlb: Just using `0.33` would make it more obvious that this is not affected by patch releases. There might be downsides to that approach, but I don't see any, right now.

raiden_contracts/constants.py
pirapira: No, and I was not seeing this comment.

raiden_contracts/constants.py
pirapira: I'm using `semver.compare()` on these strings.  Perhaps I need to find another way to compare versions.  Let me keep the `.0` at the end for now.

raiden_contracts/constants.py
pirapira: Fixed the comment.

Issue #1298, opened 2019-10-21
Provide consistent contracts.json and deployments.json in data_0.25.2


pirapira:
`data_0.33.2` contains inconsistent `contracts.json` and `..._deployment.json`s.  When a user finds a deployment following the deployment files, these deployments contain different bytecode from those in `contracts.json`.  This causes  https://github.com/raiden-network/raiden/issues/5122.

This issue keeps track of fixing this problem either by
* making `data_0.33.3`
* or modifying `data_0.33.2`

pirapira:
If we modify `data_0.33.2`, make sure to revert the Solidity sources in `data_0.33.2` back to the older sources.

PR #1307, opened 2019-10-23
Revert data to data_0.25.2

### LINKED ISSUE #1300: revert `data` to where `data_0.25.2`
because we will be working on `data_0.25.2` for a while.

pirapira:
We have deployed our contracts to the mainnet and we are no longer
making frequent deployments.  But we have to be ready to
patch the Solidity contracts.  So our current Solidity sources
must be what have been deployed.

data_0.25.2 was once inconsistent #1298 partially because
we had edited the Solidity files in `data` directory.

This closes #1300.

### What this PR does

Reverts the contents of `data` to those of `data_0.25.2`.

### Why I'm making this PR

To prevent https://github.com/raiden-network/raiden/issues/5122 in the future.

The PR was made very manually, so things might break.

----

Any reviewer can check these:

* [ ] Comment commits


palango:
Is there any tool that I can use to verify that the contracts now match the deployed contracts?

pirapira:
I connected to our node and did
```
eth.getCode(address)
```

copied and pasted the result to a file.  And eventually ran `diff`.

palango:
So, that would mean for me:
- install `solc` 0.5.4
- run `make compile_contracts` on this PR
- `getCode` for the deployments
- compare the two

karlb:
Do we really need to care about this in `raiden_contracts/data`? Don't we have `raiden_contracts/data_0.XX.0` for all cases where we need exactly the contracts as they were deployed? If we want to minimally patch a contract, we should branch off the respective release tag instead of keeping the current master branch is a specific state, shouldn't we?

pirapira:
@karlb In this PR I am trying to prepare for minimally patched contracts. I think it's a good idea to have an `alderaan` branch. So I'll write down how to use the release branches https://github.com/raiden-network/raiden-contracts/issues/1327

Issue #1329, opened 2019-10-29
deploy_tester_contract() does not reveal the error message


pirapira:
When `deploy_tester_contract()` causes a `require()` failure, `deploy_tester_contract()` raises an exception, but the exception does not contain the error message in `require()`.

This issue keeps track of fixing `deploy_tester_contract()` so that it raises an exception with the error message.

pirapira:
This is because `contract.constructor().transact()` does not fire an informative exception.  Does anybody asking for that feature in `web3.py`?

pirapira:
The documentation of `Web3.py` does not mention what happens when the transaction fails for some reasons https://web3py.readthedocs.io/en/stable/contracts.html#web3.contract.Contract.constructor.

pirapira:
It's not special because it's a constructor. I was just using `transact()` not `call()`.

pirapira:
Web3.py's ContractConstructor does not have `call()` method. It takes time to solve this one.

PR #1333, opened 2019-10-31
Error message from require() failures during deployments

pirapira:
This solves #1329.

### What this PR does

This commit modifies `deploy_tester_contract()` so it raises
an exception with a message with failed `require()`s.

### Why I'm making this PR

Before this commit, `deploy_tester_contract()` did not raise an
exception when the deployment failed for a `require()` condition.

* [ ] Comment commits
    * `python_variable`
    * `PYTHON_CONSTANT`


pirapira:
Not sure how to get this working.  `Web3.py`'s ContractConstructor does not have a `call()` function.

Issue #572, opened 2019-02-14
Test failure cases of MS contract


pirapira:
This issue keeps track of listing and implementing failing cases of MS contract.

pirapira:
* [ ] running constructor with token address being zero
* [ ] running constructor with service registry address being zero
* [ ] running constructor with token address without code
* [ ] running constructor with service_registry_address without code
* [ ] running constructor with udc_address without code
* [ ] running constructor with token_network_registry without code

* [ ] updateReward(): "Unknown TokenNetwork"
* [ ] updateReward(): "Bad reward proof"
* [ ] updateReward(): "stale nonce"
* [ ] monitor() --> updateReward() said reentrancy
* [ ] monitor() --> updateReward() "Unknown TokenNetwork"
* [ ] monitor() --> updateReward() "Bad reward proof"
* [ ] monitor() --> updateReward() "stale nonce"

* [ ] more

PR #1335, opened 2019-10-31
Test MonitoringService deployment with token address without code

pirapira:
address without code.

This is a part of testing failure cases in MonitoringService https://github.com/raiden-network/raiden-contracts/issues/572 .

Nothing.

----

Any reviewer can check these:

* [ ] Comment commits
    * `python_variable`
    * `PYTHON_CONSTANT`

### DIFF
--- a/raiden_contracts/tests/test_monitoring_service.py
+++ b/raiden_contracts/tests/test_monitoring_service.py
@@ -441,11 +441,31 @@ def test_monitoring_service_deploy_with_token_address_zero(
     with pytest.raises(TransactionFailed):
         deploy_tester_contract(
             contract_name=CONTRACT_MONITORING_SERVICE,
-            _token_address=EMPTY_HEXADDRESS,
+            _token_address=EMPTY_HEXADDRESS,  # causes error
             _service_registry_address=service_registry.address,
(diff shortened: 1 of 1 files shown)


raiden_contracts/tests/test_monitoring_service.py
```diff
@@ -437,7 +441,26 @@ def test_monitoring_service_deploy_with_token_address_zero(
...
+    token_network_registry_contract: Contract,
+) -> None:
+    # No good error message is available due to
+    # https://github.com/raiden-network/raiden-contracts/issues/1329
+    # with pytest.raises(TransactionFailed, match="token has no code"):
+    with pytest.raises(TransactionFailed):
+        deploy_tester_contract(
+            contract_name=CONTRACT_MONITORING_SERVICE,
+            _token_address=CONTRACT_DEPLOYER_ADDRESS,  # causes error
             _service_registry_address=service_registry.address,
```
err508: You could add an address here where deploy_tester_contrct doesn’t raise if it’s not to slow, so the diff to test cases is more obvious

pirapira: I'll add a comment to a fixture that performs the successful deployment.

PR #1339, opened 2019-11-12
Update eth-tester to Istanbul compatibility, update web3.

### LINKED ISSUE #868: Update to eth-tester for Istanbul compatibility

palango:
### What this PR does

Reincarnation of #1153 
Fixes #868

Updates web3 and eth-tester to recent versions. This lets us run the tests on an instanbul enabled test chain.

### Why I'm making this PR

We want to test the contracts with the latest ethereum versions and functionality.

Nothing

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep


raiden_contracts/tests/deprecation_switch_testnet.py
```diff
@@ -65,8 +65,8 @@ def deprecation_test(
 
     # Check that we can open channels and deposit on behalf of A and B
     # Some arbitrary Ethereum addresses
-    A = "0x6AA63296FA94975017244769F00F0c64DB7d7115"
-    B = "0xc9a4fad99B6d7D3e48D18d2585470cd8f27FA61e"
+    A = HexAddress(HexStr("0x6AA63296FA94975017244769F00F0c64DB7d7115"))
```
karlb: Why do we need the `HexStr`? Can't we cast to `HexAddress` directly?

raiden_contracts/tests/fixtures/base/web3_fixtures.py
```diff
@@ -56,6 +56,9 @@ def web3(ethereum_tester: EthereumTester) -> Web3:
         }
     )
 
+    # Enable strict type checks
+    web3.enable_strict_bytes_type_checking()
```
karlb: Nice!

raiden_contracts/tests/deprecation_switch_testnet.py
palango: Yeah, that one is weird:
```
mypy raiden_contracts/ setup.py
raiden_contracts/tests/deprecation_switch_testnet.py:68: error: Argument 1 to "HexAddress" has incompatible type "str"; expected "HexStr"

Comment (karlb) [reply raiden_contracts/tests/deprecation_switch_testnet.py]: It's really defined that way:
```
HexAddress = NewType('HexAddress', HexStr)
```
see https://github.com/ethereum/eth-typing/blob/965459459f1a4f1090ccc004b433c11fb4ad97b4/eth_typing/evm.py#L16

We've never cascaded NewTypes in raiden, so this was unusual for me.

PR #1349, opened 2019-12-12
Remove defaults from service registry

palango:
Those values have to be supplied in the constructor, so the defaults
will never be used.

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep


setup.py
```diff
@@ -73,7 +73,7 @@ def run(self):
             contracts_source_path,
         )
 
-        contract_manager = ContractSourceManager(contracts_source_path())
+        contract_manager = ContractSourceManager(contracts_source_path(None))
```
karlb: How did that work before?

karlb:
Both @palango and me don't have any idea why the deployments fails, right now.

Issue #1155, opened 2019-08-12
Try to upgrade solc to the latest version


pirapira:
And see whether that requires upgrades of other dependencies.

pirapira:
The urgency of #1153 depends on this.

pirapira:
When I try to use `solc 0.5.9`, tests fail because TokenNetworkRegistry doesn't get deployed.  The runtime bytecode looks small enough.  This has to be investigated.

pirapira:
Now `solc` is 0.5.11.

pirapira:
Before doing this, we need a way to selectively compile only some of the contracts.  Namely, when we want to keep using an old SecretRegistry, `contracts.json` must contain a SecretRegistry bytecode compiled by an older Solidity, and the other bytecode compiled by a newer Solidity.

pirapira:
Blocked on #1208.

PR #1363, opened 2020-02-21
Update to solc 0.6.3

karlb:
This will allow us to use the new ABI encoder without getting warnings
about it being experimental.

No behaviour changes in the contracts are intended in this commit.

Closes https://github.com/raiden-network/raiden-contracts/issues/1155.


karlb:
Changelog at https://github.com/ethereum/solidity/blob/develop/Changelog.md#063-2020-02-18

palango:
Will we run into problems like this? https://github.com/ethereum/web3.py/issues/1566

karlb:
> Will we run into problems like this? [ethereum/web3.py#1566](https://github.com/ethereum/web3.py/issues/1566)

We use web3 in our tests, so we should be fine when the tests are green as long as we don't need any specific new features. Or am I missing something?

PR #1364, opened 2020-02-25
Refactor firstBlockAllowedToMonitor

karlb:
This makes more of the logic publicly accessible, which allows us to
remove duplicated logic from the test setup and the from the MS.

Since the function signature changed, this will need a change in the MS.


raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -227,15 +233,15 @@ contract MonitoringService is Utils {
         );
     }
 
-    function isAllowedToMonitor(
+    function firstBlockAllowedToMonitor(
         TokenNetwork token_network,
```
karlb: This can be called publicly by passing in the TokenNetwork's address.

palango:
@karlb Do you want to merge this before the next release?

karlb:
> @karlb Do you want to merge this before the next release?
> 
> If so, please rebase.

There's no need to hurry this PR. I still rebased, so that it can be easily merged when convenient.

karlb:
> Maybe `firstBlockAllowedToMonitor` -> `firstBlockAllowedToMonitorChannel`

Then I'll also use `firstBlockAllowedToMonitor` instead of `pureFirstBlockAllowedToMonitor`. That
* is one less rename in the diff
* a bit shorter
* a more usual name, since we usually don't include information like "pure" in the variable name.

Issue #902, opened 2019-04-29
Try Solium on the Solidity sources


pirapira:

pirapira:
I got some useful messages.

```
TokenNetworkJoined.sol
  10:8       warning    Provide an error message for require().                               error-reason
  19:8       error      Avoid using Inline Assembly.                                          security/no-inline-assembly
  32:8       warning    Provide an error message for require().                               error-reason
  37:8       warning    Provide an error message for require().                               error-reason
  96:8       error      Avoid using Inline Assembly.                                          security/no-inline-assembly
  211:46     error      String literal must be quoted with double quotes.                     quotes
  356:8      warning    Provide an error message for require().                               error-reason
  361:8      warning    Provide an error message for require().                               error-reason
  366:8      warning    Provide an error message for require().                               error-reason
  371:8      warning    Provide an error message for require().                               error-reason
  372:8      warning    Provide an error message for require().                               error-reason
  400:8      warning    Provide an error message for require().                               error-reason
  401:8      warning    Provide an error message for require().                               error-reason
  402:8      warning    Provide an error message for require().                               error-reason
  403:8      warning    Provide an error message for require().                               error-reason
  404:8      warning    Provide an error message for require().                               error-reason
  405:8      warning    Provide an error message for require().                               error-reason
  406:8      warning    Provide an error message for require().                               error-reason
  407:8      warning    Provide an error message for require().                               error-reason
  408:8      warning    Provide an error message for require().                               error-reason
  409:8      warning    Provide an error message for require().                               error-reason
  410:8      warning    Provide an error message for require().                               error-reason
  420:8      warning    Provide an error message for require().                               error-reason
  427:56     warning    Visibility modifier "public" should come before other modifiers.      visibility-first
  440:8      warning    Visibility modifier "public" should come before other modifiers.      visibility-first
  447:8      warning    Provide an error message for require().                               error-reason
  458:8      warning    Provide an error message for require().                               error-reason
  499:8      warning    Visibility modifier "public" should come before other modifiers.      visibility-first
  501:8      warning    Provide an error message for require().                               error-reason
  502:8      warning    Provide an error message for require().                               error-reason
  503:8      warning    Provide an error message for require().                               error-reason
  516:8      warning    Provide an error message for require().                               error-reason
  520:8      warning    Provide an error message for require().                               error-reason
  527:8      warning    Provide an error message for require().                               error-reason
  535:8      warning    Provide an error message for require().                               error-reason
[...]

Issue #1144, opened 2019-07-26
CI enforcement of Solidity styles (e.g. functionNameStyle())


pirapira:
Somehow I wrote functions named `set_min_price()` and so accidentally in a Solidity file.  This issue keeps track of enforcing one style using Solium or another tool in CI.

karlb:
Closed by https://github.com/raiden-network/raiden-contracts/pull/1426.

PR #1368, opened 2020-03-10
Use solium to check all solidity files

karlb:
To make it pass, some excludes have been added. Solium is only used if
it is already installed locally.

I removed the fallback from `HumanStandardToken` because I did not know
how to make it pass solium and I don't see why we would want that
behavior and it's also not in HumanStandardToken at
https://github.com/ConsenSys/Token-Factory/blob/master/contracts/HumanStandardToken.sol.

See https://github.com/raiden-network/raiden-contracts/issues/1144, closes https://github.com/raiden-network/raiden-contracts/issues/902.

Enabling this in the CI can be done next.


karlb:
> Also, solium on CI would be nice.

Yes, we have https://github.com/raiden-network/raiden-contracts/pull/1368 for that. If you have a  suggestion on how to best run node based software in our CI, please add a comment to that issue.

> It would be good though, to know the minimum solium version to get this going? `solium 1.1.2`, which I can get through package manager, did not work.

I used Solium version 1.2.5, but that should get more well defined once we run it in CI.

PR #1374, opened 2020-03-20
0.37.0-test release

karlb:
This is part of our release test plan. I hope the release suffix does not cause any problems.

When this is merged, a release tag should be pushed to make Travis do a PyPI release.


Dominik1999:
use a + instead of a -

karlb:
> use a + instead of a -

I thought treating it as a [pre-release](https://semver.org/#spec-item-9) is more accurate than just appending [build-metadata](https://semver.org/#spec-item-10), since "A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version."

Dominik1999:
Please excuse my short note 
> use a + instead of a -

In the standup yesterday we shortly talked about that issue and it seems that if you use a `+` instead of a `-` before the `test` in the name, you will not run into trouble. I am sure @ulope can explain that better.

ulope:
Yeah ok doing a pre-release is a good idea.
But `test` isn't a valid release type according to PEP440.

The available [pre-release types and aliases are](https://www.python.org/dev/peps/pep-0440/#pre-release-spelling):
| Canonical | a     | b    | rc      |
| --------- | ----- | ---- | ------- |
| Aliases   | alpha | beta | c       |
|           |       |      | pre     |
|           |       |      | preview |

You can add arbitrary suffixes as a `local version segment` (separated with a `+` which is why I initially suggested this) but those are more supposed to be used for re-packaging of existing releases.

karlb:
>  But test isn't a valid release type according to PEP440.

Ahh, I only checked the semantic version spec and not the pep. Good to know!

PR #1379, opened 2020-04-08
Bump version: 0.37.0-beta → 0.37.0b2

konradkonrad:
### What this PR does
Bump the PyPi version of `raiden-contracts` to `0.37.0b2`.

### Why I'm making this PR
This will allow us to upload a new beta release to PyPi

I used `bump2version --config-file .bumpversion.cfg --new-version '0.37.0-beta2' 0.37.0-beta2`, but still had to
manually adjust the `VERSION` constant in `setup.py`. Maybe there are other places, that I missed?

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ from setuptools import Command
 from setuptools.command.build_py import build_py
 
 DESCRIPTION = "Raiden contracts library and utilities"
-VERSION = "0.37.0-beta"
+VERSION = "0.37.0b2"
 
 
 def read_requirements(path: str) -> List[str]:
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,7 +1,8 @@
 [bumpversion]
-current_version = 0.37.0-beta
+current_version = 0.37.0b2
 commit = True
 tag = True
 
 [bumpversion:file:setup.py]
 serialize = {major}.{minor}.{patch}
+


setup.py
```diff
@@ -11,7 +11,7 @@
 from setuptools.command.build_py import build_py
 
 DESCRIPTION = "Raiden contracts library and utilities"
-VERSION = "0.37.0-beta"
+VERSION = "0.37.0"
```
Dominik1999, line 14: why is this not `-beta2`?

ulope:
Nitpick: the "proper" / canonical spelling would be `0.37.0b2`

setup.py
konradkonrad: wait....that's exactly what I tried to fix here. give me a second!

konradkonrad:
> Nitpick: the "proper" / canonical spelling would be `0.37.0b2`

The normalization actually happens automatically. I.e. `"0.37.0-beta"` earlier always folded into `0.37.0b0` (see e.g. https://travis-ci.org/github/raiden-network/raiden-contracts/jobs/672460279#L6103 )

ulope:
Yeah but to avoid problems with tools that don't know about PEP440 normalization I think it would make sense to use the canonical version everywhere.

konradkonrad:
> Yeah but to avoid problems with tools that don't know about PEP440 normalization I think it would make sense to use the canonical version everywhere.

I see. I tried to stay consistent with the previous scheme. But I'll change it :)

Issue #1367, opened 2020-03-09
Deployer only supports interactive password input


kelsos:
Currently, the deployer scripts only support interactive password input for contract deployment.

This makes it harder if you want to use the steps in some automated way, eg. a script that does a deployment for a testnet automatically.

It would be nice for the deployer to support getting the password using a password file and/or an environment variable.

PR #1390, opened 2020-05-28
Deploy script now optionally takes `password-file`

karlb:
Without this, you will be prompted for the password.

Closes https://github.com/raiden-network/raiden-contracts/issues/1367

### DIFF
--- a/raiden_contracts/deploy/__main__.py
+++ b/raiden_contracts/deploy/__main__.py
@@ -64,6 +64,14 @@ def common_options(func: Callable) -> Callable:
 
     @click.option("--private-key", required=True, help="Path to a private key store.")
     @click.option(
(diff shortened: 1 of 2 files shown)


raiden_contracts/deploy/__main__.py
```diff
@@ -87,6 +95,7 @@ def wrapper(*args: List, **kwargs: Dict) -> Any:
 def setup_ctx(
     ctx: click.Context,
     private_key: str,
+    password_file: Optional[Path],
```
palango, line 98: Would that ever work if one doesn't provide a password file? Or, why is this optional?

karlb: If not password file is given, you get a password prompt.

Issue #1129, opened 2019-07-18
Use V2 ABI in the settlement function


pirapira:
We need four new arguments in the settlement function
* two for #1123 
* two for #1121 
but currently, this causes too deep stack.

This issue keeps track of using V2 ABI of Solidity so the stack depth constraint can be releaved.

Dominik1999:
does not include the Raiden Side

PR #1392, opened 2020-06-02
Use ABI V2 to shorten settle arg list

karlb:
This allows us to add additional arguments for hash salts and signatures
that allow settling on behalf of other users.

I kept compatibility with the existing signature and introduced a new
settle function using structs. Is this a good idea? Also a better name
than `settleChannel2` would be nice.

Closes https://github.com/raiden-network/raiden-contracts/issues/1129


raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -124,6 +125,13 @@ contract TokenNetwork is Utils {
         uint256 locked_amount;
     }
 
+    struct ParticipantSettleInput {
```
konradkonrad, line 128: imho we could go with a shorter name here:
`SettleInput` (having a mandatory `participant` argument) should be completely self-explanatory :)

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -692,6 +700,30 @@ contract TokenNetwork is Utils {
         bytes32 participant2_locksroot
     )
         public
```
konradkonrad, line 702: why wouldn't we change the signature for `settleChannel` as well?
```
function settleChannel(
  uint256 channel_identifier,
  SettleInput participant1,
  SettleInput participant2
) ...
```

konradkonrad: ah, expand-contract pattern...I see.

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -725,16 +759,16 @@ contract TokenNetwork is Utils {
 
         require(verifyBalanceHashData(
```
konradkonrad, line 760: `verifyBalanceHashData` is also never used outside the context of settlement, right? So it could use `(Participant)SettleInput` as the sole argument.

karlb:
If you can spot any additional refactoring opportunities related to `SettleInput`, let me know! @konradkonrad

PR #1394, opened 2020-06-04
Move basic Raiden types and start using them

karlb:
These can be imported from Raiden, so that both packages share the same
types for basic Raiden objects.

### DIFF
--- a/raiden_contracts/tests/fixtures/channel.py
+++ b/raiden_contracts/tests/fixtures/channel.py
@@ -6,6 +6,7 @@ from eth_typing import HexAddress
 from hexbytes import HexBytes
 from web3 import Web3
 from web3.contract import Contract
+from web3.types import Nonce
 
 from raiden_contracts.constants import TEST_SETTLE_TIMEOUT_MIN, ChannelState, MessageTypeId
 from raiden_contracts.tests.utils import (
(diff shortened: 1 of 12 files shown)


raiden_contracts/utils/type_aliases.py
```diff
@@ -1,4 +1,26 @@
 from typing import NewType
```
palango, line 1: Should we import `web3.Nonce` here as well, so we can use it from that module as well?

raiden_contracts/tests/fixtures/channel.py
```diff
@@ -648,28 +657,28 @@ def get(
 @pytest.fixture(scope="session")
 def create_balance_proof(token_network: Contract, get_private_key: Callable) -> Callable:
     def get(
-        channel_identifier: int,
+        channel_identifier: ChannelID,
         participant: HexAddress,
         transferred_amount: int = 0,
```
palango, line 662: Doesn't need to be done in this PR, but should this be `TokenAmount`?

raiden_contracts/utils/type_aliases.py
karlb: I thought it would be better if we imported that directly from web3, so that it is easier to spot which types are actually the same. But I have not strong opinion either way.

raiden_contracts/utils/proofs.py
```diff
@@ -1,13 +1,26 @@
...
+    Signature,
+    TokenAmount,
+)
 
 from .signature import sign
 
 
-def hash_balance_data(transferred_amount: int, locked_amount: int, locksroot: bytes) -> bytes:
+def hash_balance_data(
+    transferred_amount: int, locked_amount: int, locksroot: bytes
```
palango, line 22: `TokenAmount`s here as well
I also think we should have a type for the Locksroot (hash)

raiden_contracts/utils/type_aliases.py
palango: No strong feelings either, let's keep it like this then.

PR #1395, opened 2020-06-04
Remove data from old contracts releases

### LINKED ISSUE #1393: Remove old deployment data
The contracts package currently bundles infos from old deployments infos (the `data_*` folders). These are not required any more and can be removed. This will reduce the size of the package, as each folder adds more than 1MB right now.

palango:
### What this PR does

Removed data about all old deployments before Alderaan.

Resolves #1393 

### Why I'm making this PR

Each deployment adds roughly 1MB of data to the package size, even though they most likely won't ever be used again.

It's big.

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep


raiden_contracts/tests/test_contracts_compilation.py
```diff
@@ -238,7 +222,7 @@ def test_contract_manager_constructor_does_not_invent_version() -> None:
     assert manager.contracts_version is None
 
 
-@pytest.mark.parametrize("version", [CONTRACTS_VERSION, "0.9.0", "0.3._", "0.4.0"])
+@pytest.mark.parametrize("version", [CONTRACTS_VERSION])
```
karlb, line 225: It would be nice to have at least one more case, here. Even if you just put '0.37.0' in, it will check that we can load other versions than the current one as soon as we do the next release.

palango:
After a discussion with @konradkonrad I decided to keep the red eyes contract data for now, as it isn't deprecated yet.

PR #1396, opened 2020-06-05
Use byte PrivateKey instead of hex encoded strs

karlb:
The types are declared as bytes, so the actual runtime code should use bytes, too!

Followup of the type cleanup in https://github.com/raiden-network/raiden-contracts/pull/1394.

### DIFF
--- a/raiden_contracts/tests/test_deploy_script.py
+++ b/raiden_contracts/tests/test_deploy_script.py
@@ -2,7 +2,7 @@ import json
 from copy import deepcopy
 from pathlib import Path
 from tempfile import NamedTemporaryFile
-from typing import Callable, Generator, List, Optional
+from typing import IO, Generator, List, Optional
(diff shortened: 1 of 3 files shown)


raiden_contracts/tests/test_deploy_script.py
```diff
@@ -954,8 +954,8 @@ def test_deploy_token_with_balance(get_accounts: Callable, get_private_key: Call
     """ Call deploy token command with a private key with some balance """
     (signer,) = get_accounts(1)
     priv_key = get_private_key(signer)
-    with NamedTemporaryFile() as privkey_file:
-        privkey_file.write(bytearray(priv_key, "ascii"))
+    with NamedTemporaryFile("w") as privkey_file:
```
palango, line 957: Would be nice to refactor this and the next two lines, as it's done often.

Issue #1187, opened 2019-08-22
Use the chain_id instruction in Istanbul 1


pirapira:
Optionally, the ABI can be simplified because functions do not need chain_id as arguments.

karlb:
See https://ethereum.stackexchange.com/a/77550/49987

PR #1399, opened 2020-06-23
Get chain-id from opcode

karlb:
This does not work yet, due to
https://github.com/ethereum/web3.py/issues/1677.

See https://github.com/raiden-network/raiden-contracts/issues/1187.


karlb:
Starting with solidity 0.8, there is `block.chainid` to make this even easier.

PR #1401, opened 2020-07-03
Claims

palango:
This is just a POC to test how claims can look like.

I guess we don't need the `Distribution` contract and can just make the `TokenNetwork` an operator for the token directly.


raiden_contracts/data/source/raiden/Distribution.sol
```diff
@@ -0,0 +1,66 @@
...
+    ) public {
+        address receiver = recoverAddressFromClaim(
+            owner,
+            partner,
+            amount,
+            signature
+        );
+
+        require(receiver == owner, "Signature mismatch");
+        token.mintFor(amount, receiver);
```
fredo, line 41: Why should the claim be signed by the owner?

raiden_contracts/tests/fixtures/distribution.py
```diff
@@ -0,0 +1,65 @@
...
+        "_token_address": custom_token.address,
+        "_chain_id": web3.eth.chainId,
+    }
+
+
+@pytest.fixture(scope="session")
+def distribution_contract(
+    deploy_tester_contract: Callable, distribution_constructor_args: Dict
+) -> Contract:
+    """Deployed TokenNetworkRegistry contract"""
```
fredo, line 32: shouldnt this be called, `DistributionContract`

raiden_contracts/data/source/raiden/Distribution.sol
fredo: The claim should not mint for a address rather than directly depositing it into a channel shouldnt it?

fredo:
Didnt we want to create its own branch for that?

raiden_contracts/data/source/raiden/Distribution.sol
palango: This is all WIP, I just started with the most simple way. But we need a distributor address.

fredo: true

fredo: So you mint for any address which is then going to deposit it into channels, right?

palango: I think for settlement we need to further evolve this. Basically we need to be able to split a claim over two participants.

karlb:
Please use a `raiddit` branch as merge target instead of `master`.

PR #1402, opened 2020-07-09
Settle virtual channels

karlb:
* Work with new channel_ids (https://github.com/raiden-network/raiddit/issues/18)
* Allow allow operations on channels without an on-chain createChannel
* Accept claims during settlement

These features are clearly not finished, yet. They are only provided in this state to allow the client to start coding against the interface early and to reduce merge conflicts. Claim signatures are not even checked, yet.

Other changes like keeping the channel open after settle and handling burned tokens are out of scope for this PR and will be done later.

See https://github.com/raiden-network/raiddit/issues/28.


raiden_contracts/tests/conftest.py
```diff
@@ -1,2 +1,9 @@
-# pylint: disable=W0614, C0111, W0401
+"""
+isort:skip_file
+"""
+from pytest import register_assert_rewrite
+
+register_assert_rewrite("raiden_contracts.utils.events")
+register_assert_rewrite("raiden_contracts.tests.fixtures.channel")
```
konradkonrad, line 7: from [here](https://docs.pytest.org/en/stable/writing_plugins.html#assertion-rewriting) it looks as if you could give an iterable

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -523,7 +546,7 @@ contract TokenNetwork is Utils {
         channel.participants[closing_participant].is_the_closer = true;
 
         // This is the block number at which the channel can be settled.
-        channel.settle_block_number += uint256(block.number);
+        channel.settle_block_number = settlement_timeout + block.number;
```
ulope, line 549: Probably not important for a PoC but this could overflow if the contract is deployed with a very high `settlement_timeout_min`

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -696,10 +719,12 @@ contract TokenNetwork is Utils {
         uint256 participant1_transferred_amount,
         uint256 participant1_locked_amount,
         bytes32 participant1_locksroot,
+        Claim memory participant1_claim,
         address participant2,
         uint256 participant2_transferred_amount,
         uint256 participant2_locked_amount,
-        bytes32 participant2_locksroot
+        bytes32 participant2_locksroot,
+        Claim memory participant2_claim
```
ulope, line 727: Is it guaranteed (or rather do we assume) that claims will always be issued pair-wise?

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -859,12 +889,12 @@ contract TokenNetwork is Utils {
...
         // Channel represented by channel_identifier must be settled and
         // channel data deleted
-        require(channel_identifier != getChannelIdentifier(receiver, sender));
+        // require(channel_identifier != getChannelIdentifier(receiver, sender), "Bad channel_id");
 
         // After the channel is settled the storage is cleared, therefore the
         // value will be NonExistent and not Settled. The value Settled is used
         // for the external APIs
-        require(channels[channel_identifier].state == ChannelState.NonExistent);
+        require(channels[channel_identifier].state == ChannelState.NonExistent, "Channel still exists");
```
ulope, line 897: Does this check make sense with `NonExistent` meaning open as well?

raiden_contracts/data/source/raiden/TokenNetwork.sol
karlb: It is not guaranteed. Since I don't see a way to pass something like `None` instead of a solidity struct, my current approach would be to not check the signature when the amount is 0, so that the client can pass in a dummy claim instead. Better suggestions welcome!

raiden_contracts/data/source/raiden/TokenNetwork.sol
karlb: Yes, that check is questionable. It only checks for the channel does not have a settlement in progress. I'll have to revisit this when I have added support for settling multiple times.

PR #1403, opened 2020-07-10
Include burnt_tokens in balance proof.

karlb:
Include burnt_tokens in balance proof.

To send BP with burnt_tokens during settlement, you need to use `settleChannel2`, as adding more parameters to `settleChannel` is not possible without exceeding the stack size.

Closes https://github.com/raiden-network/raiddit/issues/36


Dominik1999:
is the goal directory of this PR correct? You want to merge into Raiddit, right?

karlb:
> is the goal directory of this PR correct? You want to merge into Raiddit, right?

You're right, I set the target to raiddit.

PR #1405, opened 2020-07-13
Use raiddit contracts by default

konradkonrad:
### What this PR does
This changes the default CONTRACT_VERSION flag to `None`, so `raiden` will use the latest version by default.

See: https://github.com/raiden-network/raiden/blob/develop/raiden/tests/fixtures/blockchain.py#L11


konradkonrad:
Superseeded by https://github.com/raiden-network/raiden/pull/6379

PR #1406, opened 2020-07-14
Allow settling multiple times

karlb:
The paid out amounts are saved in `withdrawn_amount`, now. This could
cause race conditions, if we also allow normal withdraws. This could be
solved by using two separate monotonic variables for withdraws and
settlements.

I'm hesitant to add another field right now to not increase the contract
size any further (see raiden-network/raiddit#39). I also expect us
to not use withdraws in the PoC, so we can do that later.

Closes https://github.com/raiden-network/raiddit/issues/35.
Closes https://github.com/raiden-network/raiddit/issues/42.


raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -789,10 +791,10 @@ contract TokenNetwork is Utils {
...
         );
 
-        // Remove the channel data from storage
-        delete channel.participants[participant1];
-        delete channel.participants[participant2];
-        delete channels[channel_identifier];
+        // Update channel state
+        participant1_state.withdrawn_amount += participant1_settlement.transferred_amount;
+        participant2_state.withdrawn_amount += participant2_settlement.transferred_amount;
+        channel.state = ChannelState.NonExistent;
```
palango, line 797: That feels a bit weird. But we don't get any advantage from differentiating more here.

PR #1407, opened 2020-07-17
Pass etherscan api key to checkverifystatus

karlb:
That call needs an api key too, now.

### DIFF
--- a/raiden_contracts/deploy/etherscan_verify.py
+++ b/raiden_contracts/deploy/etherscan_verify.py
@@ -55,7 +55,7 @@ def etherscan_verify(
     chain_id: ChainID, apikey: str, guid: Optional[str], contract_name: Optional[str]
 ) -> None:
     if guid:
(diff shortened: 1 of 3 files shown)


raiden_contracts/deploy/etherscan_verify.py
```diff
@@ -256,8 +256,13 @@ def _verify_singleton_contract(
     )
 
 
-def guid_status(etherscan_api: str, guid: str) -> Dict:
-    data = {"guid": guid, "module": "contract", "action": "checkverifystatus"}
+def guid_status(etherscan_api: str, guid: str, apikey: str,) -> Dict:
```
palango, line 259: trailing comma

karlb: Huh, how did that get trough black and the linters? I'll fix it.

palango: I've also seen this. I think black doesn't fix that for whatever reason.

PR #1408, opened 2020-07-17
Restrict minting

karlb:


palango:
That should go the the `raiddit` branch, right?

karlb:
> That should go the the `raiddit` branch, right?

Yes, it should. Fixed.

raiden_contracts/data/source/test/StandardToken.sol
```diff
@@ -73,11 +75,17 @@ contract StandardToken is Token {
...
+    function tokenNetworkMintFor(uint256 num, address target) public override {
+        require(msg.sender == token_network, "Only TokenNetwork can mint");
         balances[target] += num;
         _total_supply += num;
 
         require(balances[target] >= num, "Balance overflow");
         // assert(_total_supply >= num);
     }
+
+    function setTokenNetwork(address _token_network) public {
```
palango, line 87: We should limit which addresses are allowed to can call this.

karlb: It can only be set once and must be set during the deployment process. So in the worst case an attacker can make the deployment fail by calling this faster and force us to make a fresh deployment. The compromised token would just not get used.

We could add an owner during in the constructor and restrict only allow it to call this. I wanted to avoid the impression that any party has control over the token (and minimize code changes), but we can easily add that. Should I do it?

palango: With the explanation I think it's fine. I missed the only-settable-once part, so I guess we can live with that.

PR #1409, opened 2020-07-17
add burnt amount upon closing for happy closing

fredo:
# Description

Burning is guarantueed by submitting the burnproof performed by a monitoring service which acts on behalf of reddit.
To not have the necissity of a monitoring service to settle a channel correctly we allow providing the burnt amount voluntarely by the users and settle based on these data if no burning proof is submitted.

`closeChannel` and `updateNonClosingBalanceProof` gets a new argument `burnt_amount`


raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -130,9 +130,15 @@ contract TokenNetwork is Utils {
         bytes signature;
     }
 
+    struct BalanceProof {
```
karlb, line 133: `HashedBalanceProof` to be precise

raiden_contracts/tests/fixtures/channel.py
```diff
@@ -67,7 +71,7 @@ def get(A: HexAddress, B: HexAddress) -> Tuple:
         channel_identifier = token_network.functions.getChannelIdentifier(A, B).call()
 
         # Test the channel state on chain
-        (channel_settle_timeout, channel_state) = token_network.functions.getChannelInfo(
+        (channel_settle_timeout, channel_state,) = token_network.functions.getChannelInfo(
```
palango, line 74: Looks like this only returns a two element tuple: https://github.com/raiden-network/raiden-contracts/blob/raiddit/raiden_contracts/data/source/raiden/TokenNetwork.sol#L941

palango: This seems to work anyways

PR #1415, opened 2020-09-15
Unique channel ids

karlb:
Channel ids have only been unique for a single token network so far.
This change hashes the token network address and chain id into the
channel id, thus making it globally unique.

Advantages:
* Simplifies usage of raiden contacts (e.g. you can create channel_id ->
channel mappings)
* Identifying channels during debugging is easier
* No chance of accidentally identifying the wrong channel
* Could mitigate attacks that rely on reusing signatures
* Mostly backwards compatible

Disadvantages:
* `openChannel` costs ~3300 gas more
* We can't distinguish between channels that have never been created and
channels that have been removed. We used the sequential channel_id for
that before
* Channel ids are usually longer than before


karlb:
There's no clear preference either way, so let's be conservative and just close this.

PR #1428, opened 2021-02-24
Minimal code splitting for TokenNetwork - using a library for pure functions 

palango:
Fixes https://github.com/raiden-network/raiden-contracts/issues/401

This PR splits the a first pure function out of the token network contract into an library. It also implements the necessary infrastructure for doing this, like deploying and linking the library into the token network contract code.

I did not yet pull all possible code into the library in order to make this PR smaller and easier to review. Once this is merged it should be easy to do this and reduce the contract code size further.

I also decided to only create a library for the token network contract now, as the code size gains seems small compared to the added complexity from deploying/linking for the other contracts look unworthy.


palango:
@karlb This passes all but two tests now. I'm working on those.

If you want you can already give this a first look and provide some feedback.

.github/workflows/python-package.yml
```diff
@@ -49,6 +49,10 @@ jobs:
       run: |
         source ./venv/bin/activate
         make lint
+    - name: Test
```
karlb, line 52: This step has the same name as the one in line 56. Different names would be more helpful.

.github/workflows/python-package.yml
```diff
@@ -49,6 +49,10 @@ jobs:
...
+      run: |
+        source ./venv/bin/activate
+        pytest -svx raiden_contracts/tests/test_token_network_registry.py
```
karlb, line 55: Why do we need a separate step for this test?

karlb (commented):
Looks good so far! This will probably make changes to the etherscan verififaction necessary, too.

.github/workflows/python-package.yml
palango: Ah no, this can be removed now that it's working.

palango:
The etherscan verification are still WIP, but it can be pushed into a follow-up when necessary. The rest is good for review.

raiden_contracts/deploy/contract_verifier.py
```diff
@@ -199,6 +200,16 @@ def _verify_deployed_contract(
...
         blockchain_bytecode = self.web3.eth.get_code(contract_instance.address).hex()
         compiled_bytecode = self.contract_manager.get_runtime_hexcode(contract_name)
+
+        if contract_name == CONTRACT_TOKEN_NETWORK_REGISTRY:
+            # We need to link the libs into the contract bytecode.
+            # As this is run in the tests in fake file systems, do poor mans linking here
+            lib_address = contracts[LIBRARY_TOKEN_NETWORK_UTILS]["address"]
+            normalized_address = lib_address[2:].lower()
+            compiled_bytecode = compiled_bytecode.replace(
+                "__$34600480520cb524a2c423e33a5b4dd437$__", normalized_address
```
karlb, line 210: Does this need to be replaced by a dynamic value in a followup, or is this some kind of magic constant?

karlb (approved):

raiden_contracts/deploy/contract_verifier.py
palango: So this is the hash of the library path, I guess adding the calculation (if only for future reference) makes sense.
```py
 token_network_libs_key = "data/source/lib/TokenNetworkUtils.sol:TokenNetworkUtils"
"__${}$__".format(encode_hex(keccak(bytes(token_network_libs_key, 'utf-8')))[2:36])
> '__$34600480520cb524a2c423e33a5b4dd437$__'
```

I'm not really happy on how the linking is done here, but it's the best solution I could find. The problem is that `compile_contracts` doesn't now the address of the library yet, so it cannot do the linking. Here I'd rather not encode the relations between the contracts, but I don't see way around it.

karlb: Yes, adding that calculation would make this both easier to understand and adapt to changes.

Issue #1182, opened 2019-08-19
Deployment script needs to support reusing ServiceRegistry


pirapira:
Similarly to SecretRegistry, instances of ServiceRegistry contracts are supposed to survive some new deployment cycles.  This issue keeps track of supporting reuse of ServiceRegistry instances in the contract deployment scripts.

PR #1430, opened 2021-02-25
Allow reuse service registry

palango:
### What this PR does

Fix https://github.com/raiden-network/raiden-contracts/issues/1182

Makes the reuse of the `ServiceRegistry` possible.

- Is there actually a point in re-suing the service registry, but deploying a new UDC?

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep

### DIFF
--- a/raiden_contracts/deploy/__main__.py
+++ b/raiden_contracts/deploy/__main__.py
@@ -311,6 +311,11 @@ def raiden(
     callback=validate_address,
     help="Address of TokenNetworkRegistry that MS contract looks at",
 )
+@click.option(
+    "--service-registry-from-deployment-file",
+    type=click.Path(exists=True),
+    help="The deployment file from which ServiceRegistry should be reused",
+)
 @click.pass_context
 def services(
     ctx: Context,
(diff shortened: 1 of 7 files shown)


raiden_contracts/deploy/__main__.py
```diff
@@ -311,6 +311,11 @@ def raiden(
     callback=validate_address,
     help="Address of TokenNetworkRegistry that MS contract looks at",
 )
+@click.option(
+    "--service-registry-from-deployment-file",
+    type=click.Path(exists=True),
+    help="The deployment file from which ServiceRegistry should be reused",
+)
```
karlb, line 318: Should this default to `raiden_contracts/data/deployment_{network}.json`?

What about reusing other contracts (UDC)? It would make sense to get their addresses from the same file and not pass one file per reused contract, wouldn't it?

palango: The `UDC` can unfortunately not be reused because it needs to be initialized with the trusted contracts in an `init` call after deployment. This can only be done once.

The default makes sense, I'll add it.

karlb (approved):

raiden_contracts/deploy/__main__.py
palango: The default is hard to preset, because it differs based on the chain used. I added a comment in the deployment docs instead.

Issue #1435, opened 2021-03-04
Move further code from TokenNetwork to library


palango:
in https://github.com/raiden-network/raiden-contracts/pull/1428 only a couple of functions were pulled out to minimize the diff.

### Tasks
- Go through all remaining functions and see whether they can be moved to the library.
- Try moving `structs`/`enum`s to the library, see if it reduces contract code size.

PR #1437, opened 2021-03-04
Move `recoverAddressFrom*` methods to library

palango:
### What this PR does

Part of https://github.com/raiden-network/raiden-contracts/issues/1435

Moves some more code to the token network library

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep


raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -400,7 +400,7 @@ contract MonitoringService is Utils {
                 "\x19Ethereum Signed Message:\n221",  // 20 + 32 + 32 + 20 + 20 + 65 + 32
                 address(this),
                 chain_id,
-                uint256(MessageTypeId.MSReward),
+                uint256(6),
```
palango, line 403: Not so nice, any better ideas?

karlb: Does the compiler resolve enums during compilation? If that is the case, you should be able to have the enum in Utils (or a similar superclass if Utils contains too much baggage) and use it everywhere without it increasing the size of the resulting contracts.

karlb: [The docs](https://docs.soliditylang.org/en/latest/types.html#enums) don't say it explicitly, but it sounds like the enums are resolved to the respective ints during compilation. So it might be possible to add the enum to all contracts without increasing their size.

palango: The problem is that libraries are not allowed to inherit:
```
Error: Library is not allowed to inherit.
 --> data/source/lib/TokenNetworkUtils.sol:8:1:
  |
8 | library TokenNetworkUtils is Utils{
  | ^ (Relevant source part starts here and spans across multiple lines).
make: *** [compile_contracts] Error 1

palango: But I can try pushing it into another lib like `ECVerify`

karlb: Ahh, I didn't know that. You could also copy the Enum and add a comment like `when changing this, please also update the copy in otherFile.sol`. Not very elegant, but better than having the plain numbers, IMO.

palango: That's true. 

I added another small lib now: https://github.com/raiden-network/raiden-contracts/pull/1437/files#diff-76de3909eb0a0b12081b5e019df894b1f383260355d1f9051efb92ac6dc97929R7

That seems to work fine and is a good solution IMO.

karlb (approved):

PR #1441, opened 2021-03-16
Add `make upload-pypi` command

karlb:
This will build and upload the PyPI package. In the past, this was done
by travis. If we want to automate this again, this make target can be
used.


palango (approved):
🎖️

PR #1443, opened 2021-03-18
Fix deprecations (2021.03) and remove semver

ulope:
### What this PR does

Fixes various deprecation warnings and removes `semver` in favor of `semantic_version` (which was already required through dependencies).

For some reason black decided to reformat some files.

### Why I'm making this PR

To fix warnings

Nothing particular

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep

### DIFF
--- a/raiden_contracts/tests/test_service_registry.py
+++ b/raiden_contracts/tests/test_service_registry.py
@@ -51,7 +51,10 @@ def test_deposit_contract_without_service_registry_code(
     (A,) = get_accounts(1)
     call_and_transact(custom_token.functions.mint(100), {"from": A})
     depo = get_deposit_contract(
-        _token=custom_token.address, _release_at=UINT256_MAX, _withdrawer=A, _service_registry=A
+        _token=custom_token.address,
+        _release_at=UINT256_MAX,
+        _withdrawer=A,
+        _service_registry=A,
     )
     call_and_transact(custom_token.functions.transfer(depo.address, 100), {"from": A})
     assert custom_token.functions.balanceOf(A).call() == 0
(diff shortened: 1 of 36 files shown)


palango (approved):
LGTM

raiden_contracts/tests/test_service_registry.py
```diff
@@ -51,7 +51,10 @@ def test_deposit_contract_without_service_registry_code(
     (A,) = get_accounts(1)
     call_and_transact(custom_token.functions.mint(100), {"from": A})
     depo = get_deposit_contract(
-        _token=custom_token.address, _release_at=UINT256_MAX, _withdrawer=A, _service_registry=A
+        _token=custom_token.address,
+        _release_at=UINT256_MAX,
+        _withdrawer=A,
+        _service_registry=A,
```
netcriptus, line 57: Not related to this PR, but why are the arguments here prefixed with `_`?

netcriptus (approved):
I'd challenge the assumption that the current lint configuration really helps making the code more readable (more specifically: one import/argument per line instead of multiple columns, and maximum line length of 99 columns).

Apart from that, LGTM :shipit:

raiden_contracts/tests/test_service_registry.py
karlb: Those are the same parameter names used in the constructor of the `Deposit` solidity contract. They are chosen that way in the constructor to avoid name clashes with Deposit contract's attributes. Solidity doesn't have `self` or `this`, so this is the way we arrive at code like
```
token = _token
```
in the constructor to set the `token` attribute to the value given to the constructor.

PR #1447, opened 2021-04-06
Remove the need for local solc install for users

karlb:
Packages that use raiden-contracts' fixtures don't need a solc
installation, anymore. This is achieved by doing the linking step in
pure python.


palango (approved):
Looks good.

PR #1448, opened 2021-04-07
Fix setuptools setup_requires

ezdac:
This PR will fix the `setup.py` installation method by providing the necessary dependencies for the verify_contracts command that is employed during setup.
(see https://github.com/raiden-network/raiden-services/pull/951#issuecomment-814959179)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ config = {
     "license": "MIT",
     "keywords": "raiden ethereum blockchain",
     "install_requires": requirements,
-    "setup_requires": _get_single_requirement(requirements, "py-solc"),
+    "setup_requires": requirements,
     "packages": find_packages(),
     "include_package_data": True,
     "classifiers": [


setup.py
```diff
@@ -88,7 +88,7 @@ def run(self) -> None:  # pylint: disable=no-self-use
     "license": "MIT",
     "keywords": "raiden ethereum blockchain",
     "install_requires": requirements,
-    "setup_requires": _get_single_requirement(requirements, "py-solc"),
```
karlb, line 91: We can remove `_get_single_requirement` function, now that it is not used, anymore.

setup.py
```diff
@@ -88,7 +88,7 @@ def run(self) -> None:  # pylint: disable=no-self-use
...
+    "setup_requires": requirements,
```
karlb, line 91: Do we need `setup_requires` at all, if it is the same as install_requires?

karlb (approved):
Looks good, but why does it solve the `ModuleNotFoundError: No module named 'eth_typing'` error?

setup.py
ezdac: I tried it out locally and it doesn't seem to work if omitted completely.

setup.py
ezdac: I'm actually not sure if the **entire** set of requirements is used for the verify-contracts functionality, but it must be at least a subset of it. If we don't care about eventually unnecessary requirements, the function is not useful anymore, yes.

setup.py
karlb: I have problems finding more information about it other than that it is "deprecated in favor of the PEP 517". If it works, let's keep it for now.

ezdac:
> Looks good, but why does it solve the `ModuleNotFoundError: No module named 'eth_typing'` error?

I explained it here a little: https://github.com/raiden-network/raiden-services/pull/951#issuecomment-814959179

Inside of the verify_contracts command's `run` method, there is functionality of the package itself imported and used. The verify_contracts is run in "setup" phase when installing the package with the `setup.py` method.

This probably hasn't been noticed before, because the setup.py method hasn't been used to install the package in CI since some time. (But here I'm not sure)
Inside the `raiden-services` installation, it will now get used as a fallback mechanism, which made installation fail as whole, since no other method could be employed.

PR #1449, opened 2021-04-27
Increment cache key version

konradkonrad:
### What this PR does

Increment the cache key in gh actions workflow.

### Why I'm making this PR

Invalidate older CI caches.


palango (approved):
Looks good, but can you fix the lint error while you're at it?

PR #1451, opened 2021-04-27
Pin web3, fix resulting problems

karlb:


palango (approved):
Strange that the exceptions change like that. Was `eth-tester` updated as well?

konradkonrad (commented):
thanks for fixing this!

karlb:
> Strange that the exceptions change like that. Was `eth-tester` updated as well?

I don't think so, it's pinned as `eth-tester[py-evm]==0.5.0b3`.

PR #1452, opened 2021-04-29
Automatically install solc via solcx

karlb:
Developers won't have to manually install solc, anymore.


karlb:
@palango Do you have an idea why the second commit makes the solium installation fail?

I could just remove 207b8e2, but I would like to understand this first. One explanation would be that solium relies on an installed solc, but I saw this neither in the error message nor in the solium docs.

palango:
> @palango Do you have an idea why the second commit makes the solium installation fail?

No, doesn't make much sense from my pov either.

karlb:
PR with only the first commit is https://github.com/raiden-network/raiden-contracts/pull/1453

PR #1453, opened 2021-04-29
Automatically install solc via solcx

karlb:
Developers won't have to manually install solc, anymore.

Replaces #1452.


konradkonrad (approved):
looks easy enough :) +1

Issue #1463, opened 2021-06-08
Set service token default to RDN


karlb:
This is required in the TGE obligation arguments.

PR #1464, opened 2021-06-10
Use RDN as default token for service payments

karlb:
We promised to use RDN during the token sale. We can't prevent anyone
from deploying the contracts with a different token selected, but we can
at least choose RDN as default.

Closes https://github.com/raiden-network/raiden-contracts/issues/1463.


palango (approved):
👍

PR #1465, opened 2021-06-10
requirements.txt: don't pin web3 to 5.18.0

istankovic:


karlb:
We don't have any mechanism to freeze versions in this repo. Is it a good idea to not restrict the version in any way? I'm afraid of clearing the pip cache and suddenly getting errors I didn't get before.

Related to https://github.com/raiden-network/raiden/issues/6984

karlb (approved):

Issue #1383, opened 2020-05-07
Cooperative settlement (simple implementation and clear old code)


hackaugusto:
https://github.com/raiden-network/raiden-contracts/issues/399 was commented out to be re-enabled in the future, but after a discussion with @karlb we realized that withdraw alone provides the same capabilities as coop settle. Instead of having two primitives it makes more sense to just have one.

The only drawback is that two withdraws would be more expensive then a single coop settle. This has to be measured and weighted to see if it is significant enough to keep coop settle around.

karlb:
I've done some measurements to compare the gas cost of different approaches:

## Classic Approaches

### Two Withdraws
Don't implement anything new and expect both parties to do a withdraw if they don't want to use the channel, anymore.
Gas cost: 2 * 109_565 = 219_130
Result: open, empty channel

### Coop Settle
Activate the cooperative settle function which is currently disabled in the contract. The code to handle coop settlement in Raiden is still missing, but would look similar to our current withdraw code.
Gas cost: 60_857
Result: settled channel

## Hybrid Approaches

These two approaches reuse our current withdraw signatures, but both nodes coordinate to only have a single on-chain transaction. That reduces the number of different proofs we have to handle compared to the original coop settle, but apart from that is about as complex.
Potential attack: Instead of executing both withdraws, the transaction submitter could initiate only one withdraw (relatively harmless)

### Two Withdraws in a Single Transaction
Add a small wrapper function to execute both withdraws in a single transaction.
Gas cost: 154_140
Result: open, empty channel

### Coop Settle Based on Withdraw
Reuse the existing code to create and check withdraw proofs in a new coop settle function. The function would check both withdraw proofs, but instead of withdrawing, it settles the channel. 
Gas cost: ~100_000
Result: settled channel

palango:
@karlb Which of the hybrid approaches was the one that you and Augusto favored?

karlb:
I don't think there was a clear favorite, but I would suggest going with "Coop Settle Based on Withdraw".

palango:
Also discussed in https://github.com/raiden-network/raiden-contracts/issues/1236

palango:
Fixed by https://github.com/raiden-network/raiden-contracts/pull/1474

PR #1474, opened 2021-06-22
Cooperative settlement

palango:
### What this PR does

Implement cooperative settle (https://github.com/raiden-network/raiden-contracts/issues/1383)

#### Tasks
- [x] Add check (+ test) for withdraw of to small amount (settle withdraw < deposit - withdraws)
- [x] Fix event data, adapt tests (`test_cooperative_settle_channel_event`)
- [x] Add gas estimates for cooperative settlement
- [x] Fix event checker to be able ti handle multiple events of same kind

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep


raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -479,7 +479,77 @@ contract TokenNetwork is Utils {
...
+        delete channel.participants[data2.participant];
+        delete channels[channel_identifier];
+
+        // Remove the pair's channel counter
+        pair_hash = getParticipantsHash(data1.participant, data2.participant);
+        delete participants_hash_to_channel_identifier[pair_hash];
 
+        // Emit channel lifecycle events
+        emit ChannelClosed(channel_identifier, data1.participant, 0, 0);
+        emit ChannelSettled(channel_identifier, data1.total_withdraw, 0, data2.total_withdraw, 0);
```
karlb, line 552: Should we reuse the code from https://github.com/raiden-network/raiden-contracts/blob/15037b89cf322d1a1d309d097e06b7ed72c9e0b8/raiden_contracts/data/source/raiden/TokenNetwork.sol#L782-L809 here?

This question also shows that we are missing `storeUnlockData`. Is this a bug, or is there an explanation why we don't need it? We should add an explanation (if it is not needed) or a test to unlock after settlement (if it is needed).

raiden_contracts/tests/test_channel_cooperative_settle.py
```diff
@@ -7,147 +7,202 @@
...
+            (B, balance_B, expiration, 0x0, signature_A2),
+        )
+    with pytest.raises(ValidationError):
+        token_network.functions.cooperativeSettle(
+            channel_identifier,
+            (A, balance_A, expiration, signature_A1, signature_B1),
+            (B, balance_B, expiration, signature_B2, 0x0),
+        )
+    # empty addresses
     with pytest.raises(TransactionFailed):
```
karlb, line 91: I like to check for the specific error with
```
with pytest.raises(TransactionFailed, match="error message"):
```
whenever possible, so that we know we are actually exercising the correct require. That does require error message, of course. But we should take the opportunity to add error messages (https://github.com/raiden-network/raiden-contracts/issues/81) before the next mainnet deployment now that #401 is done.

raiden_contracts/tests/test_channel_cooperative_settle.py
```diff
@@ -312,13 +386,43 @@ def test_cooperative_settle_channel_state_withdraw(
...
+    (A, B, C) = get_accounts(3)
+    deposit_A = 20
+    deposit_B = 10
+    balance_A = 15 - 1
+    balance_B = 15
+
+    expiration = 100
+    channel_identifier = create_channel_and_deposit(A, B, deposit_A, deposit_B)
+
+    # Adding `withdraw_?` is neccessary, as `withdraw_amount` is monotonic increasing
```
karlb, line 404: I don't understand this comment.

raiden_contracts/tests/test_channel_cooperative_settle.py
```diff
@@ -173,25 +230,28 @@ def test_cooperative_settle_channel_0(
...
-    create_cooperative_settle_signatures: Callable,
+    create_withdraw_signatures: Callable,
     cooperative_settle_state_tests: Callable,
 ) -> None:
     (A, B, C) = get_accounts(3)
     deposit_A = 0
     deposit_B = 0
     balance_A = 0
     balance_B = 0
+    expiration = 100
```
karlb, line 246: You could add `expiration=100` as a default to `create_withdraw_signatures` to make the tests a bit more concise.

raiden_contracts/tests/test_channel_cooperative_settle.py
```diff
@@ -439,49 +562,73 @@ def test_cooperative_close_replay_reopened_channel(
...
     )
 
+    ev_handler.add(
+        txn_hash, ChannelEvent.WITHDRAW, check_withdraw(channel_identifier, B, balance_B)
+    )
+    # ev_handler.add(
+    #     txn_hash,
+    #     ChannelEvent.WITHDRAW,
+    #     check_withdraw(channel_identifier, A, balance_A)
+    # )
```
karlb, line 628: Why is this commented out? It looks like it should work.

raiden_contracts/tests/test_channel_cooperative_settle.py
palango: Great idea!

palango: Mh, as it is also used in every contract call this doesn't help much. I put it into a constant instead.

raiden_contracts/tests/test_channel_cooperative_settle.py
palango: The event checking code cannot handle checking two events with the same name in the same tx right now. I didn't want to implement that before a first review on the contract changes.

raiden_contracts/tests/test_channel_cooperative_settle.py
palango: Done for all tests where transaction are expected to fail. I didn't add it to the `web3` formatting tests. Do you think it's useful there as well?

karlb: I don't think so. IMO, those have limited value, anyway.

raiden_contracts/data/source/raiden/TokenNetwork.sol
palango: I factored out the mentioned lines and added some comments about why this doesn't allow remaining locks in the docstring as well as the code.

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -479,7 +506,74 @@ contract TokenNetwork is Utils {
...
+                data2.total_withdraw,
+                data2.expiration_block,
+                data2.participant_signature,
+                data2.partner_signature
+            );
+        }
+        removeChannelData(channel, channel_identifier, data1.participant, data2.participant);
+
+        // Emit channel lifecycle events
+        emit ChannelClosed(channel_identifier, data1.participant, 0, 0);
```
karlb, line 575: I'm not sure if we should emit this event, since the channel will never be in a closed state. I think I prefer to have it to make both settlement approaches more similar, but it does not match the concept fully (the nonce and balance_hash do not really apply) and removing it would also be slightly cheaper.

karlb (approved):
Really nice to finally have a coop settle after such a long time!

raiden_contracts/data/source/raiden/TokenNetwork.sol
palango: The intention was to keep the same channel lifecycle as before. In other words, I am not sure if the PFS/explorer/... rely on the close event in some sense.

But I will check on this again.

andrevmatos (approved):
I'm no solidity expert, but this looks good to me. It'd be good to have someone else more into solidity itself to review it though. I like the approach to leverage the double-withdraw signed structures, but we really need to optimise those gas usages

raiden_contracts/data/source/raiden/TokenNetwork.sol
palango: Seems like we need some adjustments in either case (with or without the `ChannelClosed` event):

- The PFS currently relies on the `ChannelClosed` event, and doesn't even handle the `ChannelSettled` event:
https://github.com/raiden-network/raiden-services/blob/37c6f4816e2a6a084bd4e7e9c0eb58b501870b66/src/pathfinding_service/service.py#L228-L229
  - However, adding the same handling for `ChannelSettled` events shouldn't be difficult
- The MS triggers a `ActionMonitoringTriggeredEvent` (https://github.com/raiden-network/raiden-services/blob/37c6f4816e2a6a084bd4e7e9c0eb58b501870b66/src/monitoring_service/handlers.py#L124) when a channel is closed. The event handler (https://github.com/raiden-network/raiden-services/blob/37c6f4816e2a6a084bd4e7e9c0eb58b501870b66/src/monitoring_service/handlers.py#L435) for this event doesn't currently checked if the channel in question is settled already.


tldr: We need to adjust the services anyways, so we can decide freely on which approach makes more logical sense.

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -475,7 +502,74 @@ contract TokenNetwork is Utils {
...
+        require(data1.total_withdraw <= data1.total_withdraw + data2.total_withdraw);
+
+        if (data1.total_withdraw > 0) {
+            this.setTotalWithdraw(
+                channel_identifier,
+                data1.participant,
+                data1.total_withdraw,
+                data1.expiration_block,
+                data1.participant_signature,
+                data1.partner_signature
```
andrevmatos, line 555: Nitpick question: can't we call `setTotalWithdraw2` directly with `data1` here? And data2 below? Wouldn't this be cheaper in gas?

palango: Well spotted!

raiden_contracts/data/source/raiden/TokenNetwork.sol
palango: We agreed to not emit the `ChannelClosed` event.

PR #1486, opened 2021-07-01
Add channel  `openWithDeposit` wrapper

### LINKED ISSUE #1436: Add function to combine open and deposit
When a channel is opened, the user usually also wants to deposit tokens into the channel. Right now, this involves two transactions and waiting for confirmation in between. Adding a function that combines these two operations will allow them to happen in a single transaction, thus reducing the time until the channel can be used. (An approve transaction has to be done in both cases, in addition to the transactions mentioned)

This function could be in an auxiliary contract. But if there is enough space left if the TokenNetwork contract, I suggest adding it there to keep the number of contracts low.

The expectation is that this is a low effort thing to add. If this is not the case, please bring up the complications in this issue before spending much time on the implementation, so we can reconsider if it's worth implementing.

palango:
### What this PR does
Fixes #1436 

Add a wrapper that opens a channel and deposits.

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep


raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -316,6 +316,39 @@ contract TokenNetwork is Utils {
...
+        address participant2,
+        uint256 settle_timeout,
+        uint256 participant1_total_deposit
+    )
+        public
+        isSafe
+        settleTimeoutValid(settle_timeout)
+        returns (uint256)
+    {
+        uint256 channel_identifier;
```
palango, line 338: Should we add a require for `participant1_total_deposit > 0` here? Otherwise the call to `participant1_total_deposit` will fail.

karlb: It will already fail with a descriptive error, AFAICS, so I don't see how another require would improve the situation.

palango: True, the only advantage would be failing earlier, but it should fail during gas estimation already.

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -1161,6 +1157,60 @@ contract TokenNetwork is Utils {
         return keccak256(abi.encodePacked(channel_identifier, sender, receiver));
     }
 
+    function setTotalDepositHelper(
```
karlb, line 1160: I don't like the name too much, as the "helper" does all of the work, not just some help. Maybe `setTotalDepositFor`>

raiden_contracts/data/gas.json
```diff
@@ -1,30 +1,31 @@
...
     "ServiceRegistry.setURL": 47223,
-    "TokenNetwork DEPLOYMENT": 3689329,
-    "TokenNetwork.closeChannel": 122790,
-    "TokenNetwork.openChannel": 115046,
-    "TokenNetwork.setTotalDeposit": 58468,
-    "TokenNetwork.setTotalWithdraw": 115102,
+    "TokenNetwork DEPLOYMENT": 3735537,
+    "TokenNetwork.closeChannel": 122819,
+    "TokenNetwork.openChannel": 115056,
+    "TokenNetwork.openChannelWithDeposit": 153255,
```
karlb, line 20: 153255 < (115056 + 58468 = 173524)

That's 12% less. A nice gas reduction in addition to being twice as fast as doing two transactions.

karlb (approved):

raiden_contracts/data/gas.json
palango: That's 20.269, so nearly the cost of a transaction.

PR #1494, opened 2021-07-27
Add participant addresses to ChannelSettled event

ezdac:
The participant1_* and participant2_* arguments were useless,
because it was not able to determine what partner corresponds
to the arguments. By adding the addresses this problem is solved.

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep


CHANGELOG.md
```diff
@@ -5,10 +5,12 @@ Documents changes that result in:
 - API changes in the package (externally used constants, externally used utilities and scripts)
 - important bug fixes between releases
 
+- Add both participants addresses to the TokenNetwork's ChannelSettled event
```
palango, line 8: Can you add a link to the PR here?

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -873,8 +875,10 @@ contract TokenNetwork is Utils {
 
         emit ChannelSettled(
             channel_identifier,
+	    participant1,
```
palango, line 878: the indention looks wrong.

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -192,8 +192,10 @@ contract TokenNetwork is Utils {
 
     event ChannelSettled(
         uint256 indexed channel_identifier,
+        address indexed participant1,
```
palango, line 195: Is there a reason you chose to make those parameters indexed?

ezdac: No, I was replicating the behaviour of the other events - I removed it just now because it is more expensive and not strictly necessary.

palango (approved):

PR #1495, opened 2021-07-28
raiden_contracts: change CONTRACTS_VERSION to 0.38.0

istankovic:
### What this PR does
Change the CONTRACTS_VERSION to 0.38.0.

### Why I'm making this PR
So that raiden uses the latest released contract code.

:warning: This PR should not be merged until we are reasonably sure that the newer contracts will not break tests in the `raiden` repository.

### DIFF
--- a/raiden_contracts/constants.py
+++ b/raiden_contracts/constants.py
@@ -9,6 +9,7 @@ from raiden_contracts.utils.type_aliases import ChainID, Locksroot
 
 # The last digit is supposed to be zero always. See `RELEASE.rst`.
 CONTRACTS_VERSION = "0.38.0"
+BESPIN_VERSION = "0.37.0"
 ALDERAAN_VERSION = "0.37.0"
 
 PRECOMPILED_DATA_FIELDS = ["abi", "bin", "bin-runtime", "metadata"]
--- a/raiden_contracts/constants.py
+++ b/raiden_contracts/constants.py
@@ -8,7 +8,7 @@ from eth_utils import keccak
 from raiden_contracts.utils.type_aliases import ChainID, Locksroot
 
 # The last digit is supposed to be zero always. See `RELEASE.rst`.
-CONTRACTS_VERSION = "0.37.0"
+CONTRACTS_VERSION = "0.38.0"
 ALDERAAN_VERSION = "0.37.0"
 
 PRECOMPILED_DATA_FIELDS = ["abi", "bin", "bin-runtime", "metadata"]


raiden_contracts/constants.py
```diff
@@ -8,7 +8,7 @@
 from raiden_contracts.utils.type_aliases import ChainID, Locksroot
 
 # The last digit is supposed to be zero always. See `RELEASE.rst`.
-CONTRACTS_VERSION = "0.37.0"
+CONTRACTS_VERSION = "0.38.0"
 ALDERAAN_VERSION = "0.37.0"
```
palango, line 12: I guess we should add - for completeness sake - a `BESPIN_VERSION` as well.

istankovic: Sounds ok, but it should have the same value as ALDERAAN, right?

palango: Yes

karlb (approved):

palango:
@istankovic Can't we just merge this? Before a new release this cannot be pulled into raiden anyways, right?

istankovic:
> @istankovic Can't we just merge this? Before a new release this cannot be pulled into raiden anyways, right?

I'm afraid we can't just pull this before either
- we fix the code in Raiden repo to support linking

or 

- get rid of split code (i.e. requiring linking) in raiden-contracts

palango:
But why? The raiden repo uses a pinned version of the contracts, right? So even if we change something here, this won't have any impact on the raiden repo.

istankovic:
> But why? The raiden repo uses a pinned version of the contracts, right?

Nope.
Worse, the version constant is being used internally to e.g. access contract-related data.

palango:
> > But why? The raiden repo uses a pinned version of the contracts, right?
> 
> Nope.
> Worse, the version constant is being used internally to e.g. access contract-related data.

https://github.com/raiden-network/raiden/blob/develop/requirements/requirements.txt#L224

istankovic:
> > > But why? The raiden repo uses a pinned version of the contracts, right?
> > 
> > 
> > Nope.
> > Worse, the version constant is being used internally to e.g. access contract-related data.
> 
> https://github.com/raiden-network/raiden/blob/develop/requirements/requirements.txt#L224

But that's not really pinned as it should be -- in the `.in` file. It's just the result of what the latest version was at the time of `deps compile`.

raiden_contracts/constants.py
istankovic: Done.

palango:
Thank you!

PR #1499, opened 2021-07-29
TokenNetwork error messages

palango:
Fixes https://github.com/raiden-network/raiden-contracts/issues/81

### What this PR does

Add error messages to whole `TokenNetwork` contract and check them in the tests.

### Why I'm making this PR

This makes it easier to develop against the contracts, see #81.

Some failures do not throw an error yet. But I think this is strictly better than before, so I'd go ahead and merge this and fix those subsequently.

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep


raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -586,7 +606,10 @@ contract TokenNetwork is Utils, Controllable {
         public
         isOpen(channel_identifier)
     {
-        require(channel_identifier == getChannelIdentifier(closing_participant, non_closing_participant));
+        require(
+            channel_identifier == getChannelIdentifier(closing_participant, non_closing_participant),
+            "TN/close: channel id not matching"
```
istankovic, line 611: Similarly here.

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -668,19 +694,19 @@ contract TokenNetwork is Utils, Controllable {
...
     {
-        require(channel_identifier == getChannelIdentifier(
-            closing_participant,
-            non_closing_participant
-        ));
-        require(balance_hash != bytes32(0x0));
-        require(nonce > 0);
+        require(
+            channel_identifier == getChannelIdentifier(closing_participant,non_closing_participant),
+             "TN/update: channel id not matching"
```
istankovic, line 699: Similarly here.

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -799,27 +825,30 @@ contract TokenNetwork is Utils, Controllable {
 
         address participant1 = participant1_settlement.participant;
         address participant2 = participant2_settlement.participant;
-        require(channel_identifier == getChannelIdentifier(participant1, participant2));
+        require(
+            channel_identifier == getChannelIdentifier(participant1, participant2),
+            "TN/settle: channel id not matching"
```
istankovic, line 830: ... and here.

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -799,27 +825,30 @@ contract TokenNetwork is Utils, Controllable {
...
+        );
 
         Channel storage channel = channels[channel_identifier];
 
-        require(channel.state == ChannelState.Closed);
+        require(channel.state == ChannelState.Closed, "TN/settle: channel not closed");
 
         // Settlement window must be over
-        require(channel.settle_block_number < block.number);
+        require(channel.settle_block_number < block.number, "TN/settle: in settlement timeout");
```
istankovic, line 838: Maybe just drop the "in"?

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -1095,9 +1136,9 @@ contract TokenNetwork is Utils, Controllable {
         pure
         returns (bytes32)
     {
-        require(participant != address(0x0));
-        require(partner != address(0x0));
-        require(participant != partner);
+        require(participant != address(0x0), "TN: participant address zero");
+        require(partner != address(0x0), "TN: partner address zero");
+        require(participant != partner, "TN: identical addresses");
```
istankovic, line 1141: I note that these are identical to the ones in `getChannelIdentifier`, but I guess that should be fine?
Assuming one could determine where it went wrong from the context...

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -1137,9 +1178,9 @@ contract TokenNetwork is Utils, Controllable {
     )
         internal
     {
-        require(channel_identifier == getChannelIdentifier(participant, partner));
-        require(total_deposit > 0);
-        require(total_deposit <= channel_participant_deposit_limit);
+        require(channel_identifier == getChannelIdentifier(participant, partner), "TN/deposit: wrong channel id");
+        require(total_deposit > 0, "TN/deposit: total_deposit is zero");
+        require(total_deposit <= channel_participant_deposit_limit, "TN/deposit: participant deposit limit reached");
```
istankovic, line 1183: Maybe just "deposit limit reached"? Participant is a long word, and it's anyway clear what we're talking about here since we know we are in `deposit`.

raiden_contracts/tests/test_channel_deposit.py
```diff
@@ -198,7 +198,7 @@ def test_channel_deposit_overflow(
     channel_identifier = create_channel(A, B)[0]
     channel_deposit(channel_identifier, A, deposit_A, B)
 
-    with pytest.raises(TransactionFailed):
+    with pytest.raises(TransactionFailed, match="abc"):
```
istankovic, line 201: What does "abc" refer to?

istankovic (approved):
Looks great, also nice to see testing. Looking forward to this landing!

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -447,7 +447,10 @@ contract TokenNetwork is Utils, Controllable {
         );
 
         // Validate that authenticated partners and the channel identifier match
-        require(channel_identifier == getChannelIdentifier(withdraw_data.participant, partner), "WD: channel id not matching");
+        require(
+            channel_identifier == getChannelIdentifier(withdraw_data.participant, partner),
+            "TN/withdraw: channel id not matching"
```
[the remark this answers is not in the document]
palango: Good idea! Will fix this!

raiden_contracts/tests/test_channel_deposit.py
palango: Nice catch 👀 
I just used this as a message when replacing all contract invocations. As this test is ignored I've missed that.

raiden_contracts/data/source/raiden/TokenNetwork.sol
palango: 👍

raiden_contracts/data/source/raiden/TokenNetwork.sol
palango: Mh, good point. I guess it would be ok to leave it as is, but an unambiguous message would be nicer! WHat do you think about preficing these here with `TN/hash:`?

istankovic: I'm fine with either way, as I don't think it would matter as the rest of the messages, I just wanted to point it out in case you were not already aware.

palango: OK, then let's leave it like it is for now and if we run into problems it's easy to change.

palango:
When I remove the `TN` prefix everywhere it greatly reduces the contract size. I see no strong argument for it, do you @istankovic ?

istankovic:
> When I remove the `TN` prefix everywhere it greatly reduces the contract size. I see no strong argument for it, do you @istankovic ?

Hm, good question. I guess if we're always in the position that we know which contract we're interacting with, then I'd say we can definitely omit the prefix.

palango:
> Hm, good question. I guess if we're always in the position that we know which contract we're interacting with, then I'd say we can definitely omit the prefix. Thinking further, the function prefixes alone may be a good indicator where the error occurs, although probably not in all cases. So I'd say, let's omit the contract prefix.

I'll do this in a follow-up, then the discussion is clearly separated.

PR #1501, opened 2021-07-30
Revert TokenNetworkUtils library

palango:
Fixes https://github.com/raiden-network/raiden-contracts/issues/1503
Fixes #1434 
Fixes #1435 

Revert the addition of the `TokenNetworkUtils` library
This eases
- deployment and the code in this package
- usage and verification of the contracts in other packages (`raiden` itself)

## Contract size changes

Those seem mostly related to the disabled yul optimizer. But we add it still fits we can add those back later.

Before:
```
running compile_contracts
TokenNetworkUtils 896 3.6458333333333335%
TokenNetwork 17217 70.05615234375%
TokenNetworkRegistry 20515 83.47574869791666%
```

After:
```
running compile_contracts
TokenNetwork 19055 77.53499348958334%
TokenNetworkRegistry 23336 94.95442708333334%
```

### What's tricky about this PR (if any)

----

Any reviewer can check these:

* [ ] If the PR is fixing a bug or adding a feature, add an entry to CHANGELOG.md.
* [ ] If the PR changed a Solidity source, run `make compile_contracts` and add the resulting `raiden_contracts/data/contracts.json` in the PR.
* [ ] If the PR is changing documentation only, add `[skip ci]` in the commit message so Travis does not waste time.
    * [ ] But, if the PR changes comments in a Solidity source, do not add `[skip ci]` and let Travis check the hash of the source.
* [ ] In Python, use keyword arguments
* [ ] Squash unnecessary commits
* [ ] Comment commits
* [ ] Follow naming conventions
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* [ ] Follow the Signature Convention in [CONTRIBUTING.md](./CONTRIBUTING.md)
* For each new contract
    * [ ] The deployment script deploys the new contract.
    * [ ] `etherscan_verify.py` runs on the new contract.
* Bookkeep
    * [ ] The gas cost of new functions are stored in `gas.json`.
* Solidity specific conventions
    * [ ] Document arguments of functions in natspec
    * [ ] Care reentrancy problems
    * if the PR adds or removes `require()` or `assert()`
        * [ ] add an entry in Changelog
        * [ ] open an issue in the client, light client, service repos so the change is reflected there
        * Just adding a message in `require()` doesn't require these steps.
* [ ] When you catch a require() failure in Solidity, look for a specific error message like `pytest.raises(TransactionFailed, match="error message"):`

[...]

PR #1506, opened 2021-08-04
Remove prefixes from error messages

palango:
Needs to be rebased on https://github.com/raiden-network/raiden-contracts/pull/1501

### What this PR does

Remove `TN` and `TNR` prefixes from contracts

### Why I'm making this PR

This slightly reduces contracts size at the expense of more explicit error messages.

Before:
```
running compile_contracts
TokenNetworkRegistry 23336 94.95442708333334
```

After:
```
running compile_contracts
TokenNetworkRegistry 23071 93.87613932291666
```

We need to decide if we want to do this.

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep


karlb (approved):
According to https://github.com/raiden-network/raiden-contracts/issues/81#issuecomment-504454806, all messages <=32 bytes generate the same contract size. We can take that into account when optimizing the message length.

palango:
As we could re-enable the Yul optimizer, those reductions aren't that important any more and we'll stay with the mroe explicit variant for now.

PR #1519, opened 2021-08-12
Use block.chainid

### LINKED ISSUE #1187: Use the chain_id instruction in Istanbul 1

karlb:
This keeps the parameter lists shorter, is always correct and slightly cheaper.

I have to rely on the chain_id from the TokenNetwork even in the python code due to ethereum/web3.py#1677.

Closes #1187.
Replaces and thus closes #1399.


raiden_contracts/data/source/raiden/TokenNetworkRegistry.sol
```diff
@@ -42,20 +41,18 @@ contract TokenNetworkRegistry is Utils, Controllable {
     /// MAX_UINT256 means no limits
     constructor(
         address _secret_registry_address,
-        uint256 _chain_id,
+        uint256 _chain_id,  // Ignored and only left for backward compatibility
```
palango, line 44: I see the point of leaving this in, but I'd rather remove it as the adaptations for this should be small.

PR #1556, opened 2021-11-11
Hardcode settlement timeouts

netcriptus:


raiden_contracts/tests/fixtures/channel.py
```diff
@@ -44,14 +44,14 @@
 
 @pytest.fixture(scope="session")
 def create_channel(token_network: Contract) -> Callable:
-    def get(A: HexAddress, B: HexAddress, settle_timeout: int = TEST_SETTLE_TIMEOUT_MIN) -> Tuple:
```
palango, line 47: Ah damn, we probably want to make this configurable so that we can set it low in the tests.

netcriptus:
Closed in favor of https://github.com/raiden-network/raiden-contracts/pull/1557

Issue #1550, opened 2021-11-08
Hardcode settlement timeout


palango:
After discussion at the retreat we want to get rid of the configurable settle timeouts at channel creation.

This issue tracks the removal of this option and dependent cleanups.

- [x] Remove `settle_timeout` parameter in `TokenNetwork.openChannel` and `TokenNetwork.openChannelWithDeposit` (https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/data/source/raiden/TokenNetwork.sol#L265)
- [x] Remove `_settlement_timeout_min` and `_settlement_timeout_max` parameters in `TokenNetwork` constructor (https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/data/source/raiden/TokenNetwork.sol#L220)
- [x] Remove `_settlement_timeout_min` and `_settlement_timeout_max` parameters in `TokenNetworkRegistry` constructor (https://github.com/raiden-network/raiden-contracts/blob/master/raiden_contracts/data/source/raiden/TokenNetworkRegistry.sol#L41)
- [x] Fix/cleanup tests

PR #1557, opened 2021-11-15
Remove configurable timeout range

netcriptus:
Solves https://github.com/raiden-network/raiden-contracts/issues/1550


raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -256,7 +256,7 @@ contract MonitoringService is Utils {
 
         // We don't actually know when the channel has been closed. So we'll
         // make a guess so that assumed_close_block >= real_close_block.
-        uint256 assumed_settle_timeout = token_network.settlement_timeout_min();
+        uint256 assumed_settle_timeout = token_network.settle_timeout();
```
karlb, line 259: Nice, this even increases accuracy! We do know the exact settle time now, don't we? If that's the case, the comment and variable name should be updated to reflect that.

karlb (approved):
Good to have this in! One thing less to worry about, some gas savings and nearly 500 lines less.

If we wanted to save even more gas, we could set the settle timeout as a constant in the contracts, but that would require recompilation of the contracts when changing the value. Probably not worth the effort for this case along, but it might be interesting to keep in mind if we spot more cases like it.

I would have preferred multiple small commits:
* Actual code changes
* Contract recompilation and gas measurements (could also be split in two)
* Deployment on testnets
* Deployment on mainnet

raiden_contracts/data/source/services/MonitoringService.sol
netcriptus: Good point, this went unnoticed ✅

raiden_contracts/data/deployment_goerli.json
```diff
@@ -16,8 +16,7 @@
       "gas_cost": 4579912,
       "constructor_arguments": [
         "0xDEF219a11e3d7485075704Ec9ade6eF994c0b5B7",
-        20,
-        555428,
+        200,
```
palango, line 19: Why did you change these? These files are written when depolying the contracts, and shouldn't be changed by hand.

netcriptus: I couldn't find where these values were being set, I assumed they should be manually altered in the deployment files.

palango: No worries, we can sit together and deploy the contracts on goerli when we're both in the office. Just to see how the process looks like and when those files are created.

Issue #1565, opened 2021-11-22
Raiden on Arbitrum: Replace block numbers with timestamps


taleldayekh:
@andrevmatos @palango to provide details.

On L2s blocks might get generated on-demand, and therefore the average block time might change quite drastically. 

Some operations in the current channel lifecycle rely on giving the other channel partner to act. This has to be true on L2s as well. We therefore propose to change timeouts from blocks to seconds, so that that invariant can be kept when block times vary.

PR #1573, opened 2021-12-03
Block number to timestamp

netcriptus:
Implements https://github.com/raiden-network/raiden-contracts/issues/1565


raiden_contracts/constants.py
```diff
@@ -29,9 +29,9 @@
 CONTRACT_DEPOSIT = "Deposit"
 
 # Timeouts
-TEST_SETTLE_TIMEOUT = 8
+TEST_SETTLE_TIMEOUT = 8 * 15
```
palango, line 32: Maybe add some comment that this are seconds

raiden_contracts/data_0.37.0/source/raiden/SecretRegistry.sol
```diff
@@ -41,7 +41,7 @@ contract SecretRegistry {
     /// @notice Get the stored block number at which the secret was revealed
     /// @param secrethash The hash of the registered secret `keccak256(secret)`
     /// @return The block number at which the secret was revealed
-    function getSecretRevealBlockHeight(bytes32 secrethash) public view returns (uint256) {
+    function getSecretRevealBlockTime(bytes32 secrethash) public view returns (uint256) {
```
palango, line 44: All files in the subfolders `data_X.X.X` should not be changed, as they're snapshots of the contracts at the given version.

raiden_contracts/tests/fixtures/channel.py
```diff
@@ -385,8 +385,8 @@ def get(tx_from: HexAddress, transfers: List[Tuple]) -> None:
                 {"from": tx_from},
             )
             assert (
-                secret_registry_contract.functions.getSecretRevealBlockHeight(secrethash).call()
-                == web3.eth.block_number
+                secret_registry_contract.functions.getSecretRevealBlockTime(secrethash).call()
+                == web3.eth.get_block("latest").timestamp  # type: ignore
```
palango, line 389: what's the typing problem here?

raiden_contracts/tests/test_channel_cooperative_settle.py
```diff
@@ -10,7 +11,7 @@
 from raiden_contracts.utils.events import check_channel_settled, check_withdraw_2
 from raiden_contracts.utils.type_aliases import BlockExpiration
 
-EXPIRATION = BlockExpiration(100)
+EXPIRATION = int(time.time()) + BlockExpiration(100) * 15
```
palango, line 14: Shouldn't this just be an relative value as before?

palango (changes requested):
Looks good in general, good work!

- The existing versioned folders shouldn't be changed
- The typing of getting the block timestamps make this quite ugly, is there a way to improve this?

raiden_contracts/constants.py
netcriptus: ✅

raiden_contracts/tests/fixtures/channel.py
netcriptus: Mypy says BlockData type doesn't have a `timestamp` member.

raiden_contracts/tests/test_channel_cooperative_settle.py
netcriptus: This was being used in an absolute way (using the 100th block as the expiration block), so we have to calculate the timestamp of the 100th block to keep things as they were.

raiden_contracts/tests/fixtures/channel.py
palango: Weird, looks like it should:
```py
class BlockData(TypedDict, total=False):
    baseFeePerGas: Wei
    ...
    timestamp: Timestamp

Comment (netcriptus) [reply raiden_contracts/tests/fixtures/channel.py:389]: It definitely should, but:

Screenshot 2021-12-09 at 11 28 25

Comment (istankovic) [reply raiden_contracts/tests/fixtures/channel.py:389]: I think I also encountered this once.
Would not be surprised to find out this is just one of the myriad of mypy bugs.

File: raiden_contracts/data_0.37.0/source/raiden/SecretRegistry.sol (hunk 2 above)
Comment (netcriptus) [reply raiden_contracts/data_0.37.0/source/raiden/SecretRegistry.sol]: ✅

File: raiden_contracts/tests/fixtures/channel.py (hunk 3 above)
Comment (weilbith) [reply raiden_contracts/tests/fixtures/channel.py:390]: I also get this issue all the time. For some reason I also don't get this field as completion candidate while coding. Dunno whats wrong.
Do we agree we can close this conversation here?

File: raiden_contracts/data/source/services/MonitoringService.sol [hunk 5]
```diff
@@ -203,7 +203,7 @@ contract MonitoringService is Utils {
             closing_participant, non_closing_participant
         );
         require(
-            block.number >= firstBlockAllowedToMonitorChannel(
+            block.timestamp >= firstTimestampAllowedToMonitorChannel(
```
palango, line 206: This will make it harder for the MS to predict when to send its tx. But I see no way around this.

raiden_contracts/data/source/raiden/TokenNetworkRegistry.sol
```diff
@@ -32,7 +32,7 @@ contract TokenNetworkRegistry is Utils, Controllable {
     /// @param _secret_registry_address The address of SecretRegistry that's used by all
     /// TokenNetworks created by this contract
     /// that can be chosen at the channel opening
-    /// @param _settle_timeout Number of blocks that need to be mined between a
+    /// @param _settle_timeout Number of seconds that need to elapse between a
```
weilbith, line 35: This is the closest I can put the comment: in line `16`, but also in the `constructor` there is the variable `(_)settle_timeout`. The naming is on discussion by the review for the `TokenNetwork` contract. This comment is just a reminder that this naming is used here too. Though you might end up with a naming that makes this just valid as is.

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -248,27 +248,27 @@ contract MonitoringService is Utils {
...
+        (settle_window, channel_state) = token_network.getChannelInfo(
             channel_identifier, closing_participant, non_closing_participant
         );
         require(channel_state == TokenNetwork.ChannelState.Closed, "channel not closed");
 
         uint256 settle_timeout = token_network.settle_timeout();
-        require(settle_block_number >= settle_timeout, "too low settle block number");
-        uint256 assumed_close_block = settle_block_number - settle_timeout;
+        require(settle_window >= settle_timeout, "too low settle block number");
+        uint256 assumed_close_timestamp = settle_window - settle_timeout;
```
palango, line 259: This is not assumed anymore after https://github.com/raiden-network/raiden-contracts/pull/1557 , right?

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -279,16 +279,16 @@ contract MonitoringService is Utils {
     {
         // avoid overflows when multiplying with percentages
         require(settle_timeout < MAX_SAFE_UINT256 / 100, "maliciously big settle timeout");
```
palango, line 281: same here, this is a know value now

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -279,16 +279,16 @@ contract MonitoringService is Utils {
...
-        require(closed_at_block < MAX_SAFE_UINT256 / 100, "maliciously big closed_at_block");
+        require(closed_at_timestamp < MAX_SAFE_UINT256 / 100, "maliciously big closed_at_timestamp");
```
palango, line 282: How would this ever happen?

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -248,27 +248,27 @@ contract MonitoringService is Utils {
         require(service_registry.hasValidRegistration(monitoring_service_address), "service not registered");
 
         TokenNetwork.ChannelState channel_state;
-        uint256 settle_block_number;
-        (settle_block_number, channel_state) = token_network.getChannelInfo(
+        uint256 settle_window;
```
weilbith, line 251: Reminder for the naming scheme application (new file). Sorry to be very verbose with these comments. 🙈

raiden_contracts/tests/fixtures/one_to_n.py
```diff
@@ -49,17 +49,17 @@ def f(
...
         amount: int = 10,
-        expiration_block: int = None,
+        expiration_timestamp: int = None,
         chain_id: int = chain_id,
         one_to_n_address: HexAddress = one_to_n_contract.address,
     ) -> dict:
-        if expiration_block is None:
-            expiration_block = web3.eth.block_number + 10
+        if expiration_timestamp is None:
+            expiration_timestamp = web3.eth.get_block("latest").timestamp + 150  # type: ignore
```
palango, line 57: Would it make sense to create a type safe wrapper for `web3.eth.get_block("latest").timestamp` and use that everywhere?

raiden_contracts/data/source/services/OneToN.sol
```diff
@@ -34,7 +34,7 @@ contract OneToN is Utils {
     event Claimed(
         address sender,
         address indexed receiver,
-        uint256 expiration_block,
+        uint256 expiration_timestamp,
```
weilbith, line 37: Reminder for the naming scheme application (new file). Sorry to be very verbose with these comments. 🙈

raiden_contracts/tests/fixtures/one_to_n.py
netcriptus: I definitely makes sense. I'll rework this part.

raiden_contracts/data/source/services/MonitoringService.sol
```diff
@@ -248,27 +248,27 @@ contract MonitoringService is Utils {
...
+        (settle_window, channel_state) = token_network.getChannelInfo(
             channel_identifier, closing_participant, non_closing_participant
         );
         require(channel_state == TokenNetwork.ChannelState.Closed, "channel not closed");
 
         uint256 settle_timeout = token_network.settle_timeout();
-        require(settle_block_number >= settle_timeout, "too low settle block number");
-        uint256 assumed_close_block = settle_block_number - settle_timeout;
+        require(settle_window >= settle_timeout, "too low settle block number");
```
weilbith, line 258: The rejected message requires some adaption.

weilbith (changes requested):
First review part for the contracts. The second part for the tests will follow now. I tried to be more careful here and also read the not touched lines of code as this is quite of a big concept change that affect many parts. 
My main concern that many comments reference to together is the naming scheme of the "timeout" concept. The rest are just a few places where the old naming is still used. Function wise everything seems fine. Good work! 👍🏾

raiden_contracts/constants.py
```diff
@@ -29,9 +29,9 @@
 CONTRACT_DEPOSIT = "Deposit"
 
 # Timeouts
-TEST_SETTLE_TIMEOUT = 8
+TEST_SETTLE_TIMEOUT = 8 * 15  # Timeout in seconds
```
weilbith, line 32: Nitpick: this is somewhat awkward. What does `8 * 15` express? Are `15` seconds something special and then take it 8 times? I would have expected something like `2 * 60` or so to express 2 minutes or so. Else just the pure number. The same goes for the timeout value below.

raiden_contracts/data/gas.json
```diff
@@ -3,32 +3,32 @@
     "CustomToken.mint": 50776,
     "CustomToken.transfer": 36556,
     "CustomToken.transferFrom": 29060,
-    "MonitoringService.claimReward": 45555,
-    "MonitoringService.monitor": 224140,
-    "OneToN.bulkClaim 1 ious": 106874,
-    "OneToN.bulkClaim 6 ious": 165202,
-    "OneToN.claim": 103516,
-    "SecretRegistry.registerSecret": 46205,
+    "MonitoringService.claimReward": 45566,
```
weilbith, line 6: Not an actual review comment. But I'm a little confused why we became more expensive. Shouldn't it stay basically equal? 🤔 
Though the relative change is so small it probably doesn't matter. Just a self-reflection of what has changed that affects gas usage.

raiden_contracts/tests/fixtures/base/utils.py
```diff
@@ -176,8 +176,9 @@ def get(txn_hash: str, message: Optional[str] = None, additional_gas: int = 0) -
 
 @pytest.fixture()
 def get_block(web3: Web3) -> Callable:
-    def get(txn_hash: HexBytes) -> int:
+    def get(txn_hash: HexBytes) -> BlockData:
```
weilbith, line 179: Nitpick (probably out-of-scope for PR): I dunno how well your IDE support for Pytest is. Do you get the signature of this function in the test cases? It kinda confuses me as the name is not very descriptive. My first intention is that this is again just a mimic of `web3.eth.get_block`. But its actual purpose is to get the block for a given transaction hash. Just that nothing is expressing this except the function parameter for the getter of this fixture where IDE support might be week. At least the code where this gets used reads not very nice.
Anyway, I guess you can just close this conversation. But in case your IDE supports reliable renaming for Pytest fixtures, consider to rename it if you agree with me. No further discussion necessary.

raiden_contracts/tests/fixtures/contracts.py
```diff
@@ -31,10 +30,10 @@ def deploy_contract_txhash(
     else:
         contract = web3.eth.contract(abi=abi, bytecode=bytecode)
 
-    mine_blocks(web3, 3)
+    web3.testing.mine(3)  # type: ignore
```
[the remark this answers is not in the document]
weilbith: The `mine_blocks` function got removed everywhere. As this function is nothing else but a wrapper for this line of code here (wonder why it ever was a function as it causes more lines of code due to the import, still required two parameters and doesn't provide better naming that would lead to better readability), I just replaced it in-place here. It would not be worth to keep this whole file of a single function of a single line just for using it twice only in this single file. Everywhere else we now need time traveling.

raiden_contracts/tests/fixtures/channel.py
```diff
@@ -359,7 +358,9 @@ def get(
             participant2_values,
         )
 
-        mine_blocks(web3, settle_timeout)
+        web3.provider.ethereum_tester.time_travel(  # type: ignore
+            web3.eth.get_block("latest").timestamp + settle_timeout + 1  # type: ignore
+        )
```
weilbith, line 363: I know this is my own code. I probably should just do it myself. But do you agree that it makes sense to shim this down to a utility function? Would also remove the need to always put this stupid `# type: ignore` and maybe save some imports of the `TEST_SETTLE_TIMEOUT` variable. Definitely should improve readability.

raiden_contracts/tests/test_channel_cooperative_settle.py
weilbith, line 14: Sorry for putting this comment at such a bad location, but does the type `BlockExpiration` still make sense this way? 🤔 Searching through its code reference doesn't seem like this is properly used anymore. Do you agree?

raiden_contracts/tests/fixtures/channel.py
weilbith: Duplicate of [this comment](https://github.com/raiden-network/raiden-contracts/pull/1573/files?authenticity_token=Kar%2BtamiGrDSqMZE9CYL98ryCissNBtNCOrxdBhZaAbFmXQ%2BYMIOwCGEXVJQ%2FU39lqwKk4i4r2nzR%2FXd0eyN5A%3D%3D&file-filters%5B%5D=.json&file-filters%5B%5D=.py&hide-deleted-files=true#r769468919). Please just close it when done. Just keep it for the content. 🤣

raiden_contracts/tests/test_channel_unlock.py
```diff
@@ -541,7 +544,7 @@ def test_channel_unlock_no_locked_amount_fail(
 ) -> None:
     """After settleChannel() is called with zero locked amount, unlock() calls fail"""
     (A, B) = get_accounts(2)
-    settle_timeout = 8
+    settle_timeout = TEST_SETTLE_TIMEOUT
```
weilbith, line 547: What is the purpose of this variable assignment? Doesn't seem to add any value to the code and is rather annoying. I think I even removed it at some places when I was editing the tests to fix them.

raiden_contracts/tests/test_channel_withdraw.py
```diff
@@ -153,20 +152,20 @@ def test_withdraw_call_near_expiration(
...
-            total_withdraw=withdraw_A,
-            expiration_block=expiration,
-            participant_signature=signature_A_for_A,
-            partner_signature=signature_B_for_A,
+            channel_identifier,
+            A,
+            withdraw_A,
+            expiration,
+            signature_A_for_A,
+            signature_B_for_A,
```
weilbith, line 168: Just for knowledge sharing: why did this work in the past and now caused all these signature mismatch errors? :thinking:
EDIT: below for the `firstTimestampAllowedToMonitorChannel` function call it works again...

raiden_contracts/tests/test_print_gas.py
```diff
@@ -156,7 +155,7 @@ def print_gas_channel_cycle(
 ) -> None:
     """Abusing pytest to print gas costs of TokenNetwork's operations"""
     (A, B, C, D) = get_accounts(4)
-    settle_timeout = 11
+    settle_timeout = TEST_SETTLE_TIMEOUT
```
weilbith, line 158: Before the `settle_timeout` was always set to `8` and we replaced it with `TEST_SETTLE_TIMEOUT` which was the equivalent. But here it is now explicitly `11`. Either this was unintended, but to me this reads like something special. Why else should the developer do this here. So do we need to take this into account here? Because now we just ignore it.

weilbith (commented):
Second part of the review for the tests. As I worked on the tests partially myself for this PR I don't have much here. Mostly nitpicks and comments to other conversations.

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -104,7 +104,7 @@ contract TokenNetwork is Utils, Controllable {
     struct WithdrawInput {
         address participant;
         uint256 total_withdraw;
-        uint256 expiration_block;
+        uint256 expiration_time;
```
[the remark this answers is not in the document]
netcriptus: This one is actually a different concept, and the name of the variable could probably be changed to `expiration_timestamp` if it would make it more clear.

When we have something like `settle_timestamp`, this refers to the name of seconds from an event until something can be settled, so it's relative time. In this case `expiration_time` means an absolute timestamp, not dependent on any other value.

raiden_contracts/constants.py
netcriptus: We were using 8 blocks as a timeout. However, in arbitrum, we can't be sure the time between each block is the same. So we add here the expected time between blocks, which is 15 seconds * 8 blocks

raiden_contracts/data/gas.json
netcriptus: This is automatically updated with the command `make update_gas_costs`.

raiden_contracts/tests/test_channel_withdraw.py
netcriptus: I don't know why this stopped working. Maybe @palango could have some explanation? Maybe something is missing to link the python call to the solidity signature?

raiden_contracts/tests/test_print_gas.py
netcriptus: Setting the settle timeout is no longer a valid thing. We have this value hardcoded, therefore this value is no longer valid. We're still using this variable bellow to check the final result of the settlement.

raiden_contracts/data/source/raiden/TokenNetwork.sol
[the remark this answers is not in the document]
netcriptus: To be fixed in https://github.com/raiden-network/light-client/issues/3015

raiden_contracts/data/source/services/MonitoringService.sol
netcriptus: To be fixed in https://github.com/raiden-network/light-client/issues/3015

raiden_contracts/data/source/services/OneToN.sol
netcriptus: To be fixed in https://github.com/raiden-network/light-client/issues/3015

weilbith (approved):
This PR is now good as is. We have follow-up issues #1580 and #1582.

palango (approved):

PR #1586, opened 2022-01-06
Arbitrum: remove settle_timeut from Channel state and events

### LINKED ISSUE #1582: Raiden on Arbitrum: remove settle_timeout from channel state and events
#1557 removed the variable settle timeout, but still left it in the channel state, wasting gas (store is expensive). We want to remove the `settle_timeout` constant (per deployment) from all stores/mappings and events, to save gas, and to ensure clients fetch the global value directly from the contracts from now on.

One reason why it's still in channel state/struct is because this variable is reused to store the point in time after which the channel becomes settleable. A possible implementation is to have this timestamp persisted in a separate mapping, which also gets cleared upon settlement. Notice that this saves gas by not needing the store space/write operations on channel open, and if the channel gets coop-settled, actually never gets set.

andrevmatos:
### What this PR does
Closes #1582 

To do this, we moved the old re-use of `channel.settle_timeout` to its own mapping, saving gas as well since this is only required upon uncooperative close.

----

Any reviewer can check these:

* [ ] Comment commits
    * `solidityFunction`
    * `_solidity_argument`
    * `solidity_variable`
    * `python_variable`
    * `PYTHON_CONSTANT`
* For each new contract
* Bookkeep


andrevmatos:
`getChannelInfo` now returns only the `state`, instead of the pair `(settle_timeout, state)` (since `settle_timeout` is now a contract's constant, instead of a per-channel variable). Should we rename the method to `getChannelState`? Should we do it on this PR, or in a follow up? Or isn't it worth it?

netcriptus (approved):
LGTM, but I think it would be a great opportunity to change the name of the function like you mentioned. I see no reason to implement this in a different PR.

netcriptus (approved):

Issue #1580, opened 2021-12-16
Raiden on Arbitrum: fix timeouts naming convention


netcriptus:
As pointed in [this](https://github.com/raiden-network/raiden-contracts/pull/1573#discussion_r769433236) and [this](https://github.com/raiden-network/raiden-contracts/pull/1573#discussion_r769438688) comments, there's some confusion regarding the naming of timeouts in the Smart Contracts.

There are 2 concepts of time going around, one for absolute timestamp, and another one for number of seconds an event, and the names do not reflect it at all.

Refactor the contracts with better naming, and unifying names that should reflect the same thing.

andrevmatos:
My suggestion would be be `_timestamp` for absolute points in time, and `_timeout` for relative time deltas, but I don't have strong feelings for any particular naming schema, as long as it's consistent.

I think we should take the time to do this renaming in #1573 , since that's handling exactly this change, so we avoid even polluting commit history with names that aren't descriptive.

weilbith:
@andrevmatos and I have discussed this topic now and settled on a scheme.
There are two different concepts of the timing functionality here for 3 cases. All of them specify a relative time value that gets added on top of the current time when an event happens. So for example a time value for settling a channel, that gets added to the timestamp when a channel gets closed.
The individual future timestamp that gets calculated has then two different meanings, depending on the context. In the case of the channel settle part, it means the channel becomes settleable after this timestamp has passed. In another case like for withdrawing funds from a channel it means it can be withdrawn until this timestamp has passed. It is our desire to express this significant difference in the code and make it very easy to read and understand without much guessing.

Thereby we ended up with calling these variables by the following concept:
- `*_timeout` (e.g. `settle_timeout`) - relative time value that gets added to event timestamp to get any of the following values
- `*able_until` (e.g. `withdrawable_until`) - absolute timestamp until which it is possible to execute a specific action
- `*able_after` (e.g. `settleable_after`) - absolute timestamp only after which it is possible to execute a specific action

PR #1590, opened 2022-01-10
rename timestamp related variables

netcriptus:
Implements https://github.com/raiden-network/raiden-contracts/issues/1580


raiden_contracts/data/source/services/UserDeposit.sol
```diff
@@ -161,7 +161,7 @@ contract UserDeposit is Utils {
 
         withdraw_plans[msg.sender] = WithdrawPlan({
             amount: amount,
-            withdraw_timestamp: block.timestamp + withdraw_delay
+            withdrawable_after: block.timestamp + withdraw_delay
```
andrevmatos, line 164: Per the description of the issue, shouldn't we also rename `withdraw_delay` to `withdraw_timeout`? Does this variable match the `_timeout` usecase, or is it something different?

andrevmatos (changes requested):
The changes here seem correct, but there're still variables which need renaming in the contracts; e.g. `expiration_time` in `TokenNetwork.setTotalWithdraw`; could you please give another pass and rename any variable or parameter name in any contract which should mach the new naming schema?

raiden_contracts/data/source/services/UserDeposit.sol
netcriptus: Good call ✅

raiden_contracts/data/source/raiden/TokenNetwork.sol
```diff
@@ -1596,7 +1596,7 @@ contract TokenNetwork is Utils, Controllable {
         uint256 channel_identifier,
         address participant,
         uint256 total_withdraw,
-        uint256 expiration_timestamp,
+        uint256 recoverable_until,
```
andrevmatos, line 1599: This function is called directly with parameters from `WithdrawInput`, therefore I think the name here should be the same from the struct:
```suggestion
        uint256 withdrawable_until,
```

andrevmatos (commented):
Just one change now, thanks!

raiden_contracts/data/source/raiden/TokenNetwork.sol
netcriptus: Nice catch ✅

andrevmatos (approved):
LGTM, nice refactoring, thank you

PR #1604, opened 2022-02-14
Deployment on Rinkeby Arbitrum

weilbith:
Superseeds #1542 (due to the massive changes recently a rebase was not possible)

This includes the fixed for the deployment tools to work with Arbitrum chains with a special handling for gas.

### DIFF
--- a/raiden_contracts/deploy/contract_deployer.py
+++ b/raiden_contracts/deploy/contract_deployer.py
@@ -22,6 +22,7 @@ from raiden_contracts.constants import (
     CONTRACT_USER_DEPOSIT,
     CONTRACTS_VERSION,
     DEPLOY_SETTLE_TIMEOUT,
+    ID_TO_CHAINNAME,
     DeploymentModule,
 )
 from raiden_contracts.contract_manager import (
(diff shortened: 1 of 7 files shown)


raiden_contracts/deploy/contract_deployer.py
```diff
@@ -77,6 +79,20 @@ def __init__(
         else:
             LOG.info("Skipped checks against the source code because it is not available.")
 
+    def _maybe_set_arbitrum_settings(self) -> None:
```
netcriptus, line 82: I'm personally not a big fan of function names with the word `maybe`. This is bad for readability and understanding of the code. How about using something more generalist, like `_adjust_gas_special_case`, or even `_adjust_chain_settings`?

This way, if we have another case that needs special settings, it can be included here as well.

weilbith: ✔️

weilbith:
I pushed a fixup commit to clearly address the review feedback.
As I got an optimistic approval, I continue and integrate the fixup commit to get this PR ready for the final CI verification.

weilbith:
Sorry. I was too fast. I thought I actually run `make lint` locally and it succeed. Apparently not.

PR #1610, opened 2022-03-01
improve gas handling for Arbitrum based networks

### LINKED ISSUE #1608: Proper gas price strategy for Arbitrum networks
Currently, Arbitrum based chains get detected and get a special handling for gas prices. Unfortunately is the currently implemented strategy very simple and only suitable for test networks (because very expensive). This needs to be replaced with a better and more accurate calculation.

weilbith:
Closes #1608 

First of all, it is not necessary to programmatically set a fixed gas price value when detecting an Arbitrum network. The gas price gets already set by a commandline parameter and does not need any further doing. The default 5 Gwei
are enough for the current prices (test network). Else the user can/must set the value higher.

As the deployment tools are meant to run by hand and usually a developer of the team, it is not necessary to add any special gas price strategies as we do for the Raiden services. But to support the correct/better usage, the deployment documentation got extended for Arbitrum.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -231,6 +231,11 @@ Deployment information is stored in a ``deployment_[CHAIN_NAME].json`` file corr
     # Based on the network id, the script verifies the corresponding deployment_[CHAIN_NAME].json file
     # using the chain name-id mapping from constants.py
 
+Deployment on an Arbitrum based network
+-----------------------
+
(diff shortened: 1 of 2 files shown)


weilbith:
@netcriptus do you know why the `codecov/patch` action is failing? I have no knowledge about the setup of this repository.

netcriptus:
> @netcriptus do you know why the `codecov/patch` action is failing? I have no knowledge about the setup of this repository.

I'm also not familiar with the setup. I can only see the warning about the removal of an untested line of code. Looks like this can be ignored.

weilbith:
@palango do you maybe know the issue here?

palango:
> @palango do you maybe know the issue here?

`Added line #L92 was not covered by tests`

So because the coverage of the diff of the PR is too low this fails.

weilbith:
Okay, it confused me because in the action summary it states the coverage is `0.00%` which usually means it somehow doesn't work properly.
Okay, so I somehow need to write a test for this. 🤔

palango:
Yeah, just ignore it.

PR #1613, opened 2022-03-03
Feature/user deposit customizable withdraw timeout

### LINKED ISSUE #1612: User Deposit Contract - allow to configure withdraw timeout
As the timeout value has changed from block numbers to time in seconds, this has become a challenge for (end-to-end) tests. The current constant value are 25min. That is way too long. In the past we could just mine a couple of blocks. But manipulating pure time is not a good non unit tests.
Therefore it should be possible to set the withdraw timeout in the constructor of the User Deposit contract. This is the same strategy as we already do for the Token Network and Token Registry contract. The deployment tools will then contain the current constant value as the default for the new optional argument.

weilbith:
Closes #1612


weilbith:
Sorry for the included bump of the pre-release stuff. I'll cut that out.

weilbith:
Sorry, my local formatting seem to have caused issues for the contract verification (though I dunno why).

raiden_contracts/data/source/services/UserDeposit.sol
```diff
@@ -6,9 +6,9 @@ import "raiden/Token.sol";
 import "raiden/Utils.sol";
 
 contract UserDeposit is Utils {
-    uint constant public withdraw_timeout = 15 * 100;  // time before withdraw is allowed in seconds
+    uint public withdraw_timeout;  // time before withdraw is allowed in seconds
 
-    // Token to be used for the deposit
+    // Token to be used foruser_deposit_withdraw_timeout the deposit
```
netcriptus, line 11: nitpick: small typo, missing space in `for user_deposit...`

netcriptus (approved):
Looking good, just small comments 👍

raiden_contracts/tests/test_deploy_data.py
```diff
@@ -205,5 +209,6 @@ def test_verify_existent_deployment_with_wrong_code(
         verifier.verify_deployed_service_contracts_in_filesystem(
             token_address=HexAddress(HexStr("0x5Fc523e13fBAc2140F056AD7A96De2cC0C4Cc63A")),
             user_deposit_whole_balance_limit=2 ** 256 - 1,
+            user_deposit_withdraw_timeout=25 * 60,
```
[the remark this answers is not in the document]
weilbith: I did so stupidly because they did so the same with the above `user_deposit_whole_balance_limit`. I always tried to mimic this. 🤷🏾 
But sure, I can do that. Do you want me to? 🙃

raiden_contracts/data/source/services/UserDeposit.sol
weilbith: 🙈

weilbith:
I marked the CodeCov report as resolved because there were no changes in funcitonality.

raiden_contracts/data/source/services/UserDeposit.sol
weilbith: Done ✔️ 
As this is a single commit, I did amend it instead of a fixup commit.

weilbith:
Holy ****. I thought a space character in a comment would not cause a hash mismatch. 😑

PR #1614, opened 2022-03-04
fix versions, add new files to manifesto

netcriptus:
This PR solves the problems we've been having to find the arbitrum deployment files when installing the new version of the contracts


raiden_contracts/data_1.0.0/source/services/UserDeposit.sol
```diff
@@ -6,9 +6,9 @@ import "raiden/Token.sol";
 import "raiden/Utils.sol";
 
 contract UserDeposit is Utils {
-    uint constant public withdraw_timeout = 15 * 100;  // time before withdraw is allowed in seconds
+    uint public withdraw_timeout;  // time before withdraw is allowed in seconds
```
andrevmatos, line 9: Maybe this should be made `immutable`? Would still allow setting it in constructor, but better inform it can't be changed latter

andrevmatos (commented):
Almost good to go, but the deployment files in `data/` should also contain the updated files from `data_1.0.0`, or LC won't be able to find them. Please, copy them there as well

weilbith (commented):
Very good. Thanks for finding this out. 😃 

I think we have agreed to not include "fake" deployments. Should mean in `1.0.0` we have no mainnet etc. But that is mostly independent of the core change you did here.

raiden_contracts/data_1.0.0/source/services/UserDeposit.sol
weilbith: Yes, we could do this at a dozen places in the contracts. We have plenty of contract properties that get set by the constructor exclusively. But the gain of doing so is very little. Putting this into the context of where we are in the project right now, I don't think it is worth the time and effort. Should be a deployment only advantage, if I'm not mistaken.

weilbith: Nevertheless this would be another PR and definitely not part of this one. Therefore I would recommend to create a new issue from this and close the conversation. We can then pull in this new issue any time. Though it isn't my PR, nor my review comment thread, so this is just my strong recommendation.

weilbith (approved):
👌🏾

PR #1633, opened 2022-04-28
v0.50.0: Arbitrum support

karlb:


weilbith (approved):
We went through the last commits together and it looks all fine. 👌🏾

